PR #37, opened 2025-04-21
Add shielded token (to archive)

andrew-fleming:
This PR proposes to add the initial design for the shielded token contract


0xisk:
@andrew-fleming what is the status of this PR?

andrew-fleming:
> what is the status of this PR?

Merging is blocked
Waiting on 1 reapproval from someone other than the last pusher. Review from 0xisk is stale because it was submitted before the most recent code changes. 👇

0xisk:
Got it thank you @andrew-fleming!

contracts/archive/src/ShieldedToken.compact
```diff
@@ -0,0 +1,151 @@
...
+  export circuit mint(recipient: Either<ZswapCoinPublicKey, ContractAddress>, amount: Uint<64>): CoinInfo {
+    assert !Utils_isKeyOrAddressZero(recipient) "ShieldedToken: invalid recipient";
+
+    _counter.increment(1);
+    const newNonce = evolve_nonce(_counter, _nonce);
+    _nonce = newNonce;
+    const ret = mint_token(_domain, amount, _nonce, recipient);
+    _totalSupply = _totalSupply + amount as Uint<64>;
+    return ret;
+  }
```
emnul, line 121: Non-blocking: Shouldn't we add an arithmetic overflow check here?

contracts/archive/src/ShieldedToken.compact
```diff
@@ -0,0 +1,151 @@
...
+ * - Should this be a part of the Shielded module itself or as an extension?
+ */
+module ShieldedToken { // DO NOT USE IN PRODUCTION!
+  import CompactStandardLibrary;
+  import "../../node_modules/@openzeppelin-midnight/utils/src/Utils" prefix Utils_;
+
+  // Public state
+  export ledger _counter: Counter;
+  export ledger _nonce: Bytes<32>;
+  export ledger _totalSupply: Uint<64>;
```
emnul, line 39: Non-blocking: Curious why `_totalSupply` is represented as a Uint<64> value in the shielded impl vs Uint<128> in the unshielded impl

contracts/archive/src/ShieldedToken.compact
```diff
@@ -0,0 +1,151 @@
...
+  export sealed ledger _symbol: Maybe<Opaque<"string">>;
+  export sealed ledger _decimals: Uint<8>;
+
+  /**
+   * @description Initializes the contract by setting the initial nonce
+   * and the metadata.
+   *
+   * @return {[]} - None.
+   */
+  export circuit initializer(
```
emnul, line 51: Created https://github.com/OpenZeppelin/compact-contracts/issues/170 to track updating this contract to use the Initialize pattern

contracts/archive/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "include": ["src/**/*.ts"],
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "dist",
+    "declaration": true,
+    "lib": ["ESNext"],
+    "target": "ES2022",
+    "module": "ESNext",
+    "moduleResolution": "node",
```
emnul, line 10: Tracking update to this file in #170

emnul (approved):
Archived !

contracts/archive/src/ShieldedToken.compact
andrew-fleming: > Non-blocking: Shouldn't we add an arithmetic overflow check here?

AFAICT we're okay. From the [Compact Ref](https://docs.midnight.network/develop/reference/compact/lang-ref#binary-arithmetic-expressions):

> Arithmetic expressions are evaluated by first evaluating the subexpressions in order from left to right. Integer addition, subtraction, or multiplication is then used on the subexpression values. The overflow and underflow behavior differs for Field and Uint operations:
> ...
> - Uint addition and multiplication cannot overflow: the static type of the result will always be large enough to hold the result value

We also have a [test](https://github.com/OpenZeppelin/compact-contracts/blob/erc20-shielded/contracts/archive/src/test/ShieldedToken.test.ts#L119-L126) for overflow anyway. We _could_ add it if we think it's worth improving the error msg though. Do you see other potential issues?

> Better yet we could even implement an _update function.

I think it's def worth looking into! The logic is much different though than just applying arithmetic on state updates like we do in the contract-based mods. Worth considering though, for sure

contracts/archive/src/ShieldedToken.compact
andrew-fleming: Good question! [CoinInfo](https://docs.midnight.network/develop/reference/midnight-api/ledger/type-aliases/CoinInfo) and [QualifiedCoinInfo](https://docs.midnight.network/develop/reference/midnight-api/ledger/type-aliases/QualifiedCoinInfo) => "Bounded to be a non-negative 64-bit integer" (in the `value` description)

So we can't really do anything about that^. And since we're in this weird in-between of unshielded ERC-ish standards with Uint<256> and the early max-uint fluctuations, Uint<128>, to me, was the safe middle ground. There are arguments for both waiting for Uint<256> (or something similar) or bringing it down to 64 bit for size parity with shielded

andrew-fleming:
@emnul yarn.lock strikes again 🤕

contracts/archive/src/ShieldedToken.compact
emnul: > We also have a [test](https://github.com/OpenZeppelin/compact-contracts/blob/erc20-shielded/contracts/archive/src/test/ShieldedToken.test.ts?rgh-link-date=2025-07-11T02%3A10%3A39Z#L119-L126) for overflow anyway. We could add it if we think it's worth improving the error msg though. Do you see other potential issues?

Ahhh okay I misunderstood the behavior. We _want_ that runtime error to occur when we downcast the addition expression and Midnight gives us exactly that. For some reason I thought we didn't want that behavior.

contracts/archive/src/ShieldedToken.compact
emnul: Gotcha, thanks for pointing that out!

emnul (approved):

PR #155, opened 2025-06-26
Add FungibleToken docs

### LINKED ISSUE #93: Add docs for FungibleToken module

andrew-fleming:
Fixes #93.

#### PR Checklist

- [ ] Tests
- [x] Documentation


docs/modules/ROOT/pages/api/fungibleToken.adoc
```diff
@@ -0,0 +1,314 @@
...
+
+This module provides the full FungibleToken module API.
+
+TIP: For an overview of the module, read the {fungible-guide}.
+
+== Core
+
+[.hljs-theme-dark]
+```ts
+import "openzeppelin-compact/fungible-token" prefix FungibleToken_;
```
0xisk, line 14:
```suggestion
import "./node-modules/@openzeppelin-compact/fungible-token/FungibleToken" prefix FungibleToken_;
```

docs/modules/ROOT/pages/api/fungibleToken.adoc
0xisk, line 1: Here’s the updated version including **Dependencies** with your example:

---

### Proposal for a General Module Template

I think we need a consistent template to follow for all modules, covering:

* **Ledger State** (Public state)
* **Witness**
* **Circuits**
* **Internal Circuits**
* **Dependencies** (Modules used within this module, if applicable)

---

### Circuit Navigation and Documentation Improvement

Regarding circuits, I found scrolling through the page cumbersome. On the right sidebar, I expected to see links for *all* circuits, but currently there is only one link, like this:

```
FungibleToken
    Core
        FungibleToken
```

I propose a more detailed and organized structure such as:

```
FungibleToken
    Ledger
        _balances
        _allowances
        ...
    Witness  (Only show if any witness functions exist)
    Circuits
        initialize()
        name()
        ...
        _spendAllowance()
    Internal Circuits
        _update()
    Dependencies (Link to that contract’s dependency page)
        CompactStandardLibrary
        Initializable  
        Utils
```

This structure improves clarity and navigation by grouping related functions and showing sections only when relevant.
Screenshot 2025-07-10 at 11-29-40 FungibleToken - OpenZeppelin Docs


cc: @emnul

0xisk (changes requested):
Thank you @andrew-fleming! The text LGTM, just left general important discussions.

docs/modules/ROOT/pages/api/fungibleToken.adoc
andrew-fleming: The idea with:
```
FungibleToken
    Core
        FungibleToken
```

is to allow this:

```
FungibleToken
    Core
        FungibleToken
        FungibleTokenMetadata (maybe)
    Extensions
        FungibleTokenPermit
        FungibleTokenCapped
        FungibleTokenSomething
        ...
    Utils
        FungibleTokenUtils
```
So that when you click on the item, it brings you to the list of that item's API. IMO that's easier to navigate and cleaner than:

```
FungibleToken
    Ledger
        _balances
        _allowances
        ...
    Witness  (Only show if any witness functions exist)
    Circuits
        initialize()
        name()
        ...
        _spendAllowance()
    Internal Circuits
        _update()
    Dependencies (Link to that contract’s dependency page)
        CompactStandardLibrary
        Initializable  
        Utils

FungibleTokenPermit
    Ledger
        None
    Witness  (Only show if any witness functions exist)
    Circuits
        permit()
        ...
    Internal Circuits
        None
    Dependencies (Link to that contract’s dependency page)
        CompactStandardLibrary
        Initializable  
        Utils

FungibleTokenSomething
    Ledger
    ...
...
```

Do you disagree? Or maybe there's a better way to handle expanding the docs? Or do you suggest kicking the can and dealing with expanding the docs later?

To better visualize, the current approach is how the [sol](https://docs.openzeppelin.com/contracts/5.x/api/token/erc20) and [cairo](https://docs.openzeppelin.com/contracts-cairo/2.0.0/api/erc20) docs are organized

andrew-fleming: If we were to use your proposed approach:

1. WDYT about removing `Ledger`? It is more often than not redundant i.e. `_name`, `_balances`. Perhaps it's helpful with nested maps. Maybe it's worth it for that, idk. I don't love having it as part of the sidebar though
2. I don't disagree that the module deps should be documented in the API, but again, does this really need to be part of the sidebar?
3. Agree with separating `Circuits`, `Internal Circuits`, and `Witnesses`

docs/modules/ROOT/pages/fungibleToken.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+Import the FungibleToken module into the implementing contract.
+It's recommended to prefix the module with `FungibleToken_` to avoid ccircuit signature clashes.
+
+```typescript
+// MyFungibleToken.compact
+pragma language_version >= 0.15.0;
+
+import CompactStandardLibrary;
+import openzeppelin-compact/fungible-token prefix FungibleToken_;
```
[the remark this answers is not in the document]
andrew-fleming: > So we don't have to use the node_modules approach internally. I actually think it's better if imports use relative paths to the source code vs node_modules

Separate discussion though^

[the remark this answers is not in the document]
andrew-fleming: > Future Improvements for Compact Imports

1. Yes!
2. Yes!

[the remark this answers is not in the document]
andrew-fleming: >> For sure, but we're also discussing it now publicly :)

> Haha no I meant if you would like a meeting or so.

Just messing with you. Yeah, we need to plan this out well

>> So we don't have to use the node_modules approach internally. I actually think it's better if imports use relative paths to the source code vs node_modules

> Yes internally it is fine but if externally we have to always dig through node_modules.

Def not ideal "./node_modules/..." 👎

emnul:
> Future Improvements for Compact Imports
Keen observations here @0xisk thanks for creating those issues! I agree with both points.

docs/modules/ROOT/pages/fungibleToken.adoc
```diff
@@ -0,0 +1,56 @@
...
+Transferring tokens to a contract may result in those tokens being locked forever.
+The FungibleToken module, however, does provide `unsafe` circuit variants for users who wish to experiment with sending tokens to contracts.
+
+WARNING: The `unsafe` circuits will eventually be deprecated after Compact supports contract-to-contract calls—meaning
+`transfer`, `_mint`, etc. are planned to eventually allow the recipients to be of the `ContractAddress` type.
+
+== Usage
+
+Import the FungibleToken module into the implementing contract.
+It's recommended to prefix the module with `FungibleToken_` to avoid ccircuit signature clashes.
```
emnul, line 40: Non-blocking: Caught a typo 🎣
```suggestion
It's recommended to prefix the module with `FungibleToken_` to avoid circuit signature clashes.
```

docs/modules/ROOT/pages/api/fungibleToken.adoc
andrew-fleming: > TBH, I have the same issue with how it is in Sol and Cairo, for example in [ERC20](https://docs.openzeppelin.com/contracts/5.x/api/token/erc20) if I am checking the transfer() from the IERC20 in order to check another function directly I have still to click again on IERC20, and then get back to that interface to see the rest of the functions and then click on the other function I want to check. Or I can just keep scrolling up or down until I see the function I want to check or I see IERC20Metadata section then I will know that the IERC20 section is finished so then either I click on the interface link on the right side again or scroll up again :) I don't think if I like this UX.

Haha yeahhh that's totally fair. It's clicky

> ...I would image the possibility to have 3rd degree listing on the right-side that appears under each section when it is opened, and then disappears again after it is closed...

I like this idea^

> I think maybe having sth like a config option in OZ docs where you can choose between some options in how would you like to see the docs, I think it is a good idea, like in Gmail for example you could choose between compact style or other styles.

This would be super cool^. I'm not sure how much work that would take and we'd also have to consider how much of a burden that would be for all the other teams (if any)

andrew-fleming: > In Ethereum everything is public by default (Unless for some applications which use some advanced Cryptography), But I think it is different in our case in Compact, it is important to distinguish in the docs between what is Public State handled by Ledger and what is Private State handled by Witness.

Mmm good point 👍

andrew-fleming: >> I don't disagree that the module deps should be documented in the API, but again, does this really need to be part of the sidebar?

> No, it does not need to be there. I think it might be nice to have. Maybe in sort of appendix section at the end of the page (And this section can be highlighted a bit differently like sort of more grayish). That appendix section could be also an option for your OZ Docs config options.

If there's a config option, then absolutely. And if we end up using the sidebar to list all sections and subsections, then I agree. Dependencies should be listed there as well

emnul (changes requested):
My only request is to add that warning I suggested 🙏

docs/modules/ROOT/pages/api/fungibleToken.adoc
emnul: Great conversation

> In Ethereum everything is public by default (Unless for some applications which use some advanced Cryptography), But I think it is different in our case in Compact, it is important to distinguish in the docs between what is Public State handled by Ledger and what is Private State handled by Witness.

This is an important point that we should definitely implement in the future.

> I think maybe having sth like a config option in OZ docs where you can choose between some options in how would you like to see the docs, I think it is a good idea, like in Gmail for example you could choose between compact style or other styles.

This new feature to the docs site is a great idea. We should open issues in the https://github.com/OpenZeppelin/docs.openzeppelin.com repo to get feedback from other teams. 

On Circuits / Internal Circuits sections

I think this is a good idea to let devs know that only the main contract should be able to access the internal circuits. They should not be made accessible in the entrypoint contract.

On Deps section:
I also think this is another important one to let devs know they need to add the dependency contract to their project directory manually or else they may be confused why the compiler is getting mad at them.

docs/modules/ROOT/pages/api/fungibleToken.adoc
```diff
@@ -0,0 +1,314 @@
...
+[.contract-item]
+[[FungibleTokenModule-_update]]
+==== `[.contract-item-name]#++_update++#++(from: Either<ZswapCoinPublicKey, ContractAddress>, to: Either<ZswapCoinPublicKey, ContractAddress>, value: Uint<128>) → []++` [.item-kind]#circuit#
+
+Transfers a `value` amount of tokens from `from` to `to`,
+or alternatively mints (or burns) if `from` (or `to`) is the zero address.
+
+Requirements:
+
+- Contract is initialized.
```
emnul, line 248: Non-blocking: This is the only circuit in the module that's not exported. We should explain why.

docs/modules/ROOT/pages/fungibleToken.adoc
```diff
@@ -0,0 +1,145 @@
...
+```typescript
+pragma language_version >= 0.15.0;
+
+import CompactStandardLibrary;
+import "./node_modules/@openzeppelin-compact/tokens/FungibleToken" prefix FungibleToken_;
+
+constructor(
+  name: Opaque<"string">,
+  symbol: Opaque<"string">,
+  decimals:Uint<8>
```
emnul, line 55: Non-blocking: Formatting fix
```suggestion
  decimals: Uint<8>
```

docs/modules/ROOT/pages/fungibleToken.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+The following example is a simple token contract with a fixed supply that's minted to the passed recipient upon construction.
+
+```typescript
+// FungibleTokenFixedSupply.compact
+
+pragma language_version >= 0.15.0;
+
+import CompactStandardLibrary;
+import "./node_modules/@openzeppelin-compact/tokens/FungibleToken" prefix FungibleToken_;
```
emnul, line 90: Great example! In the future we should add a non-trivial function to demonstrate proper usage of internal functions in the entrypoint contract.

andrew-fleming: Thank you! And for sure! When we get Ownable up, we'll have a lot more flexibility

docs/modules/ROOT/pages/api/fungibleToken.adoc
andrew-fleming: Good call!

emnul (approved):
yarn.unlocked

PR #190, opened 2025-07-23
Shielded AccessControl

### LINKED ISSUE #88: Add shielded access control
# Shielded Access Control Proposal

I aim to outline a Shielded Access Control module that would enable developers to define role-based access control mechanisms without publicly revealing the role assignments of accounts.

## Challenges

We can’t enforce the HKDF nonce derivation, since it’s computed and shared off-chain. Without circuit verification in Compact, users can use any nonce, as long as it’s unique for the PK-role pair to avoid commitment collisions. The HKDF method is a secure recommendation for randomness and privacy, but **enforcing it requires on-chain HKDF verification, which isn’t supported**. Users should follow this approach for best security. To address this challenge we should provide a HKDF nonce derivation implementation.

### Ledger

```typescript
// MerkleTree with 2^4 - 1 leaves
// Each leaf equals H(PK, role, nonce)
export ledger _roles: MerkleTree<4, Bytes<32>;

export ledger DEFAULT_ADMIN_ROLE: Bytes<32>;

```

### Private State

```typescript
// Generic role names to be modified by implementation
type RoleNames = "admin" | "minter" | "burner";

type RoleData = {
  [role in RoleNames]: {
    merkleTreeIndex: bigint; // Not sure about this one
    nonce: Uint8Array;
  };
};

type ShieldedAccessControlPrivateState = {
  readonly secretKey: Uint8Array;
  roles: RoleData
};
```

### External Circuits

```typescript
export circuit getAdminRole(role: Bytes<32>)


export circuit grantRole(pk: Bytes<32>, role: Bytes<32>, nonce: Bytes<16>): Bytes<32> {
    commitment = hashUserRole(PK || role || nonce)         
    Call _addRole(commitment)                        
    Return commitment
}

export circuit revokeRole(pk: Bytes<32>, role: Bytes<32>, nonce: Bytes<16>): [] {
    commitment = hashUserRole(PK || role || nonce)         
    Call _revokeRole(commitment)                        
    Return commitment
}

export circuit renounceRole(role: Bytes<32>, callerConfirmation: Either<ZswapCoinPublicKey, ContractAddress>): [] {
  assert callerConfirmation == left<ZswapCoinPublicKey,ContractAddress>(own_public_key()) "ShieldedAccessControl: bad confirmation";
  _revokeRole(role);
}

export circuit hasRole(role: Bytes<32>, account: Either<ZswapCoinPublicKey, ContractAddress>, nonce: Bytes<16>): Boolean

export circuit assertOnlyRole(role: Bytes<32>): [] {
}

/**
* @description creates public hash of secret key 
export circuit publicKey(sk: Bytes<32>, nonce: Bytes<16>): Bytes<32> {
 return persistent_hash<Vector<3, Bytes<32>>>([pad(32, "shieldedAccessControl:pk:"), pad(32, nonce) , sk]);
}
```

### Internal Circuits

```typescript
_grantRole(merkleTreeCommitment: Bytes<32>): Boolean
_revokeRole(merkleTreeCommitment: Bytes<32>): Boolean

```

### Witnesses

```typescript
/**
* @description Provide proof of knowledge of `role` - H(PK, role, nonce) - in MerkleTree
* Since nonce is only known locally and to the admin it's not possible 
*/ 
witness assertOnlyRole(role: Bytes<32>): MerkleTreePath<10, Bytes<32>>;

/**
* @description Getter for secret key
*/ 
witness localSecretKey();
```

### Off-chain Functions
```typescript
function HKDF(keyingMaterial, info, len)
function roleRequest(pubkey, role, nonce)

```

emnul:
## Types of changes

Closes #88 

## Summary

This pull request introduces a Shielded AccessControl module. Roles are stored as MerkleTree commitments to avoid disclosing information regarding the roles an account may have. Role commitments are created with using the following hashing scheme SHA256 ( role | accountId | instanceSalt | commitmentDomain ) where:
- `accountId` is SHA256(zcpk, nonce, instanceSalt, accountDomain)
- `role` is a unique `Bytes<32>` identifier
 - `instanceSalt` is an immutable, cryptographically strong random value provided on deployment
- `commitmentDomain` is the following zero padded 32 byte string: "ShieldedAccessControl:commitment"

In this RBAC model, role commitments behave like private bearer tokens. Possession of a valid, non-revoked role commitment grants authorization. Revocation permanently burns the role instance, requiring explicit new issuance under a new account identifier. Users must rotate their identity (new nonce) to be re-authorized. This creates stronger security invariants over traditional RBAC systems and enables privacy-preserving identity rotation.

### Privacy Assumptions

- Outside observers will know when an admin is added and how many admins exist.
- Outside observers will know which role identifiers are admin identifiers.
- Outside observers will have knowledge of all role identifiers.
- Outside observers will have knowledge of role additions and revocations.
- Outside observers will NOT be able to identify the public address of any role holder so long as secret nonce values are kept private and generated using cryptographically secure random values.

## Notes on Contract Design

- One of the goals of this Shielded Access Control implementation is to keep the API consistent with the Unshielded Access Control module. 
- Compact does not have a ZK friendly hashing function that can be used to derive state at this time, so the SHA256 hash function must be used. The usage of SHA256 incurs a large circuit size penalty on the contract which in turn slows compilation times.
- role commitments are tightly coupled to nullifiers. If an observer has knowledge of a role commitment they can immediately also compute its nullifier. All inputs that make up role commitments must be made public in this design. Thus, we cannot independently namespace a nullifier eg `nullifier = H(roleId, accountId, instanceSalt, nullifierDomain)`. With this in mind, we can save on circuit size by deriving nullifiers as `nullifier = H(roleCommitment, nullifierDomain)`. Future designs should re-evaluate this tight coupling if inputs become more private.

#### Why use an instance salt instead of a contract address for instance separation in the role commitment scheme?

The a contract address can only referenced after the contract is successfully deployed. Thus, calling the `_grantRole` circuit in a contract constructor will fail requiring a two step initialization process that is error prone and unsafe. Using an instance salt value avoids failed deployments due to this unique condition and allows roles to be completely configured before the contract is live in a production environment.

#### Why is the Merkle tree append only?
Modifying or deleting role commitments from the Merkle tree may invalidate legitimate roles that exist along the path from a leaf to the root breaking functionality. Furthermore, this behavior leaks information allowing an observer to determine the _some_ data was removed, approximate location or count of removed entries, and correlate changes with user actions or events. In order to minimize information leakage and avoid a complex Merkle tree storage management solution that would inevitably conflict with the goals of this module (maintaining the privacy of user roles), an append only solution was chosen.

#### Why re-compute commitments and nullifiers in `_grantRole` and `_revokeRole`?
Recomputing these values minimizes the scope of what must be exposed publicly. While it comes at a significant performance cost, it maximizes the allowable privacy of the module.

## Open Questions

- Should the `initialize` circuit assert that the `instanaceSalt` is not 0?
- Is 2^20 large enough for the ledger MerkleTrees? Should it be larger?
- Should `_computeXXX` circuits be exported? Exporting these circuits increases the risk that a module consumer would build flows that leak this metadata and increase linkability 

## Future Improvements

- Unlinkability over time can be improved using an `_epoch` ledger variable and a rotation circuit `export circuit rotateEpoch(): []` restricted to admin or governance
- Unlinkability can be improved using an **Invite/claim flow**: admin publishes an “issuance note” commitment; user later claims it privately by proving knowledge of the note secret, then the contract updates _operatorRoles.
	- This prevents observers from linking admin action → recipient identity commitment.
- Use ZK-friendly hashes when a stable hashing algorithm is available in the language

## Release Notes

* **New Features**
  * Added ShieldedAccessControl system providing privacy-preserving role-based access management with cryptographic commitments and revocation tracking.

* **Tests**
  * Added comprehensive test suites and mock implementations for access control modules.

* **Documentation**
  * Updated simulator documentation with generic ledger type examples.

* **Chores**
  * Added production safety warnings to test mock contracts.
  * Enhanced witness interfaces with improved generic ledger type support.


emnul:
I'm not sure how to approach role invalidation. There's no easy way to clear role commitments from the MT. Here are some ideas I had:

1. Use a `Map<Bytes<32>, Uint<64>` ledger state variable to track where a roleCommitment is in the MT. Use a  `Counter` ledger state variable to track current MT index . Use `InsertDefault()` MT to clear roleCommitment from the MT on `revokeRole`.
2. Use a Nullifier set to invalidate roleCommitments. On `revokeRole`, add roleCommitment to Nullifier set. `hasRole` will check if roleCommitment is both in the MT and not in nullifier set. `grantRole` will remove roleCommitment from nullifier set if already in MT.

I would like your thoughts, I keep going back and forth between these two designs. In my opinion, option 2 feels like a more straightforward solution without adding too much complexity

emnul:
> When you say to clear the commitment from the MT, you mean to literally change (delete) the leaf, right? I think this will be tricky bc we'd have to update the root whenever we revoke roles and this will break verification for other commitments that rely on the root being the same.

Yeah pretty much just zeroing out the leaf. I also thought we might run into trouble with the root changing from my general knowledge on how MTs work, but wasn't sure since I haven't played around with the data structure to examine how it works. In practice, I'm sure it changes the root like you mentioned.

> I imagine we'd need a mechanism that logs previous roots for previous commitments (and/or use an epoch scheme)—compact has HistoricMerkleTree to check against previous roots, but then we run into the issue of false positives as mentioned in the docs:

Yup, not my favorite approach. Super error prone and overly complex. Using the MT in an "append only" way makes it much simpler

> If we remove the commitment from the nullifier set, won't this link the new roleCommitment with the removed nullifier? We can maybe use something like a nonce (or timestamp when available) instead to avoid deleting the nullifier (if you agree)

From what I understand, if we use the commitment generation scheme described then a `roleCommitment` should be deterministic. Thus, we'd never need to add to the MT more than once for a role unless the `nonce` is lost. Definitely check my logic here tho

emnul:
> Yes, you're correct, but this still provides a link between the removed nullifier and the existing commitments. Not exactly home address + phone number-level data, but data nonetheless

That makes sense. My thinking was we could save on storage by re-using commitments, but I see now this comes with some increased risk of leaking data. We should definitely prioritize shielding data vs optimizing circuits / storage.

> IMO if a user is granted and revoked the same role 3 times, there should be three different commitments and nullifiers.

I plan on moving forward this scheme

emnul:
Nonce values cannot be cast from `Bytes<16>` to `Bytes<32>` so I'm thinking we change its type to `Bytes<32>`

emnul:
@andrew-fleming what do we think about introducing a Counter ledger variable  and a Mapping from a roleCommitment Bytes<32> to a Counter value Uint<64>? 

The idea here is to use the Counter to satisfy the uniqueness property we want for each roleCommitment. The system can be more trustless in that way. Alternatively, we could use a salt value when generating the nonce, but there's no way to enforce that on the client side.

andrew-fleming:
@emnul I think this could work. Another idea in a similar vein would be to use the MT index (assuming this is append-only) and just have a mapping for `commitment => index`

Whatever design we end up using, consider too that there's probably room for different flavors—each with their own tradeoffs. If there's enough flexibility with just one then that works too :)

emnul:
>  I think this could work. Another idea in a similar vein would be to use the MT index (assuming this is append-only) and just have a mapping for commitment => index
> 
> Whatever design we end up using, consider too that there's probably room for different flavors—each with their own tradeoffs. If there's enough flexibility with just one then that works too :)

Ahhh this cant work because we would need to convert the Uint<64> value to a Bytes<32> value to hash everything together but this isn't possible in compact at this time 🥲

andrew-fleming (commented):
Good start @emnul! I left some comments.

I'm also wondering if there's a more modular approach that we can use where HKDF and the like can be used as extensions instead of direct recommendations. We can always change the module later, so I suppose it's okay for it to be opinionated now. I'll give a deeper review when there are tests as I believe there are some issues we need to take care of that would have been caught already

Another small note: sending the pk and nonce to the admin IMO creates a shift in expectations regarding how this model works. The idea is that there's user intent to have a role, right? There's nothing to stop an admin from logging the pk and nonce for everyone. When roles are revoked from users, an admin can replay those values. I imagine this _could_ be an issue with some protocols. The model follows a user-intent design, but that assumption doesn't hold if an admin can replay. Adding a timestamp to the nonce hash could prevent this

contracts/src/access/witnesses/ShieldedAccessControlWitnesses.ts
```diff
@@ -0,0 +1,228 @@
...
+    {
+      ledger,
+      privateState,
+    }: WitnessContext<Ledger, ShieldedAccessControlPrivateState>,
+    roleId: Uint8Array,
+    account: Either<ZswapCoinPublicKey, ContractAddress>
+  ): bigint => {
+    const roleIdString = Buffer.from(roleId).toString('hex');
+    // Iterate over each MT index to determine if commitment exists
+    for (let i = 0; i < MERKLE_TREE_DEPTH; i++) {
```
andrew-fleming, line 155:
```suggestion
    for (let i = 0; i <= ledger.ShieldedAccessControl__currentMerkleTreeIndex; i++) {
```

andrew-fleming: So we're not iterating 34945023502 items each time

andrew-fleming: Or 1024 🤷‍♂️

contracts/src/access/witnesses/ShieldedAccessControlWitnesses.ts
```diff
@@ -0,0 +1,228 @@
...
+        bNonce,
+        bIndex,
+        DOMAIN,
+      ]);
+      try {
+        ledger.ShieldedAccessControl__operatorRoles.pathForLeaf(
+          BigInt(i),
+          commitment,
+        );
+        return BigInt(i);
```
andrew-fleming, line 172:
```suggestion
        const helloDarknessMyOldFriend = ledger.ShieldedAccessControl__operatorRoles.pathForLeaf(
          BigInt(i),
          commitment,
        );
        if (helloDarknessMyOldFriend.leaf === commitment) {
          return BigInt(i);
        }
```

contracts/src/access/test/ShieldedAccessControl.test.ts
```diff
@@ -0,0 +1,991 @@
...
+      ['grantRole', false, true, false, [DEFAULT_ADMIN_ROLE, Z_ADMIN]],
+      ['grantRole', false, false, false, [DEFAULT_ADMIN_ROLE, Z_ADMIN]],
+      ['revokeRole', false, true, true, [DEFAULT_ADMIN_ROLE, Z_ADMIN]],
+      ['revokeRole', true, false, true, [DEFAULT_ADMIN_ROLE, Z_ADMIN]],
+      ['revokeRole', true, true, false, [DEFAULT_ADMIN_ROLE, Z_ADMIN]],
+      ['revokeRole', false, false, true, [DEFAULT_ADMIN_ROLE, Z_ADMIN]],
+      ['revokeRole', true, false, false, [DEFAULT_ADMIN_ROLE, Z_ADMIN]],
+      ['revokeRole', false, true, false, [DEFAULT_ADMIN_ROLE, Z_ADMIN]],
+      ['revokeRole', false, false, false, [DEFAULT_ADMIN_ROLE, Z_ADMIN]],
+    ];
```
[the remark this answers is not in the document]
andrew-fleming: This def works though if you want to keep it. Let's get everything working first and then we can laser in on the testing deets

contracts/src/access/test/ShieldedAccessControl.test.ts
```diff
@@ -0,0 +1,991 @@
...
+    });
+
+    it('should not fail when authorized caller has correct nonce, index, and path', () => {
+      // Check nonce is correct
+      expect(
+        shieldedAccessControl.privateState.getCurrentSecretNonce(
+          DEFAULT_ADMIN_ROLE,
+        ),
+      ).toBe(ADMIN_SECRET_NONCE);
+
```
andrew-fleming, line 709: I think there's something wrong because this passes:

```
shieldedAccessControl.callerCtx.setCaller('maximus decimus meridius');
shieldedAccessControl.assertOnlyRole(new Uint8Array(32).fill(123456789));
```

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,450 @@
...
+   * Internal circuit without access restriction.
+   *
+   * @circuitInfo k=17, rows=108916
+   *
+   * Requirements:
+   *
+   * - An index for the intermediate role commitment produced by SHA256(roleId | account | nonce)
+   * must exist in the `_roleCommitmentIndex` map.
+   * - A nullifier for the role commitment produced by SHA256(roleId | account | nonce)
+   * must not exist in the `_roleCommitmentNullifiers` set.
```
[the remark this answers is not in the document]
andrew-fleming: https://docs.midnight.network/develop/how-midnight-works/keeping-data-private#the-commitmentnullifier-pattern

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,450 @@
...
+   * @param {Bytes<32>} roleId - The role identifier.
+   * @param {Either<ZswapCoinPublicKey, ContractAddress>} account - The account to check.
+   * @param {Bytes<32>} nonce - A nonce created using SHA256(SK | "role-nonce" | role | PK).
+   * @return {Boolean} - A boolean determining if the account has the specified role.
+   */
+  export circuit hasRole(roleId: Bytes<32>, account: Either<ZswapCoinPublicKey, ContractAddress>): Role {
+    assert(!Utils_isContractAddress(account), "ShieldedAccessControl: contract address roles are not yet supported");
+
+    const nonce = wit_secretNonce(roleId);
+    const index = wit_getRoleIndex(roleId, account);
```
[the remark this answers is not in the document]
emnul: I definitely think a `callerHasRole` circuit is a great addition, but `hasRole`, in my conceptualization of this system, would be useful for admin users and I think still deserves a place in the contract. The name could probably be improved to make it clear the caller has privileged access to user info.

contracts/src/access/test/ShieldedAccessControl.test.ts
emnul: Hmmm, I'm seeing it fail due to Compact address formatting issues unrelated to the contract itself

andrew-fleming: My bad, use this:

```
const [maximus, z_maximus] = utils.generateEitherPubKeyPair('maximus decimus meridius');

shieldedAccessControl.callerCtx.setCaller(maximus);
shieldedAccessControl.assertOnlyRole(new Uint8Array(32).fill(123456789));
```

lmk if I'm mistaken

0xisk (dismissed):
Thank you @emnul, that’s a partial review — I left some comments.

* (_From my experince in the client-side_) Midnight’s key management has two main keys: `ZswapCoinPublicKey` and `EncPublicKey`. Since inside Compact we only deal with `ZswapCoinPublicKey`, I think it’s best to explicitly call it `zcpk` or `zpk` instead of just `pk`. This should help prevent any confusion for client-side developers who might otherwise mistake it for `EncPublicKey`.
* If that makes sense, then to avoid further confusion, I’d also suggest unifying the terminology where `OwnerId` and `AccountId` are currently used interchangeably. We could instead consistently use `zcpk` for the raw key and `shieldedZcpk` for the derived identifier (`SHA256(zcpk, nonce)`).

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,537 @@
...
+   */
+  export circuit _revokeRole(roleId: Bytes<32>, accountId: Bytes<32>): Boolean {
+    const role = getRole(roleId, accountId);
+    if (!role.isApproved) {
+      return false;
+    }
+
+    _roleCommitmentNullifiers.insert(disclose(role.commitmentNullifier));
+    return true;
+  }
```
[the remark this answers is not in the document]
emnul: I had thought about doing this initially, but I could not come up with a Merkle tree management solution that wasn't overly complex or did not add significant ledger storage / circuit size overhead.

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,537 @@
...
+  import CompactStandardLibrary;
+  import "ShieldedAccessControlUtils" prefix Utils_;
+
+  /**
+   * @description A Merkle tree of role commitments stored as SHA256(roleId | account | nonce | merkleTreeIndex)
+   * @type {Bytes<32>} roleCommitment - A role commitment created by the following hash: SHA256(roleId | account | nonce | merkleTreeIndex).
+   * @type {MerkleTree<10, roleCommitment>}
+   * @type {MerkleTree<10, Bytes<32>>} _operatorRoles
+   */
+  export ledger _operatorRoles: MerkleTree<10, Bytes<32>>;
```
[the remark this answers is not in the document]
emnul: It was an arbitrarily large number I thought would be sufficient for most use cases. It's not possible to dynamically determine the tree size nor would that be desirable considering the additional initialization overhead every circuit would need.

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,626 @@
...
+  import CompactStandardLibrary;
+  import "../utils/Utils" prefix Utils_;
+  import "../security/Initializable" prefix Initializable_;
+
+  export new type RoleCommitment = Bytes<32>;
+  export new type RoleIdentifier = Bytes<32>;
+  export new type AccountIdentifier = Bytes<32>;
+  export new type AdminIdentifier = Bytes<32>;
+  export new type RoleNullifier = Bytes<32>;
+  type ZcpkOrContractAddress = Either<ZswapCoinPublicKey, ContractAddress>;
```
0xisk: I'd not have this here, as it is more of the generic type that can be used by other modules too.

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,626 @@
...
+ *  - `instanceSalt` is an immutable, cryptographically strong random value provided on deployment
+ *  - `commitmentDomain` is the following zero padded 32 byte string: "ShieldedAccessControl:commitment"
+ *
+ * In this RBAC model, role commitments behave like private bearer tokens. Possession of a valid, non-revoked role
+ * commitment grants authorization. Revocation permanently burns the role instance, requiring explicit new issuance
+ * under a new account identifier. Users must rotate their identity (new nonce) to be re-authorized. This creates
+ * stronger security invariants over traditional RBAC systems and enables privacy-preserving identity rotation.
+ *
+ * @notice Using the SHA256 hashing function comes at a significant performance cost. In the future, we
+ * plan on migrating to a ZK-friendly hashing function when an implementation is available.
```
0xisk:
```suggestion
```

Can we move this at the end with the other `@notice`?

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,626 @@
...
+ *
+ * Roles are referred to by their `Bytes<32>` identifier. These should be exposed
+ * in the top-level contract and be unique. One way to achieve this is by
+ * using `export sealed ledger` hash digests that are initialized in the top-level contract:
+ *
+ * ```compact
+ * import CompactStandardLibrary;
+ * import "./node_modules/@openzeppelin-compact/accessControl/src/ShieldedAccessControl" prefix ShieldedAccessControl_;
```
0xisk:
```suggestion
 * import "./node_modules/@openzeppelin/compact-contracts/src/access/ShieldedAccessControl" prefix ShieldedAccessControl_;
```

0xisk (commented):
Thank you @emnul, had a partial review. Left some comments and suggestions. Also I think all the `@circuitInfo` are outdated.

0xisk (commented):
Thanks @emnul! that's a final review for the SAC contract AFAICT, next one will be on the tests.

emnul:
I didn't like the dependencies on untrusted witness values in the `_revokeRole` and `_grantRole` circuits so the role commitment existence check in each of those circuits have been removed. This also significantly simplifies the code and expectations.

contracts/src/access/test/mocks/MockShieldedAccessControl.compact
0xisk: sorry for the nit! the order of the circuits here does not follow the base contract's order, lets fix the order here and therefore that would require fix to the simulator and the test.

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,667 @@
...
+ *     into `_operatorRoles` on grant. The `instanceSalt` prevents commitment collisions across
+ *     deployments that share the same role identifiers.
+ *  - `accountId` is a privacy-preserving identity commitment. `zcpk` is the user's
+ *     `ZswapCoinPublicKey`; `nonce` is a per-role secret held in local private state
+ *     (supplied by `wit_secretNonce`); `instanceSalt` ensures the same key and nonce
+ *     cannot be correlated across contracts.
+ *  - `roleNullifier` is a one-time burn token inserted into `_roleCommitmentNullifiers` on
+ *     revocation. Its presence permanently invalidates the corresponding role commitment,
+ *     making re-grant under the same `accountId` impossible without generating a new identity.
+ *  - `instanceSalt` is an immutable, cryptographically strong random value provided on deployment
```
andrew-fleming: I'd rephrase this a teeny bit. We can't guarantee that it is a cryptographically strong random value. Maybe it's better to say "it _should_ be a..."

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,667 @@
...
+ *  - `commitmentDomain` is the following zero padded 32 byte string: "ShieldedAccessControl:commitment"
+ *  - `accountIdDomain` is the following zero padded 32 byte string: "ShieldedAccessControl:accountId"
+ *  - `commitmentDomain` is the following zero padded 32 byte string: "ShieldedAccessControl:nullifier"
```
andrew-fleming:
```suggestion
 *  - `nullifierDomain` is the following zero padded 32 byte string: "ShieldedAccessControl:nullifier"
```

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,667 @@
...
+  import CompactStandardLibrary;
+  import "../utils/Utils" prefix Utils_;
+  import "../security/Initializable" prefix Initializable_;
+
+  // TODO: Standardize types across contracts https://github.com/OpenZeppelin/compact-contracts/issues/368
+  export new type RoleCommitment = Bytes<32>;
+  export new type RoleIdentifier = Bytes<32>;
+  export new type AccountIdentifier = Bytes<32>;
+  export new type AdminIdentifier = Bytes<32>;
+  export new type RoleNullifier = Bytes<32>;
```
andrew-fleming: Objectively, this looks good. Practically, I'm not convinced it's pulling as much weight as it seems. Every time this type is calculated and passed, it's already bound to a variable of (roughly) the same name. I'll say that it's helpful as a type in the maps, though...tbc in the referenced issue. I'm good with it in this module

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,667 @@
...
+   *
+   * @circuitInfo k=9, rows=375
+   *
+   * Disclosures:
+   *
+   * - A role identifier.
+   *
+   * @param {Bytes<32>} role - The role identifier.
+   *
+   * @return {Bytes<32>} roleAdmin - The admin role that controls `role`.
```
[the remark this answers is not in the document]
andrew-fleming: This applies to the rest of the circuits if you agree

andrew-fleming (commented):
Partial review so I can respond to threads in real time

contracts/src/access/test/mocks/MockShieldedAccessControl.compact
```diff
@@ -0,0 +1,203 @@
...
+            ShieldedAccessControl__instanceSalt,
+            pad(32, "ShieldedAccessControl:commitment")]
+           )
+         as ShieldedAccessControl_RoleCommitment;
+}
+
+// circuit is reimplemented in the Mock contract for testing
+export circuit _computeAccountId(
+                 role: ShieldedAccessControl_RoleIdentifier
+                 ): ShieldedAccessControl_AccountIdentifier {
```
andrew-fleming: This is a super risky practice. Changes to the module that are not reflected in the mock can silently diverge. This would at the very least require a loud note somewhere and even that is just a duct tape solution with a warning. Thoughts?

andrew-fleming: What do you think about prefixing `_unsafe` to these circuits and document the risks? Either way, users will need to calculate their accountId, right?

contracts/src/access/witnesses/ShieldedAccessControlWitnesses.ts
```diff
@@ -0,0 +1,131 @@
...
+  },
+
+  setRole: (
+    privateState: ShieldedAccessControlPrivateState,
+    role: Buffer,
+    nonce: Buffer,
+  ): ShieldedAccessControlPrivateState => {
+    const roleString = role.toString('hex');
+    privateState.roles[roleString] = nonce;
+    return privateState;
```
andrew-fleming:
```suggestion
    const roleString = role.toString('hex');
    const roles: Record<string, Uint8Array> = {};

    for (const [k, v] of Object.entries(privateState.roles)) {
      roles[k] = new Uint8Array(v);
    }

    roles[roleString] = new Uint8Array(nonce);
    return { roles }
```
Deep state copy. Avoids mutating the ps input

andrew-fleming (commented):
Nice work, @emnul! The module has good bones 🦴  I left some comments, questions, and suggestions. This is a partial review, I still have to tackle tests

contracts/src/access/test/mocks/MockShieldedAccessControl.compact
emnul: > What do you think about prefixing _unsafe to these circuits and document the risks?

I'm thinking we should add a super loud "DO NOT USE THIS CODE IN PRODUCTION ITS ONLY FOR TESTING" or something like that to all of our mock contracts and maybe somewhere in /contracts or our README

andrew-fleming: > This practice has already hugely backfired on me

Oh nooo 😢 

> Unfortunately, it seems we have no other choice but to deal with this anti-pattern or perhaps build our own tooling to work around it for now.

Our own tooling would be an interesting challenge. We still have to build the autogen simulator and mock

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,795 @@
...
+    const roleNullifier = _computeNullifier(roleCommitment);
+    const isRevoked = _roleCommitmentNullifiers.member(disclose(roleNullifier));
+
+    if (isRevoked) {
+      return false;
+    }
+
+    _roleCommitmentNullifiers.insert(disclose(roleNullifier));
+    return true;
+  }
```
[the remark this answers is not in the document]
emnul: > The only downside is that the proof gen will include both branches on the if/else. I'm not sure how expensive that will be.

I didn't see any impact on rows / cols and I like that we can shrink the footprint a good deal so I'm adding `_updateRole` in

contracts/README.md
andrew-fleming: I'd propose adding the contracts/README into another PR

contracts/README.md
```diff
@@ -0,0 +1,93 @@
...
+├── src/                    # Source files
+│   ├── access/             # Access control contracts
+│   ├── security/           # Security utility contracts
+│   ├── token/              # Token standard contracts
+│   ├── utils/              # General utility contracts
+│   ├── archive/            # Archived/deprecated contracts
+│   └── test-utils/         # Shared test helpers
+├── artifacts/              # Compiled contract outputs (generated)
+└── dist/                   # Compiled TypeScript witness outputs (generated)
+```
```
andrew-fleming: Personally, not a huge fan of including a structured layout of directories like this. Every time we add/remove/change something, we need to remember to update this and IMO it doesn't add much value. Though, @0xisk also proposed adding this in the contracts section of midnight-apps repo (if I recall correctly). I'm outnumbered if y'all both like it

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,795 @@
...
+   *
+   * @param {RoleCommitment} roleCommitment - The role commitment for a particular `(role, accountId)` pairing.
+   *
+   * @returns {RoleNullifier} roleNullifier - The associated nullifier for `roleCommitment`.
+   */
+  pure circuit _computeNullifier(roleCommitment: RoleCommitment): RoleNullifier {
+    return persistentHash<Vector<2, Bytes<32>>>(
+             [roleCommitment as Bytes<32>, pad(32, "ShieldedAccessControl:nullifier")]
+             )
+           as RoleNullifier;
```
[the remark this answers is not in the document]
emnul: The original idea was to make it more difficult to enable an enumeration attack on `(role | accountId)` pairings, but if the contracts are open source anyway this doesn't really make sense. Any attacker with moderate technical ability would be able to compute these values locally, so keeping these util circuits internal degrades UX without any real security benefit.

contracts/src/access/ShieldedAccessControl.compact
```diff
@@ -0,0 +1,757 @@
...
+   */
+  export ledger _roleCommitmentNullifiers: Set<RoleNullifier>;
+
+  /**
+   * @sealed @ledger _instanceSalt
+   * @description A per-instance value provided at initialization used to namespace
+   * commitments for this contract instance.
+   *
+   * This salt prevents commitment collisions across contracts that might otherwise use
+   * the same identifiers or domain parameters. It should be should be a cryptographically strong random value
```
0xisk:
```suggestion
   * the same identifiers or domain parameters. It should be a cryptographically strong random value
```

0xisk (commented):
Looking good @emnul left some comments.

pepebndc (approved):

PR #196, opened 2025-07-29
Fix angle brackets in docsite

andrew-fleming:
The docsite misinterpreted `Uint<128>` as `Uint(128)`


emnul (approved):
Interesting behavior, I'll watch out for this in the future!

andrew-fleming:
> Interesting behavior, I'll watch out for this in the future!

For sure, it's sneaky 🥷

PR #204, opened 2025-08-11
Add ownable shielded

### LINKED ISSUE #40: Ownable shielded

andrew-fleming:
## Types of changes

Fixes #40. Supercedes #61.

## PR Checklist

- [x] I have read the [Contributing Guide](CONTRIBUTING.md)
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added documentation of new methods and any new behavior or changes to existing behavior
- [x] CI Workflows Are Passing

### Overview

This PR proposes to add the ZOwnablePK module which is a variant of the transparent Ownable module. ZOwnablePK allows an account to perform administrative tasks on a contract (circuits protected with `assertOnlyOwner`) while also never revealing the account's public key. The foundation of this design is the user's identifier which is a hash of the user's public key and secret nonce (stored in the user's private state). This design does not enforce how the nonce is derived; therefore, users may choose to make the nonce deterministic and thus retrievable if it is lost. 

### Ownership transfers

Since public keys are neither stored nor revealed, transferring ownership requires off chain communication. In order for Alice to transfer ownership to Bob, Alice needs Bob's id (`H(pk, secretNonce)`). This handshake approach ensures that even in the most adversarial conditions, Alice and Bob are not required to reveal their respective public keys to each other

### Ownership commitment

The ownership commitment is the only user identifier stored on the ledger. The commitment consists of:
- `id`: The privacy-preserving owner identifier -> `H(pk, secretNonce)`.
- `instanceSalt`: A unique per-deployment salt that prevents commitment collisions across different contract instances, even when the same owner and nonce are used (though, it's strongly recommended to rotate the secretNonce after it's been used)
- `counter`: Incremented with each ownership transfer to ensure unlinkability—each transfer produces a completely different commitment even with the same underlying owner.
- `pad(32, "ZOwnablePK:shield:")`: A domain separator padded to 32 bytes that prevents hash collisions with other commitment schemes and enables safe protocol extensions.

### DIFF
--- a/contracts/src/access/ZOwnablePK.compact
+++ b/contracts/src/access/ZOwnablePK.compact
@@ -1,0 +1,310 @@
+// SPDX-License-Identifier: MIT
+
+pragma language_version >= 0.16.0;
+
+/**
+ * @module ZOwnablePK
+ * @description A shielded, public key-derived Ownable module.
+ *
+ * `ZOwnablePK` provides a privacy-preserving access control mechanism
+ * for contracts with a single administrative user. Unlike traditional
+ * `Ownable` implementations that store or expose the owner's public key
+ * on-chain, this module stores only a commitment to a hashed identifier
+ * derived from the owner's public key and a secret nonce.
+ * For the strongest security guarantees, use an Air-Gapped Public Key.
+ *
+ * @notice This module explicitly supports commitments derived from public keys;
+ * however, it may be possible to use contract addresses when contract-to-contract
+ * calls become available. This will be revisited when it's known if/how witnesses
+ * are used from a contract address context.
+ *
+ * @dev Features:
+ * - Obfuscated owner identity: The owner's public key is never revealed on-chain.
+ * - Stateless verification: The contract never needs access to the full public key.
+ * - Built-in support for transfer and renounce functionality.
+ * - Instance-specific salts to prevent cross-contract correlation.
+ * - Deterministic hashing with `persistentHash` to support zero-knowledge verification.
+ *
+ * @dev Commitment structure:
+ * ```
+ * id = SHA256(pk, secretNonce)
+ * commitment = SHA256(id, instanceSalt, counter, "ZOwnablePK:shield:")
+ * ```
+ * The commitment changes on each transfer due to the incrementing `counter`,
+ * providing unlinkability across ownership changes.
+ *
+ * @dev Security Considerations:
+ * - The `secretNonce` must be kept private. Loss of the nonce prevents the
+ *   owner from proving ownership or transferring it.
+ * - Ownership validation is entirely circuit-based using witness-provided values.
+ * - The `_instanceSalt` is immutable and used to differentiate deployments.
+ *
+ * @notice Best used for single-admin contracts with privacy requirements.
+ * It is not designed for multi-owner or role-based access control.
+ */
+module ZOwnablePK {
+  import CompactStandardLibrary;
+  import "../security/Initializable" prefix Initializable_;
+
+  /**
+   * @ledger _ownerCommitment
+   * @description Stores the current hashed commitment representing the owner.
+   * This commitment is derived from the public identifier (e.g., `SHA256(pk, nonce)`),
+   * the `instanceSalt`, the transfer `counter`, and a domain separator.
+   *
+   * A commitment of `default<Bytes<32>>` (i.e., zero) indicates the contract is unowned.
+   */
+  export ledger _ownerCommitment: Bytes<32>;
+  /**
+   * @ledger _counter
+   * @description Internal transfer counter used to prevent commitment reuse.
+   *
+   * Increments by 1 on every successful ownership transfer. Combined with `id` and
+   * `instanceSalt` to compute unique owner commitments over time.
+   */
+  export ledger _counter: Counter;
+  /**
+   * @sealed @ledger _instanceSalt
+   * @description A per-instance value provided at initialization used to namespace
+   * commitments for this contract instance.
+   *
+   * This salt prevents commitment collisions across contracts that might otherwise use
+   * the same owner identifiers or domain parameters. It is immutable after initialization.
+   */
+  export sealed ledger _instanceSalt: Bytes<32>;
+
+  /**
+   * @witness wit_secretNonce
+   * @description A private per-user nonce used in deriving the shielded owner identifier.
+   *
+   * Combined with the user's public key as `SHA256(pk, nonce)` to produce an obfuscated,
+   * unlinkable identity commitment. Users are encouraged to rotate this value on ownership changes.
+   */
+  export witness wit_secretNonce(): Bytes<32>;
+
+  /**
+   * @description Initializes the contract by setting the initial owner via `ownerId`
+   * and storing the `instanceSalt` that acts as a privacy additive for preventing
+   * duplicate commitments among other contracts implementing ZOwnablePK.
+   *
+   * @warning The `ownerId` must be calculated prior to contract deployment using the SHA256 hashing algorithm.
+   * Using any other algorithm will result in a permanent loss of contract access.
+   *
+   * @circuitInfo k=14, rows=14933
+   *
+   * Requirements:
+   *
+   * - Contract is not initialized.
+   * - `ownerId` is not zero.
+   *
+   * @param {Bytes<32>} ownerId - The owner's unique identifier SHA256(pk, nonce).
+   * @param {Bytes<32>} instanceSalt - Contract salt to prevent duplicate commitments if
+   * users reuse their PK and secretNonce witness (not recommended).
+   * @returns {[]} Empty tuple.
+   */
+  export circuit initialize(ownerId: Bytes<32>, instanceSalt: Bytes<32>): [] {
+    Initializable_initialize();
+
+    assert(ownerId != default<Bytes<32>>, "ZOwnablePK: invalid id");
+    _instanceSalt = disclose(instanceSalt);
+    _transferOwnership(ownerId);
+  }
+
+  /**
+   * @description Returns the current commitment representing the contract owner.
+   * The full commitment is: `SHA256(SHA256(pk, nonce), instanceSalt, counter, domain)`.
+   *
+   * @circuitInfo k=10, rows=57
+   *
+   * Requirements:
+   *
+   * - Contract is initialized.
+   *
+   * @returns {Bytes<32>} The current owner's commitment.
+   */
+  export circuit owner(): Bytes<32> {
+    Initializable_assertInitialized();
+    return _ownerCommitment;
+  }
+
+  /**
+   * @description Transfers ownership to `newOwnerId`.
+   * `newOwnerId` must be precalculated and given to the current owner off chain.
+   *
+   * @circuitInfo k=16, rows=39240
+   *
+   * Requirements:
+   *
+   * - Contract is initialized.
+   * - Caller is the the current owner.
+   * - `newOwnerId` is not an empty array.
+   *
+   * @param {Bytes<32>} newOwnerId - The new owner's unique identifier (`SHA256(pk, nonce)`).
+   * @returns {[]} Empty tuple.
+   */
+  export circuit transferOwnership(newOwnerId: Bytes<32>): [] {
+    Initializable_assertInitialized();
+
+    assertOnlyOwner();
+    assert(newOwnerId != default<Bytes<32>>, "ZOwnablePK: invalid id");
+    _transferOwnership(newOwnerId);
+  }
+
+  /**
+   * @description Leaves the contract without an owner.
+   * It will not be possible to call `assertOnlyOnwer` circuits anymore.
+   * Can only be called by the current owner.
+   *
+   * @circuitInfo k=15, rows=24442
+   *
+   * Requirements:
+   *
+   * - Contract is initialized.
+   * - Caller is the the current owner.
+   *
+   * @returns {[]} Empty tuple.
+   */
+  export circuit renounceOwnership(): [] {
+    Initializable_assertInitialized();
+
+    assertOnlyOwner();
+    _ownerCommitment.resetToDefault();
+  }
+
+  /**
+   * @description Throws if called by any account whose id hash `SHA256(pk, nonce)` does not match
+   * the stored owner commitment.
+   * Use this to only allow the owner to call specific circuits.
+   *
+   * @circuitInfo k=15, rows=24437
+   *
+   * Requirements:
+   *
+   * - Contract is initialized.
+   * - Caller's id (`SHA256(pk, nonce)`) when used in `_computeOwnerCommitment` equals
+   * the stored `_ownerCommitment`, thus verifying themselves as the owner.
+   *
+   * @returns {[]} Empty tuple.
+   */
+  export circuit assertOnlyOwner(): [] {
+    Initializable_assertInitialized();
+
+    const nonce = wit_secretNonce();
+    const callerAsEither = Either<ZswapCoinPublicKey, ContractAddress> {
+      is_left: true,
+      left: ownPublicKey(),
+      right: ContractAddress { bytes: pad(32, "") }
+    };
+    const id = _computeOwnerId(callerAsEither, nonce);
+    assert(_ownerCommitment == _computeOwnerCommitment(id, _counter), "ZOwnablePK: caller is not the owner");
+  }
+
+  /**
+   * @description Computes the owner commitment from the given `id` and `counter`.
+   *
+   * ## Owner ID (`id`)
+   * The `id` is expected to be computed off-chain as:
+   * `id = SHA256(pk, nonce)`
+   *
+   * - `pk`: The owner's public key.
+   * - `nonce`: A secret nonce scoped to the instance, ideally rotated with each transfer.
+   *
+   * ## Commitment Derivation
+   * `commitment = SHA256(id, instanceSalt, counter, domain)`
+   *
+   * - `id`: See above.
+   * - `instanceSalt`: A unique per-deployment salt, stored during initialization.
+   * This prevents commitment collisions across deployments.
+   * - `counter`: Incremented with each ownership transfer, ensuring uniqueness
+   * even with repeated `id` values. Cast to `Field` then `Bytes<32>` for hashing.
+   * - `domain`: Domain separator `"ZOwnablePK:shield:"` (padded to 32 bytes) to prevent
+   * hash collisions when extending the module or using similar commitment schemes.
+   *
+   * @circuitInfo k=14, rows=14853
+   *
+   * Requirements:
+   *
+   * - Contract is initialized.
+   *
+   * @param {Bytes<32>} id - The unique identifier of the owner calculated by `SHA256(pk, nonce)`.
+   * @param {Uint<64>} counter - The current counter or round. This increments by `1`
+   * after every transfer to prevent duplicate commitments given the same `id`.
+   * @returns {Bytes<32>} The commitment derived from `id` and `counter`.
+   */
+  export circuit _computeOwnerCommitment(
+    id: Bytes<32>,
+    counter: Uint<64>,
+  ): Bytes<32> {
+    Initializable_assertInitialized();
+    return persistentHash<Vector<4, Bytes<32>>>(
+      [
+        id,
+        _instanceSalt,
+        counter as Field as Bytes<32>,
+        pad(32, "ZOwnablePK:shield:")
+      ]
+    );
+  }
+
+  /**
+   * @description Computes the unique identifier (`id`) of the owner from their
+   * public key and a secret nonce.
+   *
+   * ## ID Derivation
+   * `id = SHA256(pk, nonce)`
+   *
+   * - `pk`: The public key of the caller. This is passed explicitly to allow
+   * for off-chain derivation, testing, or scenarios where the caller is
+   * different from the subject of the computation.
+   * We recommend using an Air-Gapped Public Key.
+   * - `nonce`: A secret nonce tied to the identity. The generation strategy is
+   * left to the user, offering different security/convenience trade-offs.
+   *
+   * The result is a 32-byte commitment that uniquely identifies the owner.
+   * This value is later used in owner commitment hashing,
+   * and acts as a privacy-preserving alternative to a raw public key.
+   *
+   * @notice This module allows ownership to be tied to an identity commitment derived
+   * from a public key and secret nonce.
+   * While typically used with user public keys, this mechanism may also
+   * support contract addresses as identifiers in future contract-to-contract
+   * interactions. Both are treated as 32-byte values (`Bytes<32>`).
+   *
+   * Requirements:
+   *
+   * - `pk` is not a ContractAddress.
+   *
+   * @param {Either<ZswapCoinPublicKey, ContractAddress>} pk - The public key of the identity being committed.
+   * @param {Bytes<32>} nonce - A private nonce to scope the commitment.
+   * @returns {Bytes<32>} The computed owner ID.
+   */
+  export pure circuit _computeOwnerId(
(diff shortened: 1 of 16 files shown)


contracts/ownable/src/ZOwnablePK.compact
```diff
@@ -0,0 +1,305 @@
...
+ * @dev Commitment structure:
+ * ```
+ * id = H(pk, secretNonce)
+ * commitment = H(id, instanceSalt, counter, "ZOwnablePK:shield:")
+ * ```
+ * The commitment changes on each transfer due to the incrementing `counter`,
+ * providing unlinkability across ownership changes.
+ *
+ * @dev Security Considerations:
+ * - The `secretNonce` must be kept private. Loss of the nonce prevents the
```
emnul, line 36: I think it should also be emphasized that the nonce should be a cryptographically strong random value. Otherwise an attacker can brute force the value and reveal the owner.

contracts/ownable/src/ZOwnablePK.compact
```diff
@@ -0,0 +1,305 @@
...
+ * - Obfuscated owner identity: The owner's public key is never revealed on-chain.
+ * - Stateless verification: The contract never needs access to the full public key.
+ * - Built-in support for transfer and renounce functionality.
+ * - Instance-specific salts to prevent cross-contract correlation.
+ * - Deterministic hashing with `persistentHash` to support zero-knowledge verification.
+ *
+ * @dev Commitment structure:
+ * ```
+ * id = H(pk, secretNonce)
+ * commitment = H(id, instanceSalt, counter, "ZOwnablePK:shield:")
```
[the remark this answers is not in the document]
emnul: Ahhh rats that's right. I recall now a previous discussion: 

> constructors are essentially limited just to modifying contract state, and cannot take external actions, such as minting (or receiving, spending, etc.) coins.

contracts/ownable/src/ZOwnablePK.compact
```diff
@@ -0,0 +1,305 @@
...
+  export sealed ledger _instanceSalt: Bytes<32>;
+
+  /**
+   * @witness secretNonce
+   * @description A private per-user nonce used in deriving the shielded owner identifier.
+   *
+   * Combined with the user's public key as `H(pk, nonce)` to produce an obfuscated,
+   * unlinkable identity commitment. Users are encouraged to rotate this value on ownership changes.
+   */
+  export witness secretNonce(): Bytes<32>;
```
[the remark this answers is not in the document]
emnul: > It's pretty clunky and awkward to explicitly pass it in during circuit calls. The user would have to either type it out, c/p from some (hopefully) secure source, or have something like a .env. If the latter two, why not just fetch it from PS and avoid the friction? If users actually want to type it out, I'd ask them why do they hate themselves so much

Imagining typing out a 32 byte hex value to make a simple function call was enough to convince me that we should just use witnesses to handle this.

contracts/src/access/ZOwnablePK.compact
0xisk, line 1: **ZOwnablePK – before vs after `transientHash`**

* **transferOwnership**
  Before: k=16, rows=39,240 → After: k=12, rows=2,441
  Δ rows: −36,799 (\~93.8% fewer)

* **renounceOwnership**
  Before: k=15, rows=24,442 → After: k=11, rows=1,474
  Δ rows: −22,968 (\~94.1% fewer)

* **assertOnlyOwner**
  Before: k=15, rows=24,437 → After: k=11, rows=1,469
  Δ rows: −22,968 (\~94.1% fewer)

* **\_computeOwnerCommitment**
  Before: k=14, rows=14,853 → After: k=11, rows=1,021
  Δ rows: −13,832 (\~93.1% fewer)

* **\_transferOwnership**
  Before: k=14, rows=14,823 → After: k=10, rows=980
  Δ rows: −13,843 (\~93.4% fewer)

* **initialize**: k=14, rows=14,933 (single value provided)

* **owner**: k=10, rows=57 (single value provided)

**Takeaway:** `transientHash` cuts rows by \~93–94% across key circuits, with lower k.

Therefore, Compact std should provide a standalone ZK-friendly hasher called  `posiedonHasher`. That can be used separately and won't be changed in future releases as the case with `transientHash`

emnul: Agreed. The current method of creating state commitments is incredibly expensive

contracts/src/access/ZOwnablePK.compact
```diff
@@ -0,0 +1,307 @@
...
+   * duplicate commitments among other contracts implementing ZOwnablePK.
+   *
+   * @dev The `ownerId` must be calculated prior to contract deployment.
+   *
+   * @circuitInfo k=14, rows=14933
+   *
+   * Requirements:
+   *
+   * - Contract is not initialized.
+   * - `ownerId` is not an empty array.
```
emnul, line 96:
```suggestion
   * - `ownerId` is not zero.
```

emnul:
I think we should be very clear that this contract relies specifically on using SHA256 as the standard commitment generation and verification algorithm. Using any other algorithm breaks the contract's verification assumptions and will result in a permanent loss of access. In the future, we can make a more generic version that doesn't depend on using a specific hashing algorithm.

contracts/src/access/test/ZOwnablePK.test.ts
```diff
@@ -0,0 +1,502 @@
...
+
+    it('should allow pure computeOwnerId', () => {
+      const eitherOwner = utils.createEitherTestUser('OWNER');
+
+      expect(() => {
+        ownable._computeOwnerId(eitherOwner, randomByteArray);
+      }).not.toThrow();
+    });
+  });
+
```
emnul, line 138: What do you think about adding a `describe('does not generate ownerId with SHA256')` and validating that all functions that make calls to `assertOnlyOwner` fail? I think these tests would make it clear that the user supplied `ownerId` MUST use SHA256. Otherwise, circuit verification would fail.

contracts/src/access/ZOwnablePK.compact
```diff
@@ -0,0 +1,307 @@
...
+   *
+   * @param {Bytes<32>} id - The unique identifier of the owner calculated by `H(pk, nonce)`.
+   * @param {Uint<64>} counter - The current counter or round. This increments by `1`
+   * after every transfer to prevent duplicate commitments given the same `id`.
+   * @returns {Bytes<32>} The commitment derived from `id` and `counter`.
+   */
+  export circuit _computeOwnerCommitment(
+    id: Bytes<32>,
+    counter: Uint<64>,
+  ): Bytes<32> {
```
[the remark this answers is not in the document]
andrew-fleming: Agreed. There may be scenarios where protocols need to prepare for an ownership transfer before the transfer occurs on chain. I was also playing with the idea of having two circuits: one for internal use and another to export, but I don't think the extra circuit and complexity is necessary

contracts/src/access/test/ZOwnablePK.test.ts
andrew-fleming: I think that's a very good test scenario to include. On it!

contracts/src/access/test/mocks/MockZOwnablePK.compact
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: MIT
+
+pragma language_version >= 0.15.0;
```
emnul, line 3: Keeps version consistent with implementation
```suggestion
pragma language_version >= 0.16.0;
```

contracts/src/access/test/simulators/ZOwnablePKSimulator.ts
```diff
@@ -0,0 +1,288 @@
...
+  /**
+   * @description Returns the shielded owner.
+   * @returns The shielded owner.
+   */
+  public owner(): Uint8Array {
+    return this.circuits.impure.owner();
+  }
+
+  /**
+   * @description
```
emnul, line 214: Non-blocking: Missing description

contracts/src/access/test/simulators/ZOwnablePKSimulator.ts
```diff
@@ -0,0 +1,288 @@
...
+   */
+  public _computeOwnerCommitment(id: Uint8Array, counter: bigint): Uint8Array {
+    return this.circuits.impure._computeOwnerCommitment(id, counter);
+  }
+
+  /**
+   * @description
```
emnul, line 245: Non-blocking: Missing desc

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -399,3 +399,199 @@ Requirements:
...
+[.contract-index]
+.Circuits
+--
+
+[.sub-index#ZOwnablePKModule]
+* xref:#ZOwnablePK-initialize[`++initialize(ownerId, instanceSalt)++`]
+* xref:#ZOwnablePK-owner[`++owner()++`]
+* xref:#ZOwnablePK-transferOwnership[`++transferOwnership(newOwnerId)++`]
+* xref:#ZOwnablePK-renounceOwnership[`++renounceOwnership()++`]
+* xref:#ZOwnablePK-assertOnlyOwner[`++assertOnlyOwner(operator, approved)++`]
```
emnul, line 430: Fixes signature
```suggestion
* xref:#ZOwnablePK-assertOnlyOwner[`++assertOnlyOwner()++`]
```

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -399,3 +399,199 @@ Requirements:
...
+
+Constraints:
+
+- k=14, rows=14853
+
+[.contract-item]
+[[ZOwnablePK-_computeOwnerId]]
+==== `[.contract-item-name]#++_computeOwnerId++#++(pk: Either<ZswapCoinPublicKey, ContractAddress>, nonce: Bytes<32>) → Bytes<32>++` [.item-kind]#circuit#
+
+Computes the unique identifier (`id`) of the owner from their public key and a secret nonce.
```
emnul, line 559:
```suggestion
Computes the unique identifier (`id`) of the owner from their context-specific public key and a secret nonce.
```

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -399,3 +399,199 @@ Requirements:
...
+
+**ID Derivation**
+`id = SHA256(pk, nonce)`
+
+- `pk`: The public key of the caller.
```
emnul, line 564:
```suggestion
- `pk`: The context-specific public key of the caller.
```

docs/modules/ROOT/pages/api/access.adoc
```diff
@@ -399,3 +399,199 @@ Requirements:
...
+
+NOTE: This module allows ownership to be tied to an identity commitment derived from a public key and secret nonce.
+While typically used with user public keys,
+this mechanism may also support contract addresses as identifiers in future contract-to-contract interactions.
+Both are treated as 32-byte values (`Bytes<32>`).
+
+Requirements:
+
+- Contract is initialized.
+- `newOwner` is not a ContractAddress.
```
emnul, line 583:
```suggestion
- `pk` is not a ContractAddress.
```

emnul (commented):
We should be very clear that users should use context-specific public keys to maximize privacy guarantees.

docs/modules/ROOT/pages/access.adoc
```diff
@@ -133,6 +133,208 @@ there is no direct way for a contract to call circuits of other contracts or tra
...
+
+==== Owner ID Computation
+
+The foundation of the system is the owner identifier, computed as:
+
+```ts
+id = SHA256(pk, nonce)
+```
+
+Where `pk` is the owner's public key and `nonce` is a secret value that may be either randomly generated
```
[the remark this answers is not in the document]
andrew-fleming: You mean the `secretNonce` should be context-specific, right? i.e. `nonce = H(passphrase, pk, domain)` for recoverability. Correct me if I'm mistaken

The public key is the `ZswapCoinPublicKey`. So having a context-specific pk, if it's similar to what we talked about offline, would be more like a `ZOwnableSK` route where `contextPK = H(sk, domain, ...)`. The validation process and design would be fundamentally different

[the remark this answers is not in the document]
emnul: Sorry if I totally goldfish brained that convo, but my concerns are with timing attacks here. I'm imagining that a sophisticated attacker would likely sample from the set of all publicly disclosed `ZSwapCoinPublicKey`s rather than the set of all possible addresses since the later is an intractable problem. In the worst case scenario where a weak nonce is used, using a context-specific public key that's never exposed on-chain maximizes the unlinkability guarantees of this implementation and is generally a best practice since the less data that can be correlated to an account the better.

contracts/src/access/ZOwnablePK.compact
```diff
@@ -0,0 +1,308 @@
...
+   * - Contract is initialized.
+   * - Caller's id (`SHA256(pk, nonce)`) when used in `_computeOwnerCommitment` equals
+   * the stored `_ownerCommitment`, thus verifying themselves as the owner.
+   *
+   * @returns {[]} Empty tuple.
+   */
+  export circuit assertOnlyOwner(): [] {
+    Initializable_assertInitialized();
+
+    const nonce = secretNonce();
```
0xisk, line 191: I was thinking about if it is a good idea to think of a doing a convention here with putting a special char at the beginning of the witness functions, so that when we read the code it does not confuse with normal circuits.

Some example:

* `wit_` — short, explicit (my pick)
* `w_` — very short, but a bit opaque
* `priv_` — emphasizes secrecy
* `zk_` — ZK-specific vibe
* `_w_` — stands out, but leading underscores sometimes imply “internal”

I think `wit_` looks good.

**Examples in your module:**

* `secretNonce` → `wit_secretNonce`

Not sure if that is a good approach to follow.

docs/modules/ROOT/pages/access.adoc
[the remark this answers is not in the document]
andrew-fleming: I'm not a huge fan of using and repeating `unique, context-specific public key`, so I'm using the term `Air-Gapped Public Key (AGPK)`. I like the metaphor :) there's now a section and refs to it in the docs. lmk what you think

AGPK:
- **Never used before:** The public key has no prior transaction history on any blockchain network.
- **Never used elsewhere:** The key is exclusively reserved for the specific contract’s administrative circuits i.e. `assertOnlyOwner`.
- **Never used again:** The private key is destroyed once ownership is renounced or transferred to another account.

contracts/src/access/ZOwnablePK.compact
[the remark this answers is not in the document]
andrew-fleming: Before I push this in the PR: thoughts @emnul? We can always wait and do it separately if we want to discuss it further

docs/modules/ROOT/pages/access.adoc
```diff
@@ -133,6 +133,239 @@ there is no direct way for a contract to call circuits of other contracts or tra
...
+*Loss of either component results in permanent, irrecoverable loss of ownership.*
+
+==== Deterministic Nonce
+
+Deriving the nonce deterministically enables recovery through derivation schemes.
+Some examples:
+
+- `H(passphrase + context)` - recoverable from passphrase only, but passphrase becomes critical single point of failure.
+- `H(publicKey + userPassphrase + context)` - requires both public key and passphrase.
+- `H(signature + context)` where `signature = sign(context)` - leverages wallet without exposing private key.
```
emnul, line 219: Should we include a note or warning here that the user must ensure that the signature scheme used in the signature based recovery mechanisms must be deterministic? 

Fun fact: Ethereum's ECDSA implementation was non-deterministic prior to 2017.

contracts/src/access/ZOwnablePK.compact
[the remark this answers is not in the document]
emnul: I like the idea! Let's push it. I'll follow the same convention in Shielded AccessControl

docs/modules/ROOT/pages/access.adoc
andrew-fleming: Mmm good point! On it

docs/modules/ROOT/pages/access.adoc
```diff
@@ -133,6 +133,239 @@ there is no direct way for a contract to call circuits of other contracts or tra
...
+reusing a public key across different on-chain activities can potentially compromise privacy
+through transaction pattern analysis.
+
+==== AGPK Principles
+
+An Air-Gapped Public Key must adhere to strict isolation principles:
+
+- *Never used before:* The public key has no prior transaction history on any blockchain network.
+- *Never used elsewhere:* The key is exclusively reserved for the specific contract's administrative circuits i.e. `assertOnlyOwner`.
+- *Never used again:* The private key is destroyed once ownership is renounced or transferred to another account.
```
emnul, line 285: I agree this is the best practice, but we should elaborate in the AGPK description that an AGPK is generated from a fresh private key pairing with no active public keys and that the expectation is the associated private key will be destroyed. That way users don't generate a new public key from a pre-existing private key with active public keys and later have to deal with the friction of migrating all funds to another fresh account or worse, simply continue using the account.

docs/modules/ROOT/pages/access.adoc
andrew-fleming: [02227ab](https://github.com/OpenZeppelin/compact-contracts/pull/204/commits/02227ab2cfa624606d33227df575206008feea88)

docs/modules/ROOT/pages/access.adoc
andrew-fleming: Good call, on it!

andrew-fleming: [418a7b5](https://github.com/OpenZeppelin/compact-contracts/pull/204/commits/418a7b564ae5276aef53beeb1a2dc452b0bd55b3) + [bd18c19](https://github.com/OpenZeppelin/compact-contracts/pull/204/commits/bd18c19094c59ed12748700d206f2a011539689a)

docs/modules/ROOT/pages/access.adoc
```diff
@@ -272,9 +280,16 @@ through transaction pattern analysis.
 
 An Air-Gapped Public Key must adhere to strict isolation principles:
 
-- *Never used before:* The public key has no prior transaction history on any blockchain network.
-- *Never used elsewhere:* The key is exclusively reserved for the specific contract's administrative circuits i.e. `assertOnlyOwner`.
-- *Never used again:* The private key is destroyed once ownership is renounced or transferred to another account.
+- *Never used before:* The private key material
+(including any seed, parent key, or entropy source from which this key is derived)
+has never generated any public key that appears in any on-chain transaction, across any blockchain network.
+The key material must be cryptographically virgin.
```
emnul, line 286: Phrasing lol
```suggestion
The key material must be cryptographically pure.
```

emnul (approved):
Great work! The new abstractions are a huge improvement and will definitely make development a lot easier for the more complex shielded implementations

docs/modules/ROOT/pages/access.adoc
andrew-fleming: Haha it conveys the point really well!

contracts/src/access/test/utils/AbstractContractSimulator.ts
0xisk, line 1: Out of scope of this PR:
I think all the Utils scripts can be in a separate package at the home directory /, along side with `contracts` and `compact`, we can also add `utils`. That package will have a name `@openzeppelin-compact/test-utils` or just `/utils`. 
If we agree we should create an issue for this.

contracts/src/access/test/mocks/MockZOwnablePK.compact
0xisk, line 1: Not in scope of this PR:
I am thinking that all the mocks should be handled separately in the home directory alongside with `contracts`, and `compact`, there will be also `mocks`. This package should be similar directory organize to the contracts, but it is meant to put a safe area of listing mock contracts for testing purposes, and that will also allow easily to make e2e mock contracts for some list of contracts together. It can be deployed separately in a way that will make it easier for devs to also use those mocks for their testings or as a start point. It should have a name of `@openzeppelin-compact/mocks`. 
If we agree, we should open a separate issue for this.

contracts/src/access/test/types/test.ts
0xisk, line 1: Not in the scope of this PR: 
As the other comments, I think that should be also separated in a separate package in the home directory. It can be called `@openzeppelin-compact/test-types` or just `/types`. The idea as the others is to simplify the contracts directory components, and to allow devs to easily reuse those types. 
If we agree, we should open an issue for this.

contracts/src/access/test/simulators/AccessControlSimulator.ts
0xisk, line 1: Not in scope of this PR:
I think Simulators should be separated out in its directory in the home package `@openzeppelin-compact/simulators`. The idea as in the other comments for the `uitls`, `types`, and `mocks`.
If we agree, we should open an issue for this.

0xisk (approved):
Great work @andrew-fleming! approved! But I left some points for future discussions or TODOs, for any point we agree on we need to open an issue for it. 

1. `instanceSalt` and `ownerId` needs a revisit.
2. We should switch to a zk-friendly hash, the current circuits info are too expensive. 
3. Clean and simplify `contracts` pkg components. The current directory structure in `contracts` has too much of sub-directories. I think that can be improved by separating responsibilities by initialing new separate packages. I imagine the `contracts` pkg should contain sub-directories each is a separate feature like `access` and then inside we will have the main `*.compact` files, and then two directories `tests`, and `witnesses`. For the `tests` I think it should have only `*.test.ts` files, without all the others. By this way if you want to compile the current `contracts` supported features, you do not have to compile also the `mocks` with it, unless you are doing testings. Below I listed some suggestions that can help:

   1. Down below I suggested a way which is basically atomic-responsibility based. The downside of it will be creating several repos, but the good side is that devs will be able to download packages separately easily as their need.
   2. Other idea would be just creating one package in the root directory that has all of `utils`, `types`, `simulators`, and `mocks`. It can be called sth like `@openzeppelin-compact/tests-utils`. The good side it is one package for us to manage and deploy, the bad side, if devs want to reuse just one of the simulator they have to download all this big package. 
   3. Extreme opinion: That basically can be applied to any of the above two suggestions, it is to also separate the `tests` entirely which will simplify the `contracts` components in a way that it only has `*.compact` and one sub directory called `witnesses`. What I mean by `tests` here I mean only all the `*.test.ts` files, because the others will be separated already if we follow any of the above suggestions. That can be called `@openzepplin-compact/tests`. 
   
TBH, I am not entirely sure if those suggestions will in principle be in a conflict with your plan @andrew-fleming  of releasing. Please let me know if I am missing sth here.

contracts/src/access/test/mocks/MockZOwnablePK.compact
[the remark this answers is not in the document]
andrew-fleming: No strong opinions on the proposed name change. `ZOwnablePK.mock.compact` is indeed nicer to look at

andrew-fleming: I think it's a good idea to both consolidate and separate the mock contracts. I'm in favor but I think this should be very low on the list of priorities. There's much more impactful work to do elsewhere

contracts/src/access/test/types/test.ts
andrew-fleming: The utils and types are almost exclusively for simulators. If we abstract the simulator logic away in a dedicated package, this includes what would be inside test-types (i.e. `IContractSimulator`)

andrew-fleming:
> But I left some points for future discussions or TODOs, for any point we agree on we need to open an issue for it.

@0xisk I started responding to each point; however, I think it's better to just open each of the suggestions/ideas as issues and/or GH discussions anyway. That way, the convo is better organized and we can easily revisit things as time passes

PR #219, opened 2025-08-29
Bump yarn version

emnul:
## Types of changes

- Bumps yarn version to 4.9.4
- Removes yarnPath in favor of using yarn via `corepack`
- Removes unused npmScope
- Updates yarn.lock


.github/actions/setup/action.yml
```diff
@@ -4,6 +4,10 @@ description: "Sets up the environment with yarn, Node.js, and turbo"
 runs:
   using: "composite"
   steps:
+    - name: Enable Corepack
+      shell: bash
+      run: corepack enable
+
     - name: Get yarn cache directory path
```
andrew-fleming, line 11: Since we're caching yarn here:

```
    - name: Setup Node.js
      uses: actions/setup-node@v4
      with:
        node-version-file: ".nvmrc"
        cache: "yarn"
```

Do we still need the manual caching with "Get yarn cache dir path"?

andrew-fleming (approved):
LGTM! Just left a question

PR #329, opened 2025-12-05
Add `paths-ignore` to `test.yml` file

emnul:
Contract tests take quite some time to build and run. We can avoid unnecessary runs by ignoring changes to certain file types

## Summary by CodeRabbit

* **Chores**
  * Optimized CI/CD pipeline to skip unnecessary test runs when only documentation and configuration files are modified.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -17,6 +17,11 @@ jobs:
       contents: read
       security-events: write
 
+    strategy:
+      fail-fast: false
+      matrix:
+        language: ["javascript-typescript", "actions"]
+
     env:
       COMPACT_INSTALLER_URL: ${{ vars.COMPACT_INSTALLER_URL }}
 
@@ -35,9 +40,11 @@ jobs:
       - name: Initialize CodeQL
         uses: github/codeql-action/init@fe4161a26a8629af62121b670040955b330f9af2 # v4
         with:
-          languages: javascript-typescript
+          languages: ${{ matrix.language }}
           # We can add custom queries later when needed
           # queries: security-extended
 
       - name: Perform CodeQL Analysis
         uses: github/codeql-action/analyze@fe4161a26a8629af62121b670040955b330f9af2 # v4
+        with:
+          category: "/language:${{ matrix.language }}"


tirumerla:
One option is to run tests only on "changed files since main" for PRs but run on everything after merged to main only if tests have changes.

.github/workflows/test.yml
```diff
@@ -2,9 +2,19 @@ name: Compact Contracts Test Suite
 
 on:
   pull_request:
+    paths-ignore:
+      - "**.md"
+      - ".gitignore"
+      - "biome.json"
+      - "turbo.json"
```
andrew-fleming: My apprehension with these two is that changes to either can allow undetected failures to sneak into the repo. This isn't a huge deal with biome, but it could be an annoying thing with turbo. If there's a change or a version update only in the json, you will only ever see it if you run it locally or after it's merged indirectly (the next PR that triggers all the checks)

andrew-fleming: If you really think this is worth it, I recommend removing `biome` and `turbo` from the `push` trigger. Worst case and something breaks, we'll catch it after it's merged

andrew-fleming (approved):
j/k^ LGTM 👍

.github/workflows/codeql.yml
```diff
@@ -33,6 +33,8 @@ jobs:
 
       - name: Setup Environment
         uses: ./.github/actions/setup
+        with:
+          gh-token: ${{ secrets.GITHUB_TOKEN }} # secrets must be passed to composite actions
```
andrew-fleming: AFAICT, we can `skip-compact` here bc codeql isn't analyzing the artifacts, just the js/ts source code. Do you agree?

.github/workflows/test.yml
```diff
@@ -28,9 +38,11 @@ jobs:
 
       - name: Setup Environment
         uses: ./.github/actions/setup
+        with:
+          gh-token: ${{ secrets.GITHUB_TOKEN }} # secrets must be passed to composite actions
 
       - name: Compile contracts (with retry)
-        run: turbo compact --filter=@openzeppelin/compact-contracts --concurrency=1
+        run: turbo compact --filter=@openzeppelin/compact-contracts
```
andrew-fleming: 🐎 Fasterrrrrrrrrr

.github/workflows/checks.yml
```diff
@@ -25,6 +25,8 @@ jobs:
 
       - name: Setup Environment
         uses: ./.github/actions/setup
+        with:
+          gh-token: ${{ secrets.GITHUB_TOKEN }} # secrets must be passed to composite actions
```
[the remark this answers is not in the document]
emnul: Great idea!

.github/workflows/codeql.yml
emnul: > AFAICT, we can skip-compact here bc codeql isn't analyzing the artifacts, just the js/ts source code. Do you agree?

Yes, also CodeQL doesn't support analysis of `*.cjs` files produced by the compact compiler anyway ([src](https://codeql.github.com/docs/codeql-overview/supported-languages-and-frameworks/#languages-and-compilers))

andrew-fleming (approved):
LGTM! Nice work

PR #424, opened 2026-04-08
Add multisig signer

andrew-fleming:
This PR proposes to add a generic signer module for use in constructing multisig contracts

## Summary by CodeRabbit

* **New Features**
  * Added Signer module for multisig governance contracts providing comprehensive signer management. Features include initialization with configurable approval thresholds, membership validation for access control, threshold enforcement for approval requirements, and dynamic signer reconfiguration with automatic validation to maintain governance integrity throughout the contract lifecycle.


contracts/src/multisig/test/Signer.test.ts
```diff
@@ -0,0 +1,341 @@
...
+      // Check thresh
+      expect(contract.getThreshold()).toEqual(THRESHOLD);
+
+      // Check signers
+      expect(contract.getSignerCount()).toEqual(BigInt(SIGNERS.length));
+      expect(() => {
+        for (let i = 0; i < SIGNERS.length; i++) {
+          contract.assertSigner(SIGNERS[i]);
+        }
+      }).to.be.ok;
```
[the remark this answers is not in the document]
andrew-fleming: fixed

contracts/src/multisig/Signer.compact
```diff
@@ -0,0 +1,247 @@
...
+  export circuit initialize<#n>(
+    signers: Vector<n, T>,
+    thresh: Uint<8>
+  ): [] {
+    for (const signer of signers) {
+      _addSigner(signer);
+    }
+
+    _changeThreshold(thresh);
+  }
```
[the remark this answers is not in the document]
andrew-fleming: Added `Initializable_initialize`. This module supports custom flows where contracts may choose not to use `initialize`; therefore, we only prevent `initialize` from being called multiple times. The custom flows support is documented and it is explicit that contracts must NOT use initialize outside of the constructor

contracts/src/multisig/Signer.compact
```diff
@@ -0,0 +1,257 @@
...
+  export circuit _changeThreshold(newThreshold: Uint<8>): [] {
+    assert(newThreshold <= _signerCount, "Signer: threshold exceeds signer count");
+    _setThreshold(newThreshold);
+  }
+
+  /**
+   * @description Sets the approval threshold without checking
+   * against the current signer count. Intended for use during
+   * contract construction or custom setup flows where signers
+   * may not yet be registered.
```
0xisk: 🟡 **non-blocking:** I think it worth adding this in a `@warning` or `@notice` separate block as it is important note.

contracts/src/multisig/Signer.compact
```diff
@@ -0,0 +1,257 @@
...
+ * this could corrupt the signer set and threshold configuration.
+ */
+module Signer<T> {
+  import CompactStandardLibrary;
+  import "../security/Initializable" prefix Initializable_;
+
+  // ─── State ──────────────────────────────────────────────────────────────────
+
+  export ledger _signers: Set<T>;
+  export ledger _signerCount: Uint<8>;
```
0xisk, line 53:
```
`Uint<8>` for both count and threshold is implicit. If the intended maximum is < 255 (e.g., 100), the module gives no way to enforce this. Consider either a named constant `MAX_SIGNERS` or a `Uint<7>` type if the design intends a lower limit.
```

non-blocking: What do you think of this suggestion by Claude?

contracts/src/multisig/Signer.compact
```diff
@@ -0,0 +1,257 @@
...
+   *
+   * @param {Vector<n, T>} signers - The initial signer set.
+   * @param {Uint<8>} thresh - The minimum number of approvals required.
+   * @returns {[]} Empty tuple.
+   */
+  export circuit initialize<#n>(
+    signers: Vector<n, T>,
+    thresh: Uint<8>
+  ): [] {
+    Initializable_initialize();
```
0xisk, line 80: non-blocking: `initialize` with an zero signers vector gives a misleading error ("threshold exceeds signer count") — should assert non-empty explicitly

0xisk (approved):
LGTM! thank you @andrew-fleming! left some non-blocking comments. 

🔴 **blocking:**  I think we should include `Initializable_assertInitialized()` for those public circuits for more clear error handing wdyt? if you agree all those can be handled in diff PR:

```
assertSigner()
assertThresholdMet()
getSignerCount()
getThreshold()
isSigner()
```

contracts/src/multisig/Signer.compact
andrew-fleming: Good idea!

contracts/src/multisig/Signer.compact
andrew-fleming: hmm idk if I agree. Consider that the threshold cannot be zero. So if we have a threshold of two, for example, and only one signer, then I assume we can agree that this is a "threshold exceeds signer count", right? If we provide zero signers, isn't that the exact same issue...just with zero signers instead of one?

contracts/src/multisig/Signer.compact
andrew-fleming: Fair suggestions but making a whole ceremony of `MAX_SIGNERS` seems like a bit much. If we extend this into other modules for consistency, we should have `MAX_TOKEN_VALUE`, `MAX_TOKEN_ID`, etc. Documenting the max amount of signers should be enough. WDYT?

Also, if/when giant signer sets are wanted, I think creating a different Signer module that uses a MerkleTree would be a better approach e.g. offering `Signers` and `SignersMerkle`

0xisk: > Documenting the max amount of signers should be enough. WDYT?

Sounds good. :+1: 

> Also, if/when giant signer sets are wanted, I think creating a different Signer module that uses a MerkleTree would be a better approach e.g. offering Signers and SignersMerkle

I agree :+1:

contracts/src/multisig/Signer.compact
0xisk: > hmm idk if I agree. Consider that the threshold cannot be zero. So if we have a threshold of two, for example, and only one signer, then I assume we can agree that this is a "threshold exceeds signer count", right? If we provide zero signers, isn't that the exact same issue...just with zero signers instead of one?

@andrew-fleming agree with you that makes more sense!

0xisk (approved):

0xisk:
https://github.com/andrew-fleming/compact-contracts/pull/3 @andrew-fleming this PR for fixing the blocking issues.

PR #444, opened 2026-04-10
AccessControl: replace ownPublicKey with witness-derived id scheme

### LINKED ISSUE #452: C-01: `AccessControl` contract

andrew-fleming:
This PR proposes to replace `ownPublicKey()` with a witness-derived identity scheme. The module defines a `wit_AccessControlSK` witness that returns a secret key, and identity is derived as `accountId = persistentHash(secretKey)`. Only the holder of the secret key can produce the corresponding `accountId`. This intentionally mirrors Solidity's `msg.sender` which provides a stable, pseudonymous identity that persists across contracts. The hash uses no per-deployment salt or domain separator by design. If a user uses the same secret key across multiple contracts (e.g., a USDC and USDT deployment both using the FungibleToken module), their `accountId` will be identical in all of them. This is opt-in linkability: users who want cross-contract unlinkability use different keys per contract at the wallet layer.

Note that shielded variants (ShieldedAccessControl) use a more complex commitment/nullifier scheme with per-deployment salts and domain separators to provide stronger privacy guarantees. Users should choose the appropriate module variant based on their privacy requirements.

Fix: https://github.com/OpenZeppelin/compact-contracts/issues/452

### Changes

- `assertOnlyRole` now derives caller identity from `H(wit_AccessControlSK())` instead of `ownPublicKey()`
- Change account type from `Either<ZswapCoinPublicKey, ContractAddress>` to `Either<Bytes<32>, ContractAddress>` across all circuit signatures and ledger types
- Add `wit_AccessControlSK` witness returning the caller's secret key
  - The witness is namespaced to the module so that circuits calling into multiple modules within a single transaction can use independent keys for each. A shared witness name would return the same key for both calls within the same proof, preventing the user from maintaining separate identities per module
- Add `computeAccountId` (pure) circuit for off-chain identity derivation
- Add `_computeAccountId` (internal) circuit using the witness

### Witnesses

- Add AccessControlWitnesses with `wit_AccessControlSK` implementation

### Effects

- Only the holder of the secret key can produce the corresponding `accountId`
- All circuit signatures and ledger types that referenced `Either<ZswapCoinPublicKey, ContractAddress>` are updated to `Either<Bytes<32>, ContractAddress>`. The account identifier is now a hash-derived commitment rather than a public key. This propagates through circuit parameters, return types, and ledger state
- Subsequent PRs will apply the same witness-derived id pattern to enable opt-in linkability between contract deployments

### Trade-offs

- Users are responsible for generating, storing, and backing up their secret keys. Key loss means identity loss. There is no on-chain recovery mechanism. The secret key must be available as raw bytes to the local proving environment during proof gen, as it is used as a hash preimage inside the circuit. The key does not require a specific cryptographic format. Any 32 bytes of sufficient entropy will work. Users may derive the key from an HSM-held master secret or store it in an encrypted vault, but the raw value must ultimately reach the prover
- The same secret key produces the same `accountId` across all unshielded contracts. Users who require unlinkability must manage separate keys per contract. This is an intentional design choice mirroring `msg.sender` semantics
- The `H(sk)` authentication mechanism is currently baked into each module rather than provided as a pluggable abstraction. When in-circuit EC signature verification becomes available in Compact, a more general authentication layer could decouple identity verification from authorization logic which would allow modules to accept identities proven by a hash preimage, ECDSA signature, or other mechanisms without modifying the authorization or business logic modules
- Signing-only HSMs that refuse to export any derived key material are incompatible with this scheme, as the secret key must reach the prover as raw bytes. KDF-capable HSMs and hardware wallets, however, can derive a purpose-specific key internally and export the derived value, keeping root key material inside the secure boundary. This follows the same pattern as BIP-32 HD key derivation used by existing wallet infrastructure

## Release Notes

- **New Features**
  - Added `computeAccountId` function for off-chain identity commitment derivation
- **Refactor**
  - Access control now uses secret-key-based identities instead of public-key-based identities
  - Updated role management to support the new identity model
- **Tests**
  - Expanded test coverage for the new identity mechanism and role operations

### DIFF
--- a/contracts/src/access/AccessControl.compact
+++ b/contracts/src/access/AccessControl.compact
@@ -84,6 +84,14 @@ pragma language_version >= 0.21.0;
  *   Any DApp may provide any implementation. The ZK proof system constrains what values
  *   can produce valid proofs.
  *
+ * @dev Canonicalization
+ * All `Either<Bytes<32>, ContractAddress>` values used as map keys in `_operatorRoles`
+ * are canonicalized before reads and writes. Canonicalization zeroes out the inactive
+ * branch of the Either, ensuring that two values with the same active branch always
+ * resolve to the same map key regardless of what data the inactive branch carries.
+ * `hasRole` canonicalizes on read. `_unsafeGrantRole` and `_revokeRole` canonicalize
+ * on write. Internal circuits use `_hasRole` which assumes pre-canonicalized inputs.
+ *
  * @notice Missing Features and Improvements:
  *
  * - Role events
--- a/contracts/src/access/AccessControl.compact
+++ b/contracts/src/access/AccessControl.compact
@@ -9,13 +9,24 @@ pragma language_version >= 0.21.0;
  * This module provides a role-based access control mechanism, where roles can be used to
  * represent a set of permissions.
  *
+ * Authorization is based on a witness-derived identity scheme. Each caller proves knowledge
+ * of a secret key by injecting it via the `wit_AccessControlSK` witness. The module computes
+ * an account identifier as `persistentHash(secretKey)` which is a commitment that hides the key
+ * while providing a stable, pseudonymous on-chain identity.
+ *
+ * Because the account identifier is `H(secretKey)` with no per-deployment salt or domain
+ * separator, the same secret key produces the same identity across all contracts. This is
+ * intentional. It provides a linkable pseudonymous identity analogous to Solidity's
+ * `msg.sender`. Users who desire cross-contract unlinkability can use different secret keys
+ * per contract at the wallet layer.
+ *
  * Roles are referred to by their `Bytes<32>` identifier. These should be exposed
  * in the top-level contract and be unique. One way to achieve this is by
  * using `export sealed ledger` hash digests that are initialized in the top-level contract:
  *
- * ```typescript
+ * ```compact
  * import CompactStandardLibrary;
- * import "./node_modules/@openzeppelin-compact/accessControl/src/AccessControl" prefix AccessControl_;
+ * import "./node_modules/@myProject/accessControl/src/AccessControl" prefix AccessControl_;
  *
  * export sealed ledger MY_ROLE: Bytes<32>;
  *
@@ -26,10 +37,10 @@ pragma language_version >= 0.21.0;
  *
  * To restrict access to a circuit, use {assertOnlyRole}:
  *
- * ```typescript
+ * ```compact
  * circuit foo(): [] {
- *  assertOnlyRole(MY_ROLE);
- *  ...
+ *  AccessControl_assertOnlyRole(MY_ROLE);
+ *  // ... rest of circuit logic
  * }
  * ```
  *
@@ -47,7 +58,7 @@ pragma language_version >= 0.21.0;
  * grant and revoke this role. Extra precautions should be taken to secure
  * accounts that have been granted it.
  *
- * @notice Roles can only be granted to ZswapCoinPublicKeys
+ * @notice Roles can only be granted to Bytes<32> identifiers
  * through the main role approval circuits (`grantRole` and `_grantRole`).
  * In other words, role approvals to contract addresses are disallowed through these
  * circuits.
@@ -64,6 +75,16 @@ pragma language_version >= 0.21.0;
  * @notice The unsafe circuits are planned to become deprecated once contract-to-contract calls
  * are supported.
  *
+ * @dev Security Considerations:
+ * - The `secretKey` must be kept private. Loss of the key prevents role holders
+ *   from proving access. Key exposure allows impersonation.
+ * - It is strongly recommended to use cryptographically secure random values for the secret key
+ *   (e.g., `crypto.getRandomValues()`). Weak or predictable keys can be brute-forced.
+ * - The `secretKey` is provided via the `wit_AccessControlSK` witness. As with all witnesses,
+ *   the contract must not assume that the witness implementation matches the developer's code.
+ *   Any DApp may provide any implementation. The ZK proof system constrains what values
+ *   can produce valid proofs.
+ *
  * @notice Missing Features and Improvements:
  *
  * - Role events
@@ -71,18 +92,17 @@ pragma language_version >= 0.21.0;
  */
 module AccessControl {
   import CompactStandardLibrary;
-  import "../utils/Utils" prefix Utils_;
 
   /**
    * @description Mapping from a role identifier -> account -> its permissions.
    * @type {Bytes<32>} roleId - A hash representing a role identifier.
-   * @type {Map<Either<ZswapCoinPublicKey, ContractAddress>, Boolean>} hasRole - A mapping from an account to a
+   * @type {Map<Either<Bytes<32>, ContractAddress>, Boolean>} hasRole - A mapping from an account to a
    * Boolean determining if the account is approved for a role.
    * @type {Map<roleId, hasRole>}
-   * @type {Map<Bytes<32>, Map<Either<ZswapCoinPublicKey, ContractAddress>, Boolean>} _operatorRoles
-   */
+   * @type {Map<Bytes<32>, Map<Either<Bytes<32>, ContractAddress>, Boolean>>} _operatorRoles
+   */
   export ledger _operatorRoles: Map<Bytes<32>,
-                                    Map<Either<ZswapCoinPublicKey, ContractAddress>, Boolean>>;
+                                    Map<Either<Bytes<32>, ContractAddress>, Boolean>>;
 
   /**
    * @description Mapping from a role identifier to an admin role identifier.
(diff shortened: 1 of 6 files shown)


contracts/src/access/test/simulators/AccessControlSimulator.ts
```diff
@@ -166,8 +165,46 @@ export class AccessControlSimulator extends AccessControlSimulatorBase {
...
+     * simulate switching between different user identities or injecting incorrect
+     * keys to test failure paths.
+     * @param newSK - The new secret key to set.
+     * @returns The updated private state.
+     */
+    injectSecretKey: (newSK: Uint8Array): AccessControlPrivateState => {
+      const updatedState = { secretKey: newSK };
+      this.circuitContextManager.updatePrivateState(updatedState);
+      return updatedState;
+    },
```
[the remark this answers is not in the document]
andrew-fleming: fixed [93591c3](https://github.com/OpenZeppelin/compact-contracts/pull/444/commits/93591c349069941b3259297740dadb271cc7428e)

contracts/src/access/witnesses/AccessControlWitnesses.ts
```diff
@@ -1,6 +1,73 @@
...
+ * @description Interface defining the witness methods for AccessControl operations.
+ * @template P - The private state type.
+ */
+export interface IAccessControlWitnesses<P> {
+  /**
+   * Retrieves the secret key from the private state.
+   * @param context - The witness context containing the private state.
+   * @returns A tuple of the private state and the secret key as a Uint8Array.
+   */
+  wit_AccessControlSK(context: WitnessContext<Ledger, P>): [P, Uint8Array];
```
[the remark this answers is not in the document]
andrew-fleming: fixed [d9d8568](https://github.com/OpenZeppelin/compact-contracts/pull/444/commits/d9d85685885034e415b7a33115d283c54f8c8101)

contracts/src/access/AccessControl.compact
```diff
@@ -9,13 +9,24 @@ pragma language_version >= 0.21.0;
...
+ * per contract at the wallet layer.
+ *
  * Roles are referred to by their `Bytes<32>` identifier. These should be exposed
  * in the top-level contract and be unique. One way to achieve this is by
  * using `export sealed ledger` hash digests that are initialized in the top-level contract:
  *
- * ```typescript
+ * ```compact
  * import CompactStandardLibrary;
- * import "./node_modules/@openzeppelin-compact/accessControl/src/AccessControl" prefix AccessControl_;
```
0xisk, line 18: ❔︎ **question:** why that change? 🤔

contracts/src/access/AccessControl.compact
```diff
@@ -123,38 +179,42 @@ module AccessControl {
...
    *
    * @param {Bytes<32>} roleId - The role identifier.
    * @return {[]} - Empty tuple.
    */
   export circuit assertOnlyRole(roleId: Bytes<32>): [] {
-    _checkRole(roleId, left<ZswapCoinPublicKey, ContractAddress>(ownPublicKey()));
+    // The account constructed via left<...>() is already canonical,
+    // but we route through _checkRole -> hasRole for consistency.
+    // The redundant canonicalization is negligible.
+    _checkRole(roleId, left<Bytes<32>, ContractAddress>(_computeAccountId()));
```
0xisk, line 199:
```suggestion
    assert(_hasRole(roleId, left<Bytes<32>, ContractAddress>(_computeAccountId())), "AccessControl: unauthorized account");
```

🔴 **blocking:**`assertOnlyRole` runs canonicalize (via `hasRole`) on a value that is already canonical. Let's use the internal `_hasRole`).

contracts/src/access/AccessControl.compact
```diff
@@ -227,22 +287,25 @@ module AccessControl {
...
+   * The caller's identity is derived from the `wit_AccessControlSK` witness as
+   * `persistentHash(secretKey)`. The `callerConfirmation` parameter must match
+   * the caller's computed account identifier to prevent accidental renunciation.
+   *
+   * @circuitInfo k=13, rows=3322
    *
    * Requirements:
    *
-   * - The caller must be `callerConfirmation`.
-   * - The caller must not be a `ContractAddress`.
```
0xisk, line 235: ❔︎ **question:** why is that removed?

contracts/src/access/AccessControl.compact
```diff
@@ -266,75 +329,127 @@ module AccessControl {
...
    * @return {Boolean} roleGranted - A boolean indicating if `roleId` was granted.
    */
   export circuit _grantRole(
                    roleId: Bytes<32>,
-                   account: Either<ZswapCoinPublicKey, ContractAddress>
+                   account: Either<Bytes<32>, ContractAddress>
                    ): Boolean {
-    assert(!Utils_isContractAddress(account), "AccessControl: unsafe role approval");
+    const isCommitment = account.is_left;
+    assert(isCommitment, "AccessControl: unsafe role approval");
```
0xisk, line 348: 🔵 **followup:** Well I think here the best would be to generalize the `isContractAddress()`

```
export pure circuit isContractAddress<T>(
                      keyOrAddress: Either<T, ContractAddress>
                      ): Boolean {
  return !keyOrAddress.is_left;
}
```

contracts/src/access/witnesses/AccessControlWitnesses.ts
```diff
@@ -1,6 +1,77 @@
-// SPDX-License-Identifier: MIT
-// OpenZeppelin Compact Contracts v0.0.1-alpha.1 (access/witnesses/AccessControlWitnesses.ts)
+import { getRandomValues } from 'node:crypto';
```
0xisk, line 4: 🔵 **followup:** As that is only node specific, we need to document that those witnesses are for testing purposes!

contracts/src/access/witnesses/AccessControlWitnesses.ts
```diff
@@ -1,6 +1,77 @@
...
+      throw new Error(
+        `withSecretKey: expected 32-byte secret key, received ${sk.length} bytes`,
+      );
+    }
+    return { secretKey: Uint8Array.from(sk) };
+  },
+};
+
+/**
+ * @description Factory function creating witness implementations for Ownable operations.
```
0xisk:
```suggestion
 * @description Factory function creating witness implementations for AccessControl operations.
```

🔵 **followup:**

contracts/src/access/AccessControl.compact
```diff
@@ -266,75 +329,127 @@ module AccessControl {
...
 
   /**
    * @description Attempts to grant `roleId` to `account` and returns a boolean indicating if `roleId` was granted.
    * Internal circuit without access restriction. It does NOT check if the role is granted to a ContractAddress.
    *
-   * @circuitInfo k=10, rows=733
+   * @circuitInfo k=11, rows=1142
    *
-   * @notice External smart contracts cannot call the token contract at this time, so granting a role to an ContractAddress may
+   * @notice External smart contracts cannot call the token contract at this time, so granting a role to a ContractAddress may
```
0xisk: 🔵 **followup:** 

```suggestion
   * @notice External smart contracts cannot call the access control contract at this time, so granting a role to a ContractAddress may
```

contracts/src/access/witnesses/AccessControlWitnesses.ts
```diff
@@ -1,6 +1,77 @@
...
+  generate: (): AccessControlPrivateState => {
+    return { secretKey: getRandomValues(new Uint8Array(32)) };
+  },
+
+  /**
+   * @description Generates a new private state with a user-defined secret key.
+   * Useful for deterministic nonce generation or advanced use cases.
+   *
+   * @param sk - The 32-byte secret nonce to use.
+   * @returns A fresh AccessControlPrivateState instance with the provided nonce.
```
0xisk:
```suggestion
   * Useful for deterministic key generation or advanced use cases.
   *
   * @param sk - The 32-byte secret key to use.
   * @returns A fresh AccessControlPrivateState instance with the provided key.
```

🔵 **followup:** they are called secretkey everywhere.

0xisk (commented):
Looks good! Thank you @andrew-fleming! Left some comments.

contracts/src/access/test/AccessControl.test.ts
0xisk, line 1: 🔵 **followup:** missing test: `contract-address` holder of a role cannot pass `assertOnlyRole`

contracts/src/access/AccessControl.compact
andrew-fleming: Yes, that is already addressed in the comment haha do you disagree with keeping the flow consistent (and defensive) with _checkRole being the gate?

contracts/src/access/AccessControl.compact
andrew-fleming: "A caller's computed account id..." cannot be a contract address but no argument against keeping it explicit

contracts/src/access/witnesses/AccessControlWitnesses.ts
andrew-fleming: In the immediate term, I agree, but wasn't the idea of separating the mock ledger dependency in the witnesses to allow them to be used outside of a testing ctx?

contracts/src/access/AccessControl.compact
andrew-fleming: I thought about that and I for sure agree that we can improve this. For the sake of the audit scope, my aim was to keep the fixes isolated to the modules as opposed to refactoring Utils and other files (as mentioned internally). Notice that this PR only touches `AccessControl-` files

~FWIW I also don't think we should even use that generic signature because Unshielded ops, for example, use this ordering convention `Either<ContractAddress, UserAddress>` ([ref](https://docs.midnight.network/compact/standard-library/exports#mintunshieldedtoken))~

Mmm I somewhat take back this last comment. It still seems weird to have a generic if we're really only using this for one purpose

contracts/src/access/AccessControl.compact
andrew-fleming: c/p and forgot to change it back

andrew-fleming: fixed [691b5ba](https://github.com/OpenZeppelin/compact-contracts/pull/444/commits/691b5ba2f3a8920428ff192d54be8fb2b8aa9cce)

contracts/src/access/AccessControl.compact
andrew-fleming: fixed [4a3695b](https://github.com/OpenZeppelin/compact-contracts/pull/444/commits/4a3695beb2181168eae343b67951a3c515e3ab14)

contracts/src/access/AccessControl.compact
andrew-fleming: fixed [cbb907b](https://github.com/OpenZeppelin/compact-contracts/pull/444/commits/cbb907b2dd444103bec568298d668ccbe7c589dd)

contracts/src/access/test/AccessControl.test.ts
andrew-fleming: fixed [786ba62](https://github.com/OpenZeppelin/compact-contracts/pull/444/commits/786ba62fcb771cf2e5afb6e58bd54e16fef5ff2b)

contracts/src/access/AccessControl.compact
0xisk: > do you disagree\
>

Yes I disagree on repeating the step of `canonicalization` However as discussed before that this will be in a follow up issue.

contracts/src/access/AccessControl.compact
0xisk: Agree, lets track this in a followup issue separately.

contracts/src/access/witnesses/AccessControlWitnesses.ts
0xisk: > but wasn't the idea of separating the mock ledger dependency in the witnesses to allow them to be used outside of a testing ctx?

I can clarify that in another related issue, I don't see it is necessary here. We can keep the current one but just we need to comment that this will be only node specific. So I will also mark this point in the followup issue.

0xisk (approved):
LGTM! Thank you Andrew! For the lest of the follow ups they can be traced separately   here: https://github.com/OpenZeppelin/compact-contracts/issues/492

Issue #474, opened 2026-05-04
9. Per-recipient `Forwarder` preset


0xisk:
**🧐 Motivation**

A multisig that receives deposits from many distinct recipients needs a stable per-recipient deposit address so the integrator can attribute each deposit on-chain rather than via fragile out-of-band metadata.

**📝 Details**

A minimal per-recipient contract whose address acts as the recipient identifier. Deposits land on the Forwarder and are atomically forwarded to a sealed parent multisig in the same transaction; funds never sit on the Forwarder. No access control on deposits; the parent address is immutable.

PR #526, opened 2026-05-27
feat(multisig): add Forwarder + ForwarderPrivate modules

0xisk:
## Types of changes

*Put an \`\` in the boxes that apply*

Fixes https://github.com/OpenZeppelin/compact-contracts/issues/474

## PR Checklist

- [ ] I have read the [Contributing Guide](../CONTRIBUTING.md#your-first-code-contribution)
- [ ] I have added tests that prove my fix is effective or that my feature works. See [GUIDELINES.md#testing](../GUIDELINES.md#testing) for more information.
- [ ] I have added documentation for new methods or changes to existing method behavior. See [GUIDELINES.md#documentation](../GUIDELINES.md#documentation) for more information.

## Summary by CodeRabbit

* **New Features**
  * Multisig forwarder contracts: shielded and unshielded deposits automatically forwarded to a configured parent while tracking per-color cumulative received totals.
  * Private forwarder variant: parent specified via commitment and supports authenticated drains with change handling.

* **Tests**
  * Extensive test suites and simulator presets validating initialization, deposits, drains, overflow guards, and property-based checks.
  * Test coverage enabled (v8) with 95% thresholds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenZeppelin/compact-contracts/pull/526?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


contracts/src/multisig/Forwarder.compact
```diff
@@ -0,0 +1,156 @@
...
+   *
+   * Requirements:
+   *
+   * - Contract must not be initialized.
+   *
+   * @param {Bytes<32>} parent - The parent address.
+   *
+   * @returns {[]} Empty tuple.
+   */
+  export circuit _init(parent: Bytes<32>): [] {
```
pepebndc: **[Should fix] Zero-address guard missing**

`_init` accepts an all-zero `parent` without validation. A misconfigured deployment forwards every coin permanently to the zero address — no recovery path exists.

```compact
export circuit _init(parent: Bytes<32>): [] {
  assert(parent != zeroBytes<32>(), "Forwarder: zero parent");
  Initializable_initialize();
  _parent = disclose(parent);
}
```

contracts/src/multisig/ForwarderPrivate.compact
```diff
@@ -0,0 +1,152 @@
...
+   *
+   * - Contract must not be initialized.
+   *
+   * @param {Bytes<32>} parentCommitment - `persistentHash([parentAddr, salt])`
+   * computed off-chain by the deployer (see `_calculateParentCommitment`).
+   *
+   * @returns {[]} Empty tuple.
+   */
+  export circuit _init(parentCommitment: Bytes<32>): [] {
+    Initializable_initialize();
```
pepebndc, line 59: **[Should fix] Zero-commitment guard missing**

`_init` accepts an all-zero `parentCommitment` without validation. Since the commitment is the sole gate for `_drain`, a zero commitment deployed by mistake makes all accumulated shielded coins permanently undrainable.

```compact
export circuit _init(parentCommitment: Bytes<32>): [] {
  assert(parentCommitment != zeroBytes<32>(), "ForwarderPrivate: zero commitment");
  Initializable_initialize();
  _parentCommitment = disclose(parentCommitment);
}
```

contracts/src/multisig/ForwarderPrivate.compact
```diff
@@ -0,0 +1,152 @@
...
+   * stored commitment.
+   * @param {Bytes<32>} salt - The salt. Operational secret; never
+   * appears on the public transcript.
+   * @param {Uint<128>} value - The amount to send.
+   *
+   * @returns {ShieldedSendResult} The result containing the sent coin
+   * and any change.
+   */
+  export circuit _drain(
+    coin: QualifiedShieldedCoinInfo,
```
pepebndc, line 119: **[Should fix] Salt loss = permanent fund lock — must be documented here**

Salt is the sole authorization gate. There is no rotation, revocation, or recovery path. If the operator loses the salt, every shielded coin accumulated in this contract is permanently inaccessible — equivalent to losing a hot-wallet private key.

This risk needs a prominent warning on the `@param salt` NatDoc line of `_drain` itself (not just the module header), so any developer reading only this circuit's signature sees it clearly before deploying.

contracts/src/multisig/presets/forwarder/ForwarderPrivate.compact
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * @description Computes the parent commitment from a `(parentAddr, salt)`
+ * pair. Pure circuit — used off-chain by the deployer to compute the
+ * constructor argument, and inside `drain` for the preimage check.
+ *
+ * @param {Bytes<32>} parentAddr - The parent address.
+ * @param {Bytes<32>} salt - The salt.
+ *
+ * @returns {Bytes<32>} The commitment `persistentHash([parentAddr, salt])`.
+ */
```
pepebndc, line 113: **[Should fix] Public circuit named with a leading underscore**

Every other public-facing circuit in this preset — `deposit`, `drain`, `getParentCommitment` — has no leading underscore, which is the correct convention for this layer. Re-exporting `_calculateParentCommitment` with a `_` prefix breaks that contract: in this codebase `_` consistently signals "module-internal only."

Rename to `calculateParentCommitment`.

pepebndc (changes requested):
left some comments

pepebndc:
## Non-blocking suggestions

Three additional things worth considering before this ships as a stable library API.

**Overflow test missing in ForwarderShielded**

`ForwarderUnshielded.test.ts` correctly tests the `MAX_U128` overflow boundary. `ForwarderShielded` wraps the same `_recordReceived` guard and should test it independently — line coverage alone won't catch this behavioral gap.

**Double-initialization revert tests**

None of the three preset test suites assert that calling the constructor (or `_init`) a second time reverts. This is a standard OZ Compact test for any module with an initialization guard.

**Domain separation in `_calculateParentCommitment`**

`persistentHash<Vector<2, Bytes<32>>>([parentAddr, salt])` has no domain tag. If any other circuit in the system uses `persistentHash` with two `Bytes<32>` inputs, a preimage crafted in that context could theoretically satisfy this commitment. A fixed domain prefix (e.g. `["ForwarderPrivate", parentAddr, salt]`) eliminates the surface entirely.

contracts/src/multisig/Forwarder.compact
andrew-fleming: I'd stick with `initialize` to be consistent with the rest of the lib and the extensibility pattern

contracts/src/multisig/Forwarder.compact
```diff
@@ -0,0 +1,156 @@
...
+
+  export ledger _parent: Bytes<32>;
+  export ledger _received: Map<Bytes<32>, Uint<128>>;
+
+  // ─── Init ───────────────────────────────────────────────────────
+
+  /**
+   * @description Initializes the forwarder with a parent address.
+   * Called once from the preset constructor. The parent is the
+   * recipient of every forwarded coin and is immutable after init.
```
andrew-fleming, line 37: If the claim is that the parent is immutable, it should be `sealed`, no?

contracts/src/multisig/Forwarder.compact
0xisk: Resolved in 96c1b9c — _init now asserts parent != default<Bytes<32>> with message "Forwarder: zero parent" and the Requirements: block is updated.

contracts/src/multisig/ForwarderPrivate.compact
0xisk: Resolved in 96c1b9c — _init asserts parentCommitment != default<Bytes<32>> with message "ForwarderPrivate: zero commitment".

contracts/src/multisig/ForwarderPrivate.compact
0xisk: Resolved in 96c1b9c — added a prominent @warning block right on the @param salt line in _drain, and mirrored the same warning on the preset's drain wrapper.

contracts/src/multisig/presets/forwarder/ForwarderPrivate.compact
0xisk: Resolved in 96c1b9c — preset re-export renamed to calculateParentCommitment. Module-internal helper keeps the _ prefix (consistent with the codebase convention).

0xisk:
> ## Non-blocking suggestions
> Three additional things worth considering before this ships as a stable library API.
> 
> **Overflow test missing in ForwarderShielded**
> 
> `ForwarderUnshielded.test.ts` correctly tests the `MAX_U128` overflow boundary. `ForwarderShielded` wraps the same `_recordReceived` guard and should test it independently — line coverage alone won't catch this behavioral gap.

- Addressed in 81e119e — added on the shielded path in the new module test. Note: Zswap bounds ShieldedCoinInfo.value: Uint<64>, so the Uint<128> accumulator overflow inside _recordReceived is structurally unreachable from finite shielded deposits. The test instead asserts that a MAX_U128 coin.value is rejected (Zswap upstream) and that the Uint<64> ceiling is accepted; the Uint<128> boundary is exercised directly via the unshielded path which shares the same _recordReceived helper.


> **Double-initialization revert tests**
> 
> None of the three preset test suites assert that calling the constructor (or `_init`) a second time reverts. This is a standard OZ Compact test for any module with an initialization guard.

- Addressed in 81e119e — preset constructors are the only init entry, so following the Signer/MockSigner convention I added MockForwarder + MockForwarderPrivate test fixtures that expose initialize(...) and the not-initialized path (via an isInit flag), then asserted the double-init revert + the not-initialized-state-touching reverts at the module level.



> **Domain separation in `_calculateParentCommitment`**
> 
> `persistentHash<Vector<2, Bytes<32>>>([parentAddr, salt])` has no domain tag. If any other circuit in the system uses `persistentHash` with two `Bytes<32>` inputs, a preimage crafted in that context could theoretically satisfy this commitment. A fixed domain prefix (e.g. `["ForwarderPrivate", parentAddr, salt]`) eliminates the surface entirely.

- Addressed in 96c1b9c — commitment is now persistentHash<Vector<3, Bytes<32>>>([pad(32, "ForwarderPrivate:commitment"), parentAddr, salt]). @circuitInfo for _drain updated 47778 → 47811 to reflect the shape change.

0xisk:
@coderabbitai review

0xisk:
@coderabbitai review this pr.

0xisk:
@coderabbitai full review

contracts/src/multisig/witnesses/presets/ForwarderPrivateWitnesses.ts
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.0.1-alpha.1 (multisig/witnesses/ForwarderPrivateWitnesses.ts)
```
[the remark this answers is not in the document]
0xisk: fixed

contracts/src/multisig/witnesses/presets/ForwarderShieldedWitnesses.ts
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.0.1-alpha.1 (multisig/witnesses/ForwarderShieldedWitnesses.ts)
```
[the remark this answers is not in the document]
0xisk: fixed

contracts/src/multisig/witnesses/presets/ForwarderUnshieldedWitnesses.ts
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.0.1-alpha.1 (multisig/witnesses/ForwarderUnshieldedWitnesses.ts)
```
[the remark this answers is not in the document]
0xisk: fixed

contracts/src/multisig/presets/forwarder/ForwarderPrivate.compact
```diff
@@ -0,0 +1,123 @@
...
+
+import CompactStandardLibrary;
+import {
+  _init,
+  _deposit,
+  _drain,
+  _calculateParentCommitment,
+  _parentCommitment
+} from "../../ForwarderPrivate" prefix ForwarderPrivate_;
```
andrew-fleming: What's the benefit to listing the circuits if we're still prefixing them?

contracts/src/multisig/Forwarder.compact
```diff
@@ -0,0 +1,160 @@
...
+ */
+module Forwarder {
+  import CompactStandardLibrary;
+  import { initialize, assertInitialized } from "../security/Initializable" prefix Initializable_;
+  import { UINT128_MAX } from "../utils/Utils" prefix Utils_;
+
+  // ─── State ──────────────────────────────────────────────────────
+
+  export ledger _parent: Bytes<32>;
+  export ledger _received: Map<Bytes<32>, Uint<128>>;
```
andrew-fleming: I'm not sure I understand the value here. The parent is ultimately responsible for tracking the balances of its children. AFAICT this is just tracking what passes through, right? Is this worth having? Consider too the cumulative state from all deployed children. LMK if I'm missing something

contracts/src/multisig/Forwarder.compact
```diff
@@ -0,0 +1,160 @@
...
+   * @returns {[]} Empty tuple.
+   */
+  export circuit _depositUnshielded(color: Bytes<32>, amount: Uint<128>): [] {
+    Initializable_assertInitialized();
+    _recordReceived(color, amount);
+    receiveUnshielded(disclose(color), disclose(amount));
+    sendUnshielded(
+      disclose(color),
+      disclose(amount),
+      left<ContractAddress, UserAddress>(ContractAddress { bytes: _parent })
```
andrew-fleming: I'm not crazy about this approach. We're bypassing type safety which I think is quite important with the different param order for:
- shielded: `Either<ZswapCoinPublicKey, ContractAddress>`
- unshielded: `Either<ContractAddress, UserAddress>`

Can't we define a generic `RecipientType` so that consuming contracts must explicitly define the types? Otherwise, I think it's really easy to accidentally do something like pass the bytes of a UserAddress when it should have been a contract

contracts/src/multisig/ForwarderPrivate.compact
```diff
@@ -0,0 +1,175 @@
...
+   * @param {Bytes<32>} salt - The salt. Operational secret; never
+   * appears on the public transcript.
+   *
+   * @warning **Salt loss is permanent fund loss.** The salt is the sole
+   * drain authorization. There is no rotation, revocation, or recovery
+   * path. If the operator misplaces the salt, every shielded coin
+   * accumulated at this contract is forever inaccessible — equivalent
+   * to losing a hot-wallet private key. Back the salt up offline before
+   * the first deposit and treat it with the same hygiene as a signing
+   * key.
```
andrew-fleming: Should we change the name "salt" if it's this important? As noted above in the param definition, maybe "opSecret" or something might be better to signify how crucial it is not to lose it. WDYT?

andrew-fleming (commented):
Good start, @0xisk! I left some comments and questions

contracts/src/multisig/Forwarder.compact
0xisk: Agreed.

contracts/src/multisig/Forwarder.compact
0xisk: Agreed that as follows with BitGo to have it immutable, in case if we want to make it settable we will need to have a setter with ownable checker which is out of the scope. Let me know if you disagree.

contracts/src/multisig/presets/forwarder/ForwarderPrivate.compact
0xisk: Checked the compiler code and actually no optimization benefits, I thought we could have some benefit form the named imports. And we need the prefix as long as there is no go-to-definition lsp feature from the compiler. So I agree to remove them and leave only prefixes.

contracts/src/multisig/Forwarder.compact
0xisk: > The parent is ultimately responsible for tracking the balances of its children.\

You are right, after thinking about this again I think it is redundant.

contracts/src/multisig/Forwarder.compact
0xisk: good catch @andrew-fleming! working on fixing that.

contracts/src/multisig/ForwarderPrivate.compact
0xisk: Makes a lot of sense! :

0xisk:
@andrew-fleming Thanks for the good points! :ok\_hand: all fixed:

| # | Thread                                                                                                       | Commit                                                                                                                                                                                      |
| - | ------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1 | [`initialize` naming](https://github.com/OpenZeppelin/compact-contracts/pull/526#discussion_r3315155113)     | Fixed in [6641069](https://github.com/OpenZeppelin/compact-contracts/commit/6641069)                                                                                                        |
| 2 | [`sealed` parent](https://github.com/OpenZeppelin/compact-contracts/pull/526#discussion_r3315196480)         | Fixed in [6641069](https://github.com/OpenZeppelin/compact-contracts/commit/6641069)                                                                                                        |
| 3 | [named imports vs prefix](https://github.com/OpenZeppelin/compact-contracts/pull/526#discussion_r3332049818) | Fixed in [6641069](https://github.com/OpenZeppelin/compact-contracts/commit/6641069) (presets/mocks), [70742e7](https://github.com/OpenZeppelin/compact-contracts/commit/70742e7) (modules) |
| 4 | [`_received` value](https://github.com/OpenZeppelin/compact-contracts/pull/526#discussion_r3332121220)       | Fixed in [9bce9d5](https://github.com/OpenZeppelin/compact-contracts/commit/9bce9d5)                                                                                                        |
| 5 | [recipient type safety](https://github.com/OpenZeppelin/compact-contracts/pull/526#discussion_r3332164337)   | Fixed in [0a7d41b](https://github.com/OpenZeppelin/compact-contracts/commit/0a7d41b)                                                                                                        |
| 6 | [`salt` → `opSecret`](https://github.com/OpenZeppelin/compact-contracts/pull/526#discussion_r3332188437)     | Fixed in [189ca65](https://github.com/OpenZeppelin/compact-contracts/commit/189ca65)                                                                                                        |

contracts/src/multisig/Forwarder.compact
andrew-fleming: Yes, I agree! I'm bringing this up more for the comment "The parent...is immutable after init". I don't think we should make this claim because we're not enforcing it

contracts/src/multisig/presets/forwarder/ForwarderPrivate.compact
andrew-fleming: Ahh optimization is a clever consideration. Womp, womp. Thanks for checking it out

pepebndc (approved):

contracts/src/multisig/Forwarder.compact
andrew-fleming: I actually think it'd be better if we don't enforce immutability. For contracts that will not change the parent, it's as simple as don't create/expose a circuit for it. For contracts that want to change the parent, this can be easily added...either in the module itself or it can be deferred for the contract to handle. Otherwise, users are forced to do something like create their own workaround or build their own module

contracts/src/multisig/witnesses/presets/ForwarderPrivateWitnesses.ts
andrew-fleming, line 1: Instead of having a bajillion empty witnesses/PS, can we just define it once in `multisig/test/` or even in `test-utils/` and import it into the simulator? I only ask now in this specific context because it's getting unnecessarily bloated and messy

contracts/src/multisig/Forwarder.compact
```diff
@@ -0,0 +1,112 @@
...
+   * @param {ShieldedCoinInfo} coin - The incoming shielded coin.
+   *
+   * @returns {[]} Empty tuple.
+   */
+  export circuit _depositShielded(coin: ShieldedCoinInfo): [] {
+    Initializable_assertInitialized();
+    receiveShielded(disclose(coin));
+    sendImmediateShielded(
+      disclose(coin),
+      left<ZswapCoinPublicKey, ContractAddress>(ZswapCoinPublicKey { bytes: _parent.bytes }),
```
andrew-fleming, line 82: I'd be really careful here. What happens if a user instantiates the generic with `ContractAddress`? The encoding is different so that'd be a big problem and there's nothing preventing a user from doing this. I think the generic needs to be `Either<T1, T2>`

andrew-fleming: Or better yet, to avoid ALL confusion, we could just split this into two modules and define the concrete types (shielded or unshielded `Either<>`). There's really not a whole lot of duplication in the module...just a single state variable and initializer. WDYT?

andrew-fleming (commented):
Looking good @0xisk! I left some comments

0xisk:
Thanks @andrew-fleming for your followup review. I opened an issue having all those comments will solve them on a following PR.

PR #562, opened 2026-06-09
fix(security): track initialization state per module

### LINKED ISSUE #556: Consider different initialization pattern
Currently, there's a shared transitive dependency compiler bug https://github.com/LFDT-Minokawa/compact/issues/270. It's unclear if the solution will be that transitive deps that resolve to the same file will or will not share ledger state across all importing modules. Thus, at least for now, I propose each module tracks its own initializable state e.g.

```ts
module MyModule {
  ledger _isInitialized: Boolean;

  export circuit initialize(): {
    assert(!_isInitialized, "MyModule: contract already inintialized");
    _isInitialized = true;
    (...)
  }

  export circuit foo(): [] {
    assertInitialized();
    (...)
  }

  export circuit bar(): [] {
    assertInitialized();
    (...)
  }

  // Internal to reduce noise
  circuit assertInitialized(): [] {
    assert(_isInitialized, "MyModule: contract not initialized")
  }

  // rest of module
}
```

0xisk:
## Types of changes

Fixes #556

Composing two modules that both import the shared `Initializable` from the **same directory** collapses their `_isInitialized` ledger state into a single slot (compiler bug [LFDT-Minokawa/compact#270](https://github.com/LFDT-Minokawa/compact/issues/270)). Initializing one module then marks the other initialized and blocks its own initialization. The behaviour is directory-layout dependent and a security hazard for any contract that composes OZ modules.

**Reproduced** on both the repo toolchain (compact 0.29.0 / runtime 0.14.0) and the latest (0.31.0 / 0.16.0): same-directory composition yields one shared `_isInitialized` slot; different-directory yields two. Not fixed upstream.

### Change

Each module now owns its `_isInitialized` ledger flag and inlines its initialization assertions, removing the shared transitive dependency. `Initializable` is kept (stateful) for direct single-import use, with a technical warning documenting #270.

- **security:** keep `Initializable` stateful; add a #270 technical warning.
- **access/token:** inline a per-module init flag + assertions in `Ownable`, `ZOwnablePK`, `ShieldedAccessControl`, `FungibleToken`, `NonFungibleToken`, `MultiToken`, with module-specific revert messages.
- **test:** update revert-message expectations to the per-module strings.
- **test(integration):** new `initStateIsolation` spec proving the collision (shared `Initializable`) and the fix (per-module flags), plus `vitest.integration.config.ts` and a `test:integration` script.

> **Breaking:** the public ledger key moves from `Initializable__isInitialized` to per-module keys (`Ownable__isInitialized`, `FungibleToken__isInitialized`, …). Targeted at the 0.2.0 release.

## PR Checklist

- [x] I have read the Contributing Guide
- [x] I have added tests that prove my fix is effective (unit suite green; new integration spec proves bug + fix)
- [x] I have added documentation for new methods or changes to existing method behavior

#### Further comments

The map-of-modules and stateless-helper alternatives were considered. Per-module flags were chosen because each module's init state is semantically independent (so this is the correct terminal layout regardless of how #270 resolves), it is the cheapest on the hot path, and — under CMA, which cannot add/remove ledger variables — it is the layout deployed contracts can keep permanently.

## Summary by CodeRabbit

* **New Features**
  * Added integration test suite for validating composed contracts with independent module initialization.

* **Bug Fixes**
  * Improved initialization state management to ensure modules maintain independent initialization status when composed together.

* **Tests**
  * Added comprehensive integration tests verifying initialization isolation across composed contracts.
  * Updated error message expectations in unit tests.

* **Documentation**
  * Added integration test suite documentation explaining test structure and expected behaviors.

* **Chores**
  * Added NPM scripts for running integration test workflows.

### DIFF
--- a/contracts/src/access/Ownable.compact
+++ b/contracts/src/access/Ownable.compact
@@ -85,8 +85,14 @@ pragma language_version >= 0.21.0;
  */
 module Ownable {
   import CompactStandardLibrary;
-  import "../security/Initializable" prefix Initializable_;
   import "../utils/Utils" prefix Utils_;
+
+  /**
+   * @description Initialization flag, tracked per-module to avoid the compiler's
+   * shared transitive-dependency state bug (LFDT-Minokawa/compact#270). See the
+   * Initializable module for rationale.
+   */
+  export ledger _isInitialized: Boolean;
 
   export ledger _owner: Either<Bytes<32>, ContractAddress>;
 
@@ -117,9 +123,36 @@ module Ownable {
    * @returns {[]} Empty tuple.
    */
   export circuit initialize(initialOwner: Either<Bytes<32>, ContractAddress>): [] {
-    Initializable_initialize();
+    assertNotInitialized();
+    _isInitialized = true;
     assert(!_isTargetZero(initialOwner), "Ownable: invalid initial owner");
     _transferOwnership(initialOwner);
+  }
+
+  /**
+   * @description Asserts that the contract has been initialized, throwing an error if not.
+   *
+   * Requirements:
+   *
+   * - Contract must be initialized.
+   *
+   * @return {[]} - Empty tuple.
+   */
+  circuit assertInitialized(): [] {
+    assert(_isInitialized, "Ownable: contract not initialized");
+  }
+
+  /**
+   * @description Asserts that the contract has not been initialized, throwing an error if it has.
+   *
+   * Requirements:
+   *
+   * - Contract must not be initialized.
+   *
+   * @return {[]} - Empty tuple.
+   */
+  circuit assertNotInitialized(): [] {
+    assert(!_isInitialized, "Ownable: contract already initialized");
   }
 
   /**
@@ -134,7 +167,7 @@ module Ownable {
    * @returns {Either<Bytes<32>, ContractAddress>} - The contract owner.
    */
   export circuit owner(): Either<Bytes<32>, ContractAddress> {
-    Initializable_assertInitialized();
+    assertInitialized();
     return _owner;
   }
 
@@ -158,7 +191,7 @@ module Ownable {
    * @returns {[]} Empty tuple.
    */
   export circuit transferOwnership(newOwner: Either<Bytes<32>, ContractAddress>): [] {
-    Initializable_assertInitialized();
+    assertInitialized();
     const isContractAddr = !newOwner.is_left;
     assert(!isContractAddr, "Ownable: unsafe ownership transfer");
     _unsafeTransferOwnership(newOwner);
@@ -184,7 +217,7 @@ module Ownable {
    * @returns {[]} Empty tuple.
    */
   export circuit _unsafeTransferOwnership(newOwner: Either<Bytes<32>, ContractAddress>): [] {
-    Initializable_assertInitialized();
+    assertInitialized();
     assertOnlyOwner();
     assert(!_isTargetZero(newOwner), "Ownable: invalid new owner");
     _unsafeUncheckedTransferOwnership(newOwner);
@@ -205,7 +238,7 @@ module Ownable {
    * @returns {[]} Empty tuple.
    */
   export circuit renounceOwnership(): [] {
-    Initializable_assertInitialized();
+    assertInitialized();
     assertOnlyOwner();
 
     // We use `left` here because `assertOnlyOwner` would return a misleading
@@ -233,7 +266,7 @@ module Ownable {
    * @returns {[]} Empty tuple.
    */
   export circuit assertOnlyOwner(): [] {
-    Initializable_assertInitialized();
+    assertInitialized();
     if (_owner.is_left) {
       assert(_computeAccountId() == _owner.left, "Ownable: caller is not the owner");
     } else {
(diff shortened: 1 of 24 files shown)


contracts/src/access/Ownable.compact
```diff
@@ -85,9 +85,15 @@ pragma language_version >= 0.21.0;
...
   import CompactStandardLibrary;
-  import "../security/Initializable" prefix Initializable_;
   import "../utils/Utils" prefix Utils_;
 
+  /**
+   * @description Initialization flag, tracked per-module to avoid the compiler's
+   * shared transitive-dependency state bug (LFDT-Minokawa/compact#270). See the
+   * Initializable module for rationale.
+   */
+  export ledger _isInitialized: Boolean;
```
andrew-fleming, line 95: 👍

contracts/src/security/Initializable.compact
```diff
@@ -6,6 +6,24 @@ pragma language_version >= 0.21.0;
...
+ * their `_isInitialized` flags collapse into a SINGLE shared ledger slot: calling
+ * `initialize()` on one marks the other as initialized, and the other can never be
+ * initialized independently. Modules in different directories get separate slots.
+ * The result is unpredictable and a security hazard for composed contracts.
+ *
+ * For this reason the OpenZeppelin modules no longer use this module: each tracks
+ * its own `_isInitialized` ledger flag inline. This module is kept for direct,
+ * single-import use and can be reintroduced across modules once #270 is resolved
+ * (note: CMA upgrades cannot add/remove ledger variables, so already-deployed
+ * contracts cannot migrate their initialization layout).
```
andrew-fleming: I agree with being loud here about the issue, but this is only an problem with deploying contracts. If a contract is already deployed, it would already be initialized. Edge cases would be if users fail to follow the extensibility pattern and don't initialize during construction. IMO the highlighted section doesn't add much value

contracts/src/security/Initializable.compact
```diff
@@ -6,6 +6,24 @@ pragma language_version >= 0.21.0;
 /**
  * @module Initializable
  * @description Initializable provides a simple mechanism that mimics the functionality of a constructor.
+ *
+ * @notice ⚠️ TECHNICAL WARNING — DO NOT import this module into other OpenZeppelin
+ * modules, and avoid importing it into more than one module of a single contract.
```
andrew-fleming: I'd rephrase here because this is kind of confusing. Instead of what not to do (which is more directed toward the maintainers (us)), it might be more clear if it's prescriptive for users e.g. "Only use this at the top contract layer". WDYT?

contracts/test/integration/README.md
```diff
@@ -0,0 +1,33 @@
...
+- **`SharedInitCollision`** — two modules in the same directory that both import
+  the shared, stateful `Initializable`. The compiler deduplicates the transitive
+  `_isInitialized` into one ledger slot, so initializing one module initializes
+  the other. The spec's "the bug" block asserts that collision.
+- **`ComposedTokens`** — the production `FungibleToken` + `NonFungibleToken`
+  (same directory), each now owning its `_isInitialized` flag. The "the fix"
+  block asserts the two initializations are independent.
+
+If the compiler ever isolates transitive ledger state (compiler#270 resolved),
+the "the bug" block must be inverted or removed.
```
andrew-fleming: **folllowup**: This is fine for now since this is new and small, but this is going to get very messy as we populate this with more scenarios. Not a right-now thing, but we should consider a better organizational strat otherwise the readme will be unreadable

andrew-fleming (commented):
Great work, @0xisk! I left a few minor comments. Also, don't forget to remove `compact:security` from `dependsOn` in turbo.json (for all the `compact:<target>` tasks since Initializable is no longer a dep for the other modules)

contracts/test/integration/README.md
0xisk: Well I thought of really making it very simple for now.

contracts/src/security/Initializable.compact
0xisk: Agree.

0xisk:
Thanks @andrew-fleming! Pushed the fixes:

- **Rephrase `Initializable` warning to be prescriptive + drop CMA note** — bd2353a
- **Drop `compact:security` from `compact:access` / `compact:token` deps in `turbo.json`** — 1716e49
- **Cover new per-module init guards (`_spendAllowance`, `_unsafeMint`)** — d38b451
- **Trim integration README to high-level (also drops the hardcoded `0.29.0`)** — bab6be8

Notes:
- **FungibleToken re-init test:** skipped. `initialize` writes the sealed `_decimals`/`_name`/`_symbol`, so the compiler forbids exposing it as a callable circuit — re-init can't occur post-construction, nothing to test.
- **MultiToken guard reorder (CodeRabbit):** skipped. `_unsafeTransferFrom`/`_unsafeTransfer`/`_unsafeMint` already `assertInitialized()`; adding it to the wrappers would double-assert on-circuit.
- **README org strategy followup:** addressed by trimming now; happy to revisit a per-scenario structure when more scenarios land.

contracts/test/integration/README.md
andrew-fleming: Haha not a right now thing! But it's done :) we can better organize later

andrew-fleming (approved):
LGTM!

contracts/test/integration/README.md
0xisk: A perfect case for a low priority followup :)

PR #566, opened 2026-06-09
refactor(multisig): make forwarder parent type-safe

### LINKED ISSUE #565: dev: address remaining Forwarder review feedback from #526
## Context

Follow-up to #526 (`feat(multisig): add Forwarder + ForwarderPrivate modules`).
That PR is approved and ready to merge. Three review threads from @andrew-fleming
remain open and are deferred here so #526 is not held open. The work will land in
a dedicated PR stacked on top of #526.

All paths below are under `contracts/src/multisig/`.

### 1. Make the parent type-safe by splitting the generic module

[Thread](https://github.com/OpenZeppelin/compact-contracts/pull/526#discussion_r3378294654)

`Forwarder<T>` rebuilds the recipient at the send site
(`ZswapCoinPublicKey { bytes: _parent.bytes }` / `UserAddress { bytes: _parent.bytes }`).
Nothing stops a user from instantiating `Forwarder<ContractAddress>`; the encoding
differs, so coins would be wrapped under the wrong `Either` arm and forwarded to the
wrong recipient.

**Proposed:** split into two concrete modules that store the parent as its real
recipient `Either<>` and pass it straight to the send call (no `.bytes` rebuild, no
hard-coded `left`/`right`):

- `ForwarderShielded` — `_parent: Either<ZswapCoinPublicKey, ContractAddress>`
- `ForwarderUnshielded` — `_parent: Either<ContractAddress, UserAddress>`

A single generic `Either<T1, T2>` can't serve both deposit kinds, so the split is
the type-safe path. Presets and the test mock/simulators update to the explicit
`Either<>` API.

### 2. Don't enforce parent immutability on the public Forwarder

[Thread](https://github.com/OpenZeppelin/compact-contracts/pull/526#discussion_r3377932919)

`_parent` is `sealed` and the docs claim "immutable after init". The module should
not force immutability: a preset that won't change the parent simply omits a setter;
one that wants to can add it.

**Proposed:** drop `sealed`, remove the immutability claims from the docs.
`ForwarderPrivate` stays `sealed` — its commitment is the sole drain gate and the
preset already documents "deploy a new forwarder to change parent" as deliberate.

### 3. Consolidate the empty witness files

[Thread](https://github.com/OpenZeppelin/compact-contracts/pull/526#discussion_r3378156969)

Five identical `Record<string, never>` witness files exist for the forwarder
contracts.

**Proposed:** define the empty private state + witnesses once under
`multisig/test/` and import it into the simulators; delete the per-contract files.

## Acceptance criteria

- [ ] Generic `Forwarder<T>` replaced by concrete `ForwarderShielded` /
      `ForwarderUnshielded` modules storing `Either<>` parents.
- [ ] No `sealed` on the public forwarder parent; docs no longer claim immutability.
      `ForwarderPrivate` unchanged.
- [ ] Empty witnesses defined once and shared by the forwarder simulators; redundant
      files removed.
- [ ] Module + preset tests pass (`vitest run Forwarder`).
- [ ] All three threads on #526 resolved.

## Dependencies

Depends on #526. The fix PR targets `post-release` and is stacked on the `feat/forwarder` branch.

0xisk:
> [!IMPORTANT]
> **Depends on #526** (`feat/forwarder`). This branch is stacked on it.
> The diff will also show #526's commits until that PR merges into
> `post-release`; review only the `refactor(multisig): make forwarder
> parent type-safe` commit here.

## Types of changes

Fixes #565

Addresses the three remaining review threads from @andrew-fleming on #526.
The forwarder is not yet released, so these refine unmerged code.

- **Type safety.** The generic `Forwarder<T>` rebuilt the recipient from
  `_parent.bytes` and wrapped it in a hard-coded `Either` arm, so
  instantiating it with the wrong address kind silently mis-encoded the
  recipient. It is replaced by two concrete modules that store the real
  recipient type and pass it straight to the send call:
  - `ForwarderShielded` — `Either<ZswapCoinPublicKey, ContractAddress>`
  - `ForwarderUnshielded` — `Either<ContractAddress, UserAddress>`

  The deployer now chooses the recipient arm explicitly; there is no way
  to mis-encode it.
- **Immutability.** Dropped `sealed` from the public forwarder parent and
  removed the "immutable after init" claim. A preset that keeps the
  parent fixed simply omits a setter; a consuming contract may add one.
  `ForwarderPrivate` stays sealed (its commitment is the sole drain gate).
- **Test cleanup.** Empty private state + witnesses defined once in
  `multisig/test/EmptyWitnesses.ts` and shared by the forwarder
  simulators; five redundant per-contract witness files removed. The
  combined mock is split into `MockForwarderShielded` /
  `MockForwarderUnshielded` (the two modules cannot share one
  `Initializable` instance).

## PR Checklist

#### Further comments

A single `Either<T1, T2>` generic could not serve both deposit kinds
(shielded and unshielded need different arm orderings), so the two-module
split is the type-safe path Andrew preferred. All forwarder module +
preset suites pass (`vitest run Forwarder`, 40 tests); `biome` and `tsc`
are clean.

## Summary by CodeRabbit

* **New Features**
  * Introduced ForwarderShielded and ForwarderUnshielded modules enabling atomic value forwarding to configurable parent recipients.
  * Added initialization guards ensuring proper contract setup before deposit operations.

* **Improvements**
  * Preset contracts now support flexible parent recipient types.


contracts/src/multisig/test/Forwarder.test.ts
andrew-fleming, line 1: 🔵 **followup:** we should add tests for when the parent is a contract address (including a zero address failure case)

contracts/src/multisig/test/EmptyWitnesses.ts
andrew-fleming, line 1: 👍

contracts/src/multisig/ForwarderShielded.compact
```diff
@@ -0,0 +1,83 @@
...
+ * gated by the `Initializable` module: every circuit asserts the module
+ * has been initialized via `initialize`.
+ *
+ * The parent is set once at init and the module exposes no setter, so a
+ * preset that omits one keeps the parent fixed. A consuming contract
+ * that needs to change the parent may add its own setter circuit.
+ */
+module ForwarderShielded {
+  import CompactStandardLibrary;
+  import "../security/Initializable" prefix Initializable_;
```
andrew-fleming: Since these are new modules, should we just get rid of this now before we merge?

contracts/src/multisig/ForwarderUnshielded.compact
```diff
@@ -0,0 +1,87 @@
...
+ * gated by the `Initializable` module: every circuit asserts the module
+ * has been initialized via `initialize`.
+ *
+ * The parent is set once at init and the module exposes no setter, so a
+ * preset that omits one keeps the parent fixed. A consuming contract
+ * that needs to change the parent may add its own setter circuit.
+ */
+module ForwarderUnshielded {
+  import CompactStandardLibrary;
+  import "../security/Initializable" prefix Initializable_;
```
andrew-fleming: Same question here

contracts/test-utils/address.ts
```diff
@@ -111,3 +113,34 @@ export const ZERO_ADDRESS = {
...
+): Either<ContractAddress, UserAddress> => ({
+  is_left: false,
+  left: encodeToAddress(''),
+  right: encodeToUserAddress(str),
+});
+
+export const ZERO_USER_ADDRESS: Either<ContractAddress, UserAddress> = {
+  is_left: false,
+  left: encodeToAddress(''),
+  right: { bytes: zeroUint8Array() },
```
andrew-fleming, line 149: 👍

andrew-fleming (commented):
Changes look good, @0xisk! I just left two questions that I think we should address before merging

0xisk:
Thanks @andrew-fleming! Pushed the fixes:

- **Drop the `Initializable` import from `ForwarderShielded` / `ForwarderUnshielded`** — 734a9a1
  Each new module now owns its `_isInitialized` flag and inlines `assertInitialized` / `assertNotInitialized` with module-specific messages, so we don't ship the shared transitive-dependency pattern (compiler#270) in new code. Left `Signer` / `ForwarderPrivate` as-is.
- **followup: contract-address parent tests (+ zero-address failure)** — e520c26
  Both modules now cover a contract-address parent on init, a zero-contract-address failure, and parent read-back. Added `createEitherTestUnshieldedContract` / `ZERO_UNSHIELDED_CONTRACT` helpers for the unshielded contract arm.

Verified: `vitest run Forwarder` (40 tests), `tsc`, and `biome` all clean.

contracts/src/multisig/ForwarderUnshielded.compact
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * @module ForwarderUnshielded
+ * @description Public-parent forwarder for unshielded coins. Provides an
+ * atomic forward pattern: receive an unshielded amount of a given color
+ * and immediately send it to the configured parent recipient.
+ *
+ * The parent is stored as its real recipient type
+ * `Either<ContractAddress, UserAddress>` and passed straight to the send
+ * call. The deployer chooses the arm explicitly (a contract or a user),
+ * so there is no way to mis-encode the recipient.
```
andrew-fleming: I think we should include a warning that it's up to users to ensure that the parent is set up to handle funds; otherwise, they could be stuck. It might be worth mentioning in `initialize` as well if you agree

contracts/src/multisig/ForwarderUnshielded.compact
```diff
@@ -0,0 +1,122 @@
...
+   * @param {Either<ContractAddress, UserAddress>} parent - The parent
+   * recipient of every forwarded amount: a contract address (`left`) or
+   * a user address (`right`).
+   *
+   * @returns {[]} Empty tuple.
+   */
+  export circuit initialize(parent: Either<ContractAddress, UserAddress>): [] {
+    const isZero = parent.is_left
+      ? default<ContractAddress> == parent.left
+      : default<UserAddress> == parent.right;
```
andrew-fleming: **folllowup**: I feel like we can make this generic in Utils -> `isEitherZero<T1, T2>`

contracts/src/multisig/ForwarderUnshielded.compact
```diff
@@ -0,0 +1,122 @@
...
+    assert(!isZero, "ForwarderUnshielded: zero parent");
+    assertNotInitialized();
+    _isInitialized = true;
+    _parent = disclose(parent);
```
andrew-fleming:
```suggestion
    _parent = Utils_canonicalize<ContractAddress, UserAddress>(parent);
```
For consistency and to maintain a defensive stance

contracts/src/multisig/ForwarderShielded.compact
```diff
@@ -0,0 +1,118 @@
...
+   * parent recipient of every forwarded coin: a coin public key
+   * (`left`) or a contract address (`right`).
+   *
+   * @returns {[]} Empty tuple.
+   */
+  export circuit initialize(parent: Either<ZswapCoinPublicKey, ContractAddress>): [] {
+    assert(!Utils_isKeyOrAddressZero(parent), "ForwarderShielded: zero parent");
+    assertNotInitialized();
+    _isInitialized = true;
+    _parent = disclose(parent);
```
andrew-fleming: Same suggestion with canonicalize

andrew-fleming (commented):
I left a few comments but this looks about good to go. I think we should improve our tests by checking the zswap inputs and outputs to ensure the forwarder is forwarding correctly on the zswap level. Let's call it a *followup*. This is also true of the other modules that handle send and receive

0xisk:
> I left a few comments but this looks about good to go. I think we should improve our tests by checking the zswap inputs and outputs to ensure the forwarder is forwarding correctly on the zswap level. Let's call it a _followup_. This is also true of the other modules that handle send and receive

I agree and the best way for doing this is by hacing the simulator runs those tests on the local node infra so we are more sure about that.

0xisk:
### Forwarder recipient finding + rebased on `main`

Confirmed on preprod (shielded case) that **an atomic forward to a contract recipient cannot be delivered**. A shielded/unshielded send to a contract is valid only if that contract claims the output in the *same* transaction. An atomic forwarder runs only its own circuit, so a third-party contract parent never claims it and the deposit is rejected (unclaimed output, node error `186`).

**Fix (pushed):**
- `ForwarderShielded` / `ForwarderUnshielded`: narrow the circuit parameter to the deliverable arm (`ZswapCoinPublicKey` / `UserAddress`), so a contract parent can't be expressed and no forwarder can be deployed in a brick-on-deposit state.
- `_parent` stays a generic `Either` on the ledger. A future CMA *circuit* upgrade can then add contract delivery (if cross-contract UTXO claiming lands) without a *state-layout* migration.
- `ForwarderPrivate`: coin-public-key only, for the same reason plus a contract recipient would leak the address in cleartext.

Rebased onto `main` (incl. #610); `ForwarderPrivate` now uses the inline `_isInitialized` flag too.

This supersedes the earlier `canonicalize` / `isEitherZero<T1, T2>` / contract-address-parent-test suggestions, which no longer apply to the narrowed param. I'll handle the remaining open item (warn that the parent must be able to spend the forwarded funds) next.

> Note: the unshielded case is reasoned from the same in-tx-claim rule; it isn't independently lab-confirmed like the shielded one.

contracts/src/multisig/ForwarderUnshielded.compact
0xisk: Added in 5445992: a `@notice` in both the module header and `initialize`, for ForwarderShielded and ForwarderUnshielded.

contracts/src/multisig/ForwarderShielded.compact
0xisk: Done. Dropped the `Initializable` dependency for the inline `_isInitialized` flag, aligned with #610 during the rebase.

contracts/src/multisig/ForwarderUnshielded.compact
0xisk: Done. Same inline `_isInitialized` flag here too, via the #610 rebase.

contracts/src/multisig/test/Forwarder.test.ts
0xisk: Superseded by 487964e: the parent param is narrowed to the deliverable arm, so a contract-address parent is no longer expressible. Kept the zero-parent failure case for the working arm.

contracts/src/multisig/ForwarderUnshielded.compact
0xisk: Superseded by 487964e: the param is now a single arm (`UserAddress`), so the zero-check is on one value rather than an `Either`. No generic helper needed.

contracts/src/multisig/ForwarderUnshielded.compact
0xisk: Superseded by 487964e: `initialize` now stores `right<ContractAddress, UserAddress>(parent)`, which zeroes the inactive arm by construction, so there is no dual-arm input left to canonicalize.

contracts/src/multisig/ForwarderShielded.compact
0xisk: Same as above: `initialize` stores `left<ZswapCoinPublicKey, ContractAddress>(parent)`, canonical by construction (487964e).

contracts/src/multisig/test/ForwarderPrivate.test.ts
```diff
@@ -214,6 +237,51 @@ describe('ForwarderPrivate module', () => {
...
+  // written only at init; it is unchanged after a drain and no recipient field
+  // is added. (Read via the getter circuit — the module is imported with a
+  // prefix only, so it is not in the public ledger reader.)
+  //
+  // INV-17 (recipient privacy): the parent coin public key flows only into the
+  // `sendShielded` recipient, where it is encrypted inside the Zswap output and
+  // never appears on the public transcript. Confirmed end-to-end on preprod (a
+  // coin-public-key recipient occurs 0 times in the published tx); not
+  // simulator-observable, so it is asserted by the residual-surface check here.
+  describe('drain — residual public surface (INV-12 / INV-17 / INV-25)', () => {
```
andrew-fleming, line 265: 👍

contracts/src/multisig/ForwarderUnshielded.compact
```diff
@@ -0,0 +1,152 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.0.1-alpha.1 (multisig/ForwarderUnshielded.compact)
```
andrew-fleming, line 2:
```suggestion
// OpenZeppelin Compact Contracts v0.2.0 (multisig/ForwarderUnshielded.compact)
```

contracts/src/multisig/ForwarderShielded.compact
```diff
@@ -0,0 +1,149 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.0.1-alpha.1 (multisig/ForwarderShielded.compact)
```
andrew-fleming, line 2:
```suggestion
// OpenZeppelin Compact Contracts v0.2.0 (multisig/ForwarderShielded.compact)
```

contracts/src/multisig/test/mocks/MockForwarderShielded.compact
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.0.1-alpha.1 (multisig/test/mocks/MockForwarderShielded.compact)
```
andrew-fleming, line 2:
```suggestion
// OpenZeppelin Compact Contracts v0.2.0 (multisig/test/mocks/MockForwarderShielded.compact)
```

andrew-fleming: **followup**: We should also apply this to all the mocks and remove the version

```
// WARNING: FOR TESTING PURPOSES ONLY.
// This contract exposes internal circuits and bypasses safety checks that the
// corresponding production contract relies on. DO NOT deploy or use this
// contract in any production application.
```

contracts/src/multisig/test/mocks/MockForwarderUnshielded.compact
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.0.1-alpha.1 (multisig/test/mocks/MockForwarderUnshielded.compact)
```
andrew-fleming, line 2:
```suggestion
// OpenZeppelin Compact Contracts v0.2.0 (multisig/test/mocks/MockForwarderUnshielded.compact)
```

andrew-fleming (approved):
Great work, @0xisk! Non-blocking final suggestions. We can apply the suggestions in a new PR to avoid an approve + merge lag. Your call

PR #616, opened 2026-06-19
fix(token): resolve zero-value revert audit findings (L-01, L-02)

### LINKED ISSUE #425: L-01: Zero-value transfers may revert outside an assertion
**Source:** OpenZeppelin Midnight - Compact Contracts Audit

## Description

The `_update` function in the `MultiToken` contract [reduces the source account's balance](https://github.com/OpenZeppelin/compact-contracts/blob/d8ee045f30153f49b11b82a2f891066ee402a25f/contracts/src/token/MultiToken.compact#L320) without checking if the token `id` record has been initialized. In practice, this will only occur when the `value` is set to zero. This means a zero-value [transfer](https://github.com/OpenZeppelin/compact-contracts/blob/d8ee045f30153f49b11b82a2f891066ee402a25f/contracts/src/token/MultiToken.compact#L415), [mint](https://github.com/OpenZeppelin/compact-contracts/blob/d8ee045f30153f49b11b82a2f891066ee402a25f/contracts/src/token/MultiToken.compact#L502) or [burn](https://github.com/OpenZeppelin/compact-contracts/blob/d8ee045f30153f49b11b82a2f891066ee402a25f/contracts/src/token/MultiToken.compact#L528) of an uninitialized token will revert due to an invalid lookup, rather than with an explicit error message.

## Recommendation

Consider skipping the balance update when the `value` is zero.

### LINKED ISSUE #426: L-02: Non-Existent and Zero Allowances Have Inconsistent Behavior
**Source:** OpenZeppelin Midnight - Compact Contracts Audit

## Description

`FungibleToken` stores allowances in the nested ledger map [`_allowances`](https://github.com/OpenZeppelin/compact-contracts/blob/d8ee045f30153f49b11b82a2f891066ee402a25f/contracts/src/token/FungibleToken.compact#L63-L64). The public [`allowance`](https://github.com/OpenZeppelin/compact-contracts/blob/d8ee045f30153f49b11b82a2f891066ee402a25f/contracts/src/token/FungibleToken.compact#L256-L266) circuit returns `0` when either the owner key or spender key is missing. However, [`transferFrom`](https://github.com/OpenZeppelin/compact-contracts/blob/d8ee045f30153f49b11b82a2f891066ee402a25f/contracts/src/token/FungibleToken.compact#L316-L324) forwards to [`_unsafeTransferFrom`](https://github.com/OpenZeppelin/compact-contracts/blob/d8ee045f30153f49b11b82a2f891066ee402a25f/contracts/src/token/FungibleToken.compact#L348-L359), which unconditionally calls `_spendAllowance(fromAddress, spender, value)`.

[`_spendAllowance`](https://github.com/OpenZeppelin/compact-contracts/blob/d8ee045f30153f49b11b82a2f891066ee402a25f/contracts/src/token/FungibleToken.compact#L597-L614) first asserts that both `_allowances.member(owner)` and the spender sub-map membership are true. As a result, `transferFrom(owner, to, 0)` reverts unless a prior approval created the relevant map entries, for example through [`_approve`](https://github.com/OpenZeppelin/compact-contracts/blob/d8ee045f30153f49b11b82a2f891066ee402a25f/contracts/src/token/FungibleToken.compact#L563-L579), even though the effective allowance is semantically `0`. This diverges from ERC-20-style expectations that a missing entry is equivalent to a zero allowance and can cause unexpected reverts in composed flows that legitimately compute a zero transfer amount.

## Recommendation

Consider bypassing allowance spending when `value == 0`, or treating missing allowance entries as `0` within `_spendAllowance` so that it matches the semantics of `allowance`. Consider also adding tests for zero-value `transferFrom` calls without a pre-existing allowance entry.

0xisk:
## Types of changes

Fixes #425
Fixes #426

Resolves the two audit findings (L-01, L-02) that were parked behind the
unconditional-downcast compiler limitation
[LFDT-Minokawa/compact#226](https://github.com/LFDT-Minokawa/compact/issues/226),
now resolved upstream. With conditional downcasts available, a zero-value
balance/allowance update can be guarded without the constraint-level failure
the audit comments described.

### L-01 — `MultiToken._update` (#425)

A zero-value `transferFrom`, `_transfer`, `_mint` or `_burn` of a token `id`
that was never initialized reverted on the bare `_balances.lookup(id)` write
inside `_update`, rather than failing with an explicit assertion or succeeding
as a no-op.

**Fix:** guard both balance writes on `disclose(value > 0)`. A zero-value
update leaves balances unchanged, so the source write is skipped and the
recipient block no longer initializes the `id`. A non-zero transfer/burn of an
uninitialized id still fails with the explicit `"MultiToken: insufficient
balance"` assertion as before.

### L-02 — `FungibleToken._spendAllowance` (#426)

`transferFrom(owner, to, 0)` reverted unless a prior approval had created the
`_allowances` entry, because `_spendAllowance` asserted membership before
reading the allowance. This diverged from `allowance`, which treats a missing
entry as zero.

**Fix:** read the current allowance via `allowance` (missing entry → `0`) and
guard the spend on `disclose(value > 0)`. A zero-value spend is now a no-op that
neither reverts nor writes a spurious zero entry; a non-zero spend without
allowance still fails with `"FungibleToken: insufficient allowance"`.

### On `disclose(value > 0)`

The guard gates a ledger write, so its boolean must be disclosed. `value` is
already public on every balance write (and a non-zero spend already discloses
the updated allowance), so disclosing *whether* `value` is zero leaks nothing
new.

### Note on `@circuitInfo`

The touched circuits' constraint counts shift by a few rows. The `@circuitInfo`
annotations are intentionally left unchanged here: the existing values do not
match the toolchain in this environment even for circuits this PR does not touch
(e.g. `balanceOf`), so they are best regenerated wholesale with the canonical
release toolchain rather than edited piecemeal in this fix.

## PR Checklist

- [ ] I have added documentation for new methods or changes to existing method behavior. See [GUIDELINES.md#documentation](../GUIDELINES.md#documentation) for more information.

#### Further comments

Both fixes compile to valid ZKIR (prover/verifier keys generate for every mock
circuit), confirming the conditional downcast + disclosure is provable. Added
tests cover: zero-value `transferFrom` / `_transfer` / `_mint` / `_burn` on an
uninitialized `id` (MultiToken), and zero-value `transferFrom` / `_spendAllowance`
without a pre-existing allowance plus an existing-allowance-unchanged case
(FungibleToken). All 554 token suite tests pass.

## Bug Fixes
* Token transfers with zero amounts no longer fail when allowance entries don't exist; missing allowances are treated as zero.
* Zero-value balance operations no longer create unnecessary state entries.

## Tests
* Added comprehensive test coverage for zero-amount token operations with missing allowance entries.
* Extended tests for zero-value transfers, mints, and burns on uninitialized tokens.

### DIFF
--- a/contracts/src/token/FungibleToken.compact
+++ b/contracts/src/token/FungibleToken.compact
@@ -688,16 +688,19 @@ module FungibleToken {
     const canonOwner = Utils_canonicalize<Bytes<32>, ContractAddress>(owner);
     const canonSpender = Utils_canonicalize<Bytes<32>, ContractAddress>(spender);
 
-    assert((_allowances.member(disclose(canonOwner)) &&
-            _allowances.lookup(canonOwner).member(disclose(canonSpender))),
-           "FungibleToken: insufficient allowance"
-           );
-
-    const currentAllowance = _allowances.lookup(canonOwner).lookup(disclose(canonSpender));
-    const MAX_UINT128 = 340282366920938463463374607431768211455;
-    if (currentAllowance < MAX_UINT128) {
-      assert(currentAllowance >= value, "FungibleToken: insufficient allowance");
-      _approve(canonOwner, canonSpender, currentAllowance - value as Uint<128>);
+    // A missing allowance entry is treated as a zero allowance, mirroring
+    // `allowance`, and a zero-value spend is a no-op. Guarding on `value`
+    // avoids reverting (outside an assertion) on a missing entry and avoids
+    // writing a spurious zero entry for a zero-value spend (see audit L-02).
+    // Whether `value` is zero is already public (a non-zero spend discloses the
+    // updated allowance), so disclosing the guard leaks nothing new.
+    if (disclose(value > 0)) {
+      const currentAllowance = allowance(canonOwner, canonSpender);
+      const MAX_UINT128 = 340282366920938463463374607431768211455;
+      if (currentAllowance < MAX_UINT128) {
+        assert(currentAllowance >= value, "FungibleToken: insufficient allowance");
+        _approve(canonOwner, canonSpender, currentAllowance - value as Uint<128>);
+      }
     }
   }
 
--- a/contracts/src/token/FungibleToken.compact
+++ b/contracts/src/token/FungibleToken.compact
@@ -691,9 +691,7 @@ module FungibleToken {
     // A missing allowance entry is treated as a zero allowance, mirroring
     // `allowance`, and a zero-value spend is a no-op. Guarding on `value`
     // avoids reverting (outside an assertion) on a missing entry and avoids
-    // writing a spurious zero entry for a zero-value spend (see audit L-02).
-    // Whether `value` is zero is already public (a non-zero spend discloses the
-    // updated allowance), so disclosing the guard leaks nothing new.
+    // writing a spurious zero entry for a zero-value spend.
     if (disclose(value > 0)) {
       const currentAllowance = allowance(canonOwner, canonSpender);
       const MAX_UINT128 = 340282366920938463463374607431768211455;
--- a/contracts/src/token/MultiToken.compact
+++ b/contracts/src/token/MultiToken.compact
@@ -425,11 +425,19 @@ module MultiToken {
     if (!Utils_isTargetZero(disclose(canonFrom))) {
       const fromBalance = balanceOf(canonFrom, id);
       assert(fromBalance >= value, "MultiToken: insufficient balance");
-      const newBalance = fromBalance - value;
-      _balances.lookup(id).insert(disclose(canonFrom), disclose(newBalance));
+      // Skip the write when `value` is zero. A zero-value update leaves the
+      // balance unchanged, so there is nothing to write, and guarding here
+      // avoids `_balances.lookup(id)` reverting outside an assertion when `id`
+      // has never been initialized (see audit L-01). Whether `value` is zero is
+      // already public (every balance write discloses it), so disclosing the
+      // guard leaks nothing new.
+      if (disclose(value > 0)) {
+        const newBalance = fromBalance - value;
+        _balances.lookup(id).insert(disclose(canonFrom), disclose(newBalance));
+      }
     }
 
-    if (!Utils_isTargetZero(disclose(canonTo))) {
+    if (!Utils_isTargetZero(disclose(canonTo)) && disclose(value > 0)) {
       if (!_balances.member(disclose(id))) {
         _balances.insert(
           disclose(id),
--- a/contracts/src/token/MultiToken.compact
+++ b/contracts/src/token/MultiToken.compact
@@ -428,9 +428,7 @@ module MultiToken {
       // Skip the write when `value` is zero. A zero-value update leaves the
       // balance unchanged, so there is nothing to write, and guarding here
       // avoids `_balances.lookup(id)` reverting outside an assertion when `id`
-      // has never been initialized (see audit L-01). Whether `value` is zero is
-      // already public (every balance write discloses it), so disclosing the
-      // guard leaks nothing new.
+      // has never been initialized.
       if (disclose(value > 0)) {
         const newBalance = fromBalance - value;
         _balances.lookup(id).insert(disclose(canonFrom), disclose(newBalance));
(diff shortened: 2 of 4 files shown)


contracts/src/token/FungibleToken.compact
```diff
@@ -688,16 +688,19 @@ module FungibleToken {
...
-    const MAX_UINT128 = 340282366920938463463374607431768211455;
-    if (currentAllowance < MAX_UINT128) {
-      assert(currentAllowance >= value, "FungibleToken: insufficient allowance");
-      _approve(canonOwner, canonSpender, currentAllowance - value as Uint<128>);
+    // A missing allowance entry is treated as a zero allowance, mirroring
+    // `allowance`, and a zero-value spend is a no-op. Guarding on `value`
+    // avoids reverting (outside an assertion) on a missing entry and avoids
+    // writing a spurious zero entry for a zero-value spend (see audit L-02).
+    // Whether `value` is zero is already public (a non-zero spend discloses the
+    // updated allowance), so disclosing the guard leaks nothing new.
```
andrew-fleming: The entire state is already public. What is there to leak?

contracts/src/token/MultiToken.compact
```diff
@@ -425,11 +425,19 @@ module MultiToken {
...
       const fromBalance = balanceOf(canonFrom, id);
       assert(fromBalance >= value, "MultiToken: insufficient balance");
-      const newBalance = fromBalance - value;
-      _balances.lookup(id).insert(disclose(canonFrom), disclose(newBalance));
+      // Skip the write when `value` is zero. A zero-value update leaves the
+      // balance unchanged, so there is nothing to write, and guarding here
+      // avoids `_balances.lookup(id)` reverting outside an assertion when `id`
+      // has never been initialized (see audit L-01). Whether `value` is zero is
+      // already public (every balance write discloses it), so disclosing the
+      // guard leaks nothing new.
```
andrew-fleming: Same question regarding leakage

contracts/src/token/test/FungibleToken.test.ts
```diff
@@ -1138,6 +1159,21 @@ describe('FungibleToken', () => {
         ).rejects.toThrow('FungibleToken: insufficient allowance');
       });
 
+      it('should allow a zero-value spend without a pre-existing allowance', async () => {
+        // A missing entry is treated as zero and a zero-value spend is a no-op,
+        // so this must not revert and must not create an entry (audit L-02).
```
andrew-fleming: I don't think it's a good idea to leave audit refs like this. It assumes there is and always will be one audit. I'd remove the `(audit L-02)` entirely. If you _really_ want to reference it, leave a full link to a reference (issue or audit page or something) so that this is not to be confused with a future audit

andrew-fleming: Ditto with the rest of the audit refs in the mods and tests

andrew-fleming (commented):
Looking good, @0xisk! Left some comments on the docs

contracts/src/token/FungibleToken.compact
0xisk: Agreed — the token state is fully public, so there is nothing to leak. Dropped the privacy framing (and the audit ref) in fe92578; the comment now just explains the zero-value guard.

contracts/src/token/MultiToken.compact
0xisk: Same fix in fe92578: removed the leakage framing and the audit ref, kept the rationale for skipping the zero-value write.

contracts/src/token/test/FungibleToken.test.ts
0xisk: Done in fe92578: removed all inline `(audit L-01/L-02)` refs from the token modules and tests. Left the generic "author and audit their own witnesses" disclaimer in the witness files, since that is not a finding reference.

0xisk:
@andrew-fleming addressed your review in `fe92578`:

- Dropped the "already public … leaks nothing new" framing from the `_spendAllowance` (FungibleToken) and `_transfer` (MultiToken) zero-value guards — the state is fully public, so there is nothing to leak.
- Removed every inline `(audit L-01/L-02)` ref from the token modules and tests.

andrew-fleming (approved):
LGTM!

PR #621, opened 2026-06-23
feat(token): add native shielded token standard `MIP-0011`

### LINKED ISSUE #544: Add Native Shielded Token standard `MIP-11`
**🧐 Motivation**
The library has contract-based token standards (`FungibleToken`, `Shielded Contract-Based Tokens`) but none for **native shielded tokens** — Zswap shielded UTXO coins. This is the primitive the Monument use case needs (phase-one mint/burn), and the base for compliant RWA tokens later.

**📝 Details**
Standard for minting/burning native shielded Zswap tokens, with a modular access-control layer (Ownable / AccessControl / Multisig) chosen by the consumer. Ledger state must be designed up front (not upgradable; see CMA). Phase-one scope = mint/burn, unblocked now; phase-two transfer depends on contract-to-contract + custom-spend logic.
Roadmap: Libraries → Shielded Native Token (Priority MNF: high).

### LINKED ISSUE #646: Add NST Core
**🧐 Motivation**

**📝 Details**

0xisk:
## Types of changes

Adds the native shielded (Zswap) token standard — the single-token
`NativeShieldedToken` and the multi-type `NativeShieldedTokenFamily` — with
supply accounting factored out as an **opt-in extension** rather than a core
feature.

### Why supply is an extension

Tracking supply on-chain is a privacy trade-off, not just modularity. A
contract-mediated burn is **amount-private** on its own: the coin operations
(`receiveShielded` / `sendImmediateShielded` / `sendShielded`) emit only
commitments and nullifiers, and the `disclose()` wrappers they require are
compiler permission markers, not disclosure sinks. The only thing that writes a
burned amount into the public transcript is the supply-counter ledger write.

So:

- A bare `_burn` / `_burnFromContract` hides the burned amount.
- Composing the supply extension trades that privacy for on-chain auditability
  (`totalSupply()`).
- Mint amounts are public regardless, via the protocol-level `shieldedMints`
  effect — the extension changes burn visibility, not mint visibility.

### What's here

- `token/NativeShieldedToken`, `token/NativeShieldedTokenFamily` — core
  mint/burn, no supply state. The mint/burn circuits keep their coin ops and
  `disclose` wrappers.
- `token/extensions/NativeShieldedTokenSupply` (scalar) and
  `…/NativeShieldedTokenFamilySupply` (per-domain) — standalone accounting
  modules mirroring `NativeShieldedTokenDerivedNonce`. The consumer pairs
  `_addMinted` / `_addBurned` with the matching mint/burn op. Supply getters do
  not gate on init (a counter reading 0 pre-init is correct).
- `token/extensions/NativeShieldedTokenDerivedNonce` — caller-free nonce
  derivation.
- Unit tests (simulator) + integration specs (local stack), including a
  burn-amount privacy spec.

> The MIP/design docs and the dependency/build changes (`package.json`,
> `yarn.lock`) are intentionally left out of this PR. The integration suite
> needs the `@midnight-ntwrk/*` dev-deps to install and run, so that CI job
> will be red until those land.

## PR Checklist

- [ ] I have added documentation for new methods or changes to existing method behavior.

## Summary by CodeRabbit

* **New Features**
  * Added support for native shielded tokens with token metadata, minting, burning, and supply tracking.
  * Introduced optional derived nonce handling for safer coin creation flows.
  * Added per-domain token support, enabling separate colors and supply accounting per domain.

* **Bug Fixes**
  * Strengthened validation for invalid recipients, wrong token colors, insufficient balances, and invalid burn refunds.
  * Improved replay and commitment-collision protections during minting.

* **Tests**
  * Added broad unit and integration coverage for mint, burn, supply, privacy, and initialization behavior.


contracts/src/token/NativeShieldedToken.compact
```diff
@@ -0,0 +1,395 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.1.0 (token/NativeShieldedToken.compact)
```
andrew-fleming:
```suggestion
// OpenZeppelin Compact Contracts v0.2.0 (token/NativeShieldedToken.compact)
```

andrew-fleming: Ditto with the other relevant files

contracts/src/token/NativeShieldedToken.compact
andrew-fleming, line 1: Sooo this module is almost an exact duplicate of the family variant. The three differences I see are:
- sealed domain in `initialize`
- `domain` param in token ops
- assert msg includes "Family" in prefix e.g. `NativeShieldedTokenFamily: ...`

I propose that we avoid duplication here and I see a couple of options:
1. Use family variant as the core. Here is what that looks like within the single token mod: https://gist.github.com/andrew-fleming/52991f1d47c3fb05f6247e34ffe5b010 . I benchmarked this and the circuit constraints are identical. This gets a little weird though because `name`, `symbol`, and `decimals` are imported and reexported as-is. We'd also need to make the assertion msgs generic (omit "Family" from the prefix). Finally, NST would import the family module without a prefix. See the gist for how this all looks
2. Create a separate `Core` module and use family and single NST as thin wrappers. This also breaks the lib's convention by having a shared module that's not meant to be exported to users. IMO this is the better approach. We just need to think about how this will be organized in the repo

The other option is to keep it as-is and duplicate the logic. If there's a bug or a change, it's additional maintenance and easy to miss/forget things. LMK what you think

andrew-fleming: Also, consider that the token variant option is binary, you either have a single token or family tokens. For that reason, the "Family" word in the prefix isn't a big deal. If you're using family and there's a `NativeShieldedToken: ...` issue, it's obvious where you need to debug

contracts/src/token/extensions/NativeShieldedTokenFamilySupply.compact
```diff
@@ -0,0 +1,153 @@
...
+   * exceed the overflow-checked minted total.
+   *
+   * @param {Bytes<32>} domain - 32-byte token discriminator.
+   * @param {Uint<128>} amount - The burned amount.
+   * @return {[]} - Empty tuple.
+   */
+  export circuit _addBurned(domain: Bytes<32>, amount: Uint<128>): [] {
+    const current = totalBurned(domain);
+    _totalBurned.insert(disclose(domain), disclose((current + amount) as Uint<128>));
+  }
```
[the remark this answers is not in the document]
andrew-fleming: +1

contracts/src/token/extensions/NativeShieldedTokenFamilySupply.compact
andrew-fleming, line 1: If you agree with avoiding repetition in the core modules, we can apply the same approach to the extensions

contracts/src/token/test/NativeShieldedToken.test.ts
```diff
@@ -0,0 +1,291 @@
...
+
+let token: NativeShieldedTokenSimulator;
+
+describe('NativeShieldedToken (Fungible profile)', () => {
+  describe('initialization', () => {
+    beforeEach(async () => {
+      token = await deploy(INIT);
+    });
+
+    it('should expose the constructor metadata (INV-14)', async () => {
```
andrew-fleming: Let's remove the INV refs

contracts/src/token/test/NativeShieldedToken.test.ts
```diff
@@ -0,0 +1,291 @@
...
+        'NativeShieldedToken: invalid recipient',
+      );
+    });
+
+    // INV-3 (Uint<128> overflow of totalMinted) — the guard
+    // `MAX_UINT128 - _totalMinted >= amount` is present, but with `amount`
+    // capped at Uint<64> it would take ~2^64 mints to approach the bound, so
+    // it is not reachable in a unit test without a state-injection hook. The
+    // guard itself is trivially correct; left unexercised by design.
+    it.skip('should revert on Uint<128> overflow of totalMinted (INV-3)', () => {});
```
andrew-fleming: totalMinted is not part of the core module

contracts/src/token/NativeShieldedToken.compact
```diff
@@ -0,0 +1,395 @@
...
+ *   emits the amount in the protocol-level `shieldedMints` effect, which no
+ *   amount of accounting choice can suppress. The extension changes burn
+ *   visibility, not mint visibility.
+ *
+ * - When the supply extension IS composed: `totalMinted()` is EXACT (color
+ *   derivation guarantees every coin of this contract's color originates from
+ *   this module's mints); `totalBurned()` is a LOWER BOUND (holders can send
+ *   coins straight to the burn address, bypassing the contract); and
+ *   `totalSupply()` = minted - burned is an UPPER BOUND on circulating supply.
+ *   Exact circulating supply is not knowable for native shielded tokens.
```
andrew-fleming: I'd remove this point. The doc should express what this module does, not the effects of things that are not included. Leave that in the relevant extension doc

contracts/src/token/NativeShieldedToken.compact
```diff
@@ -0,0 +1,395 @@
...
+   * - Contract is initialized.
+   * - `coin.color` is this contract's token color.
+   * - `amount` is less than or equal to `coin.value`.
+   *
+   * @param {QualifiedShieldedCoinInfo} coin - The contract-held coin to burn from.
+   * @param {Uint<128>} amount - Value to destroy. Must be <= `coin.value`.
+   * @return {Maybe<ShieldedCoinInfo>} - The change coin retained by the
+   * contract, or `none` if the coin was burned in full.
+   */
+  export circuit _burnFromContract(
```
andrew-fleming: Looking at this in an impl, I wonder if `burnFromSelf` would be a better name. The rationale being that "contract" is vague: I read it as "if the implementing contract, or any other contract that holds coins from this contract, wants to burn, they should use this circuit". This would be a standard proposal change

contracts/vitest.integration-net.config.ts
andrew-fleming, line 1: Integration tests are failing I assume bc of the sim update

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
```diff
@@ -0,0 +1,109 @@
...
+   *
+   * Requirements:
+   *
+   * - The chain is not already seeded.
+   * - `initNonce` is not zero (a zero seed is indistinguishable from an
+   *   unseeded chain).
+   *
+   * @param {Bytes<32>} initNonce - Nonce chain seed; subsequent values evolve
+   * from it via the internal counter. Choose unpredictably (e.g. random bytes).
+   * @return {[]} - Empty tuple.
```
andrew-fleming: What does it matter if it's `123` or unpredictable "random bytes"? I'm not sure I'm following

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
```diff
@@ -0,0 +1,109 @@
...
+   * @return {Bytes<32>} - The next derived coin nonce.
+   */
+  export circuit _deriveNonce(): Bytes<32> {
+    assert(_nonce != default<Bytes<32>>, "NativeShieldedTokenDerivedNonce: chain not seeded");
+    _counter.increment(1);
+    const chainValue = evolveNonce(_counter, _nonce);
+    _nonce = chainValue;
+    return persistentHash<Vector<2, Bytes<32>>>(
+      [pad(32, "NativeShieldedToken:nonce"), chainValue]
+    );
```
andrew-fleming: Sooooo I'm wondering if we can remove `evolveNonce` here. The only value I see in it is that it makes the nonce look random when this variant is explicitly not. Same thing with the domain...if nonce and counter are public and deterministic, what value does the expensive hasher actually have?

contracts/src/token/extensions/NativeShieldedTokenFamilySupply.compact
```diff
@@ -0,0 +1,153 @@
...
+   * @dev Checks for overflow in order to output a readable error message.
+   *
+   * @param {Bytes<32>} domain - 32-byte token discriminator.
+   * @param {Uint<64>} amount - The minted amount.
+   * @return {[]} - Empty tuple.
+   */
+  export circuit _addMinted(domain: Bytes<32>, amount: Uint<64>): [] {
+    const current = totalMinted(domain);
+    const MAX_UINT128 = 340282366920938463463374607431768211455;
+    assert(MAX_UINT128 - current >= amount, "NativeShieldedTokenFamilySupply: arithmetic overflow");
```
andrew-fleming: We're checking for overflow to provide a nice error msg in the event of an overflow.`_addBurned` overflow and totalSupply underflow rely on the compiler (less-nice error msgs). Should we improve the others?

andrew-fleming: Also, we can get `MAX_UINT128` from utils

contracts/src/token/extensions/NativeShieldedTokenFamilySupply.compact
```diff
@@ -0,0 +1,153 @@
...
+  /**
+   * @description Adds `amount` to the exact minted total for `domain`. Call
+   * once per successful `NativeShieldedTokenFamily._mint`.
+   * @dev Checks for overflow in order to output a readable error message.
+   *
+   * @param {Bytes<32>} domain - 32-byte token discriminator.
+   * @param {Uint<64>} amount - The minted amount.
+   * @return {[]} - Empty tuple.
+   */
+  export circuit _addMinted(domain: Bytes<32>, amount: Uint<64>): [] {
```
andrew-fleming: It might be worth documenting that this module does not catch mint-side drifting (duplicated `_addMinted` or a wrong domain)

contracts/src/token/test/NativeShieldedToken.test.ts
```diff
@@ -0,0 +1,291 @@
...
+      expect(color.length).toBe(32);
+      // Stable across calls (same domain + same contract address).
+      expect(await token.tokenColor()).toEqual(color);
+    });
+
+    it('should start with zero supply counters (INV-2, INV-4)', async () => {
+      expect(await token.totalMinted()).toBe(0n);
+      expect(await token.totalBurned()).toBe(0n);
+      expect(await token.totalSupply()).toBe(0n);
+    });
```
andrew-fleming: This should be a standalone file (`NativeShieldedTokenSupply`) for unit testing...one test file per module. We can integrate it with NST for integration tests

andrew-fleming: I'm thinking that the `test/` directory should match the structure it's testing e.g. `test/extensions/NativeShieldedDerivedNonce.test.ts`. Otherwise, this looks pretty busy. Maybe there's a better solution

contracts/src/token/test/NativeShieldedToken.test.ts
```diff
@@ -0,0 +1,291 @@
...
+      // token module's init flag; init enforcement stays on _mint/_burn.
+      expect(await token.totalMinted()).toBe(0n);
+      expect(await token.totalBurned()).toBe(0n);
+      expect(await token.totalSupply()).toBe(0n);
+    });
+
+    it('should revert _deriveNonce before the chain is seeded (INV-13)', async () => {
+      await expect(token._deriveNonce()).rejects.toThrow(
+        'NativeShieldedTokenDerivedNonce: chain not seeded',
+      );
```
andrew-fleming: Flagging this as well 🎌 these tests should be scoped to `NativeShieldedToken` as that's the file name

contracts/test/integration/_harness/cma.ts
andrew-fleming, line 1: Is this needed? If so, I'd say that the CMA tests PR should be prioritized and merged first

andrew-fleming (commented):
Excellent work, @0xisk! Partial review: I left some comments and questions. Irrespective of how big of a refactor we do, if any, I'd really consider moving each extension to its own PR (as you proposed offline). I'll review the integration tests once they're working

contracts/src/token/NativeShieldedToken.compact
0xisk: Great points Andrew! I agree with all, I am leaning to option 2.

contracts/src/token/NativeShieldedToken.compact
0xisk: Makes sense! we will need also to update it on the MIP.

contracts/src/token/extensions/NativeShieldedTokenFamilySupply.compact
0xisk: Both done in `f27d85a`: `_addMinted` uses `Utils_UINT128_MAX()`, and `_addBurned` enforces `burned <= minted`, which gives a readable error and makes the overflow/underflow paths unreachable.

contracts/src/token/NativeShieldedToken.compact
0xisk: Done in `f27d85a`: all native shielded token modules now carry v0.2.0 headers.

contracts/src/token/NativeShieldedToken.compact
0xisk: Done in `f27d85a`. Extracted a shared `NativeShieldedTokenCore`; NST and Family are thin wrappers (single token forwards a sealed `_domain`, family forwards the per-call `domain`). Took the separate-core option.

contracts/src/token/extensions/NativeShieldedTokenFamilySupply.compact
0xisk: Done in `f27d85a`. Same approach for the extensions: a shared `NativeShieldedTokenSupplyCore`, with `NativeShieldedTokenSupply` (scalar, single fixed key) and `NativeShieldedTokenFamilySupply` (per-domain) as wrappers.

contracts/src/token/NativeShieldedToken.compact
0xisk: Done in `f27d85a`. Dropped the supply-accounting section from the module doc; it now points to the extension.

contracts/src/token/NativeShieldedToken.compact
0xisk: Done in `f27d85a`. Renamed `_burnFromContract` to `_burnFromSelf` across modules, mocks, simulators, and tests.

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
0xisk: > Sooooo I'm wondering if we can remove evolveNonce here...

Resolved in `f27d85a` as related to your other comment. DerivedNonce is counter-only now and the seed is gone, so the unpredictability guidance no longer applies.

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
0xisk: Done in `f27d85a`. Dropped the seed and the hash chain; `_deriveNonce` is now `evolveNonce(_counter, tag)` in a single hash: unique via the counter, namespaced by the tag.

contracts/src/token/extensions/NativeShieldedTokenFamilySupply.compact
0xisk: Done in `f27d85a`. `NativeShieldedTokenSupplyCore` now documents that it does not catch mint-side drift (duplicate or wrong-domain `_addMinted`).

contracts/src/token/test/NativeShieldedToken.test.ts
0xisk: Done in `eccd544`. Removed the INV-* tags from the test names.

contracts/src/token/test/NativeShieldedToken.test.ts
0xisk: Done in `eccd544`. Supply assertions moved to the standalone `NativeShieldedTokenSupply` suite; the core token suite no longer touches supply.

contracts/src/token/test/NativeShieldedToken.test.ts
0xisk: Done in `eccd544`. One mock, simulator, and suite per module, mirrored under `test/extensions/` (Supply, FamilySupply, DerivedNonce each standalone).

contracts/src/token/test/NativeShieldedToken.test.ts
0xisk: Done in `eccd544`. `NativeShieldedToken.test.ts` is scoped to the core module now; supply and nonce moved out.

contracts/src/token/extensions/NativeShieldedTokenFamilySupply.compact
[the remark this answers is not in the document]
0xisk: Done in `f27d85a`: `_addBurned` now enforces `burned <= minted` in the shared `NativeShieldedTokenSupplyCore`, so a wrong-domain, duplicate, or unpaired burn reverts instead of pushing supply negative.

contracts/src/token/NativeShieldedToken.compact
```diff
@@ -0,0 +1,395 @@
...
+ *   and never exposed as a circuit parameter, eliminating caller-supplied
+ *   domain misuse. The coin's `color` is `tokenType(_domain, kernel.self())`,
+ *   computed at call time (never in the constructor). Only this contract can
+ *   ever mint coins of its color.
+ *
+ * - Metadata follows the FungibleToken (ERC-20) convention: contract-wide
+ *   `name`, `symbol`, and `decimals`, fixed at construction.
+ *
+ * - Supply totals are scalar ledger cells (cheaper circuits than the
+ *   token-family flavor's per-domain maps).
```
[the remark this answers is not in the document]
0xisk: Resolved in `f27d85a`: the stale scalar-ledger-cells note was removed when the module doc was reworked. The doc now states supply accounting is opt-in via the `NativeShieldedTokenSupply` extension.

contracts/src/token/extensions/NativeShieldedTokenSupply.compact
```diff
@@ -0,0 +1,134 @@
...
+   * @dev No overflow guard is needed: every coin of the paired contract's
+   * color originates from `_addMinted`-tracked mints, so burned can never
+   * exceed the overflow-checked minted total.
+   *
+   * @param {Uint<128>} amount - The burned amount.
+   * @return {[]} - Empty tuple.
+   */
+  export circuit _addBurned(amount: Uint<128>): [] {
+    _totalBurned = disclose((_totalBurned + amount) as Uint<128>);
+  }
```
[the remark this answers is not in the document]
0xisk: The supply extension moved to its own PR (#638, tracked by #640). `_addBurned` there enforces `burned <= minted` in the shared `NativeShieldedTokenSupplyCore`, so an unpaired or wrong-domain burn reverts.

contracts/test/integration/_harness/network.ts
```diff
@@ -0,0 +1,63 @@
...
+    walletNetworkId: 'undeployed' as NetworkId,
+    networkId: 'undeployed',
+    indexer:
+      process.env.MIDNIGHT_INDEXER_URL ??
+      'http://127.0.0.1:8088/api/v4/graphql',
+    indexerWS:
+      process.env.MIDNIGHT_INDEXER_WS_URL ??
+      'ws://127.0.0.1:8088/api/v4/graphql/ws',
+    node: process.env.MIDNIGHT_NODE_URL ?? 'http://127.0.0.1:9944',
+    nodeWS: 'ws://127.0.0.1:9944',
```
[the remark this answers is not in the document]
0xisk: The integration suite moved out of this PR into its own (tracked by #642); the `nodeWS` env-override fix will be handled there.

contracts/test/integration/specs/nativeShieldedToken/privacy.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  it('should NOT expose a secret-nonce mint recipient in public data (recipient-private)', async () => {
+    const secretNonce = new Uint8Array(randomBytes(32));
+    const res = await kit.deployed.callTx._mint(selfRecipient, 1_000n, secretNonce);
+
+    const blob = publicEffectsHexBlob(res.public);
+    // The secret nonce is the linchpin: it is not present in any public field,
+    // so the on-chain commitment cannot be recomputed by an enumerator.
+    expect(blob.includes(toHex(secretNonce))).toBe(false);
+  });
```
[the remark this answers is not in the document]
0xisk: The integration suite moved to its own PR (tracked by #642); the absence-only assertion will be tightened there (positive control plus case-normalized hex).

contracts/src/token/NativeShieldedToken.compact
```diff
@@ -0,0 +1,271 @@
...
+ *
+ * @notice Out of scope (phase two, pending contract-to-contract support):
+ * `balanceOf`, `allowance`, and transfer mediation are not representable for
+ * native shielded UTXOs — once a user holds a coin, the contract cannot observe
+ * or restrict its movement.
+ */
+module NativeShieldedToken {
+  import CompactStandardLibrary;
+  import "./NativeShieldedTokenCore" prefix Core_;
+
```
andrew-fleming, line 69: So this design runs into a problem with not being able to re-export the core state without having gross-looking keys (`NativeShieldedToken_Core__name`). This is an issue because if we remove the prefix from the imports (so we can export clean keys), the family variant will clash with `_mint` and `_burn`

We can solve this with a dual import approach using named imports for the ledger state vars we want to export. The idea being that circuits unambiguously use the prefix and state variables are named and re-exported as-is

```suggestion
  import "./NativeShieldedTokenCore" prefix Core_;
  import { _name, _symbol, _decimals, _isInitialized } from "./NativeShieldedTokenCore";
  export { _name, _symbol, _decimals, _isInitialized };

```

Let me know what you think or if you have a better design. If you agree, we can keep the state reads that were originally in the integration tests (it's a future PR but just being anticipatory and trying to solve problems before we have them)

andrew-fleming: If we go with the suggestion, I'd leave a comment saying something like "The import/re-export is only for state variables to namespace the keys appropriately in the implementing contract" or something along those lines

contracts/src/token/NativeShieldedTokenFamily.compact
```diff
@@ -0,0 +1,273 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.2.0 (token/NativeShieldedTokenFamily.compact)
+
+pragma language_version >= 0.23.0;
+
+/**
+ * @module NativeShieldedTokenFamily
+ * @description A module for issuing a FAMILY of native shielded (Zswap)
+ * token types from one contract, discriminated by a per-call domain.
+ *
```
andrew-fleming, line 10: I'm thinking we should explicitly document that contracts must not import both NST and NST family in the same contract because they share the same core state

contracts/src/token/NativeShieldedTokenCore.compact
andrew-fleming, line 1: **followup**: We should have a mock, sim, and test suite specifically targeting the bare base circuits

contracts/src/token/NativeShieldedTokenCore.compact
```diff
@@ -0,0 +1,336 @@
...
+  export circuit _mint(
+                   domain: Bytes<32>,
+                   recipient: Either<ZswapCoinPublicKey, ContractAddress>,
+                   amount: Uint<64>,
+                   nonce: Bytes<32>
+                   ): ShieldedCoinInfo {
+    assertInitialized();
+    assert(!Utils_isKeyOrAddressZero(recipient), "NativeShieldedToken: invalid recipient");
+
+    return mintShieldedToken(disclose(domain), disclose(amount), disclose(nonce), disclose(recipient));
```
andrew-fleming, line 243: Hmm even with `assertInitialized()`, users _could_ bypass this by calling `initialize` first and then minting or reading the token color after. In the docs, I think we should explicitly disallow doing any token operations in the constructor aside from `initialize`

andrew-fleming (commented):
The changes look great @0xisk! I just left a few comments

0xisk:
Great feedback @andrew-fleming! All fixed.

contracts/src/token/test/NativeShieldedTokenCore.test.ts
andrew-fleming, line 1: Should we include tests that check the getters return the correct ledger val?

contracts/src/token/NativeShieldedToken.compact
```diff
@@ -0,0 +1,286 @@
...
+ * `NativeShieldedTokenFamily` instead.
+ *
+ * This module is a thin wrapper over `NativeShieldedTokenCore`: it stores the
+ * one `sealed _domain` and forwards it to every core op, so the mint/burn logic
+ * and metadata live in the shared core, not here.
+ *
+ * @notice A contract MUST NOT import both `NativeShieldedToken` and
+ * `NativeShieldedTokenFamily`. Both wrap `NativeShieldedTokenCore`, so they
+ * share one set of core state (a single init flag and one name/symbol/decimals)
+ * and their `_mint`/`_burn` circuit names collide. Pick one flavor per contract.
```
andrew-fleming:
```suggestion
 * share one set of core state (a single init flag and one name/symbol/decimals).
 * Pick one flavor per contract.
```
This mixes up the justification. Mixing modules wouldn't cause collisions bc we have function overloading

contracts/src/token/NativeShieldedToken.compact
```diff
@@ -0,0 +1,286 @@
...
+ */
+module NativeShieldedToken {
+  import CompactStandardLibrary;
+  import "./NativeShieldedTokenCore" prefix Core_;
+
+  // Circuits use the `Core_` prefix above. The named import/re-export below is
+  // only for the core state variables, so the implementing contract's ledger
+  // keys read as `_name` etc. rather than `NativeShieldedTokenCore__name`. The
+  // circuit names (`_mint`, `_burn`, ...) are intentionally NOT imported
+  // unprefixed, to avoid clashing with this module's own circuits.
```
andrew-fleming:
```suggestion
  // keys read as `_name` etc. rather than `NativeShieldedTokenCore__name`. The
  // circuit names are intentionally NOT imported unprefixed, to avoid potential clashes.
```
The single variant wouldn't clash. "Potential" is more honest here

contracts/src/token/NativeShieldedTokenFamily.compact
```diff
@@ -0,0 +1,288 @@
...
+ * single-token issuer, use `NativeShieldedToken` instead.
+ *
+ * This module is a thin wrapper over `NativeShieldedTokenCore`: it forwards the
+ * caller's per-call `domain` to every core op, so the mint/burn logic and
+ * metadata live in the shared core, not here.
+ *
+ * @notice A contract MUST NOT import both `NativeShieldedTokenFamily` and
+ * `NativeShieldedToken`. Both wrap `NativeShieldedTokenCore`, so they share one
+ * set of core state (a single init flag and one name/symbol/decimals) and their
+ * `_mint`/`_burn` circuit names collide. Pick one flavor per contract.
```
andrew-fleming: If you agree with the doc suggestion in the single variant, I'd apply it here as well for consistency

andrew-fleming (commented):
Very nice, @0xisk ! A few minor (but important) suggestions. Afterwards, we should be good to go

contracts/src/token/test/NativeShieldedTokenCore.test.ts
andrew-fleming: *just ensuring the state and getters are in concert

contracts/src/token/NativeShieldedToken.compact
0xisk: Done in d4448a9 — dropped the collision claim; overloading means no clash here.

contracts/src/token/NativeShieldedToken.compact
0xisk: Done in 7e71af3 — softened to "potential" clashes.

contracts/src/token/NativeShieldedTokenFamily.compact
0xisk: Agreed — applied the same to the family variant in 969f36f for consistency.

contracts/src/token/test/NativeShieldedTokenCore.test.ts
0xisk: Added in 64e2b8d: a `simulator wiring` suite reads `getPublicState()` and asserts `name`/`symbol`/`decimals`/`isInitialized` return the ledger value verbatim, plus a distinct-metadata case pinning each getter to its own slot. Mock now re-exports the core ledger so it surfaces. CI green.

andrew-fleming (approved):
LGTM! Great work 🚀

PR #625, opened 2026-06-24
Add allowlist

### LINKED ISSUE #607: Add Allowlist
Per-account permit

andrew-fleming:
Resolves #607 

The original idea was to consolidate allowlist with blocklist (similar to [ERC20Restricted](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Restricted.sol)). The reason it was dismissed from this design was because we would not be able to cleanly support using both in a single contract due to compiler limitations. The other idea with consolidation would be to have two different sets: one for allows and one for blocks; however, this forces users who just want one to import ledger state that won't be used. Finally, keeping them as separate modules allows for it be explicit: There shouldn't be any confusion if a contract is using an allow list or a block list

As an aside, it is the author's opinion that these are `security` primitives; however, there is an argument that this belongs in `accesscontrol/`

## Release Notes

* **New Features**
  * Introduced allowlist functionality for account-based membership management and access control
  * Provides account validation capabilities and dynamic membership list management
  * Supports membership queries and enforcement mechanisms for access control

### DIFF
--- a/contracts/src/security/Allowlist.compact
+++ b/contracts/src/security/Allowlist.compact
@@ -31,8 +31,8 @@ module Allowlist {
   // State
   // ---------------------------------------------------------------------------
 
-  /** @description Approved accounts. Absent or `false` means not allowed. */
-  export ledger _allowed: Map<Bytes<32>, Boolean>;
+  /** @description Approved accounts. Absent means not allowed. */
+  export ledger _allowed: Set<Bytes<32>>;
 
   // ---------------------------------------------------------------------------
   // Views
@@ -41,14 +41,13 @@ module Allowlist {
   /**
    * @description Returns whether `account` is currently allowed.
    *
-   * @circuitInfo k=9, rows=341
+   * @circuitInfo k=9, rows=305
    *
    * @param {Bytes<32>} account - The account to query.
    * @return {Boolean} - True if `account` is a member of the allowlist.
    */
   export circuit isAllowed(account: Bytes<32>): Boolean {
-    return _allowed.member(disclose(account)) &&
-           _allowed.lookup(disclose(account));
+    return _allowed.member(disclose(account));
   }
 
   // ---------------------------------------------------------------------------
@@ -59,7 +58,7 @@ module Allowlist {
    * @description Asserts that `account` is allowed. The composing contract
    * decides when this gate applies.
    *
-   * @circuitInfo k=9, rows=341
+   * @circuitInfo k=9, rows=305
    *
    * Requirements:
    *
@@ -85,7 +84,7 @@ module Allowlist {
    * @return {[]} - Empty tuple.
    */
   export circuit _allow(account: Bytes<32>): [] {
-    _allowed.insert(disclose(account), true);
+    _allowed.insert(disclose(account));
   }
 
   /**
@@ -97,6 +96,6 @@ module Allowlist {
    * @return {[]} - Empty tuple.
    */
   export circuit _disallow(account: Bytes<32>): [] {
-    _allowed.insert(disclose(account), false);
+    _allowed.remove(disclose(account));
   }
 }
--- a/contracts/src/security/Allowlist.compact
+++ b/contracts/src/security/Allowlist.compact
@@ -1,0 +1,108 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.2.0 (security/Allowlist.compact)
+
+pragma language_version >= 0.23.0;
+
+/**
+ * @module Allowlist
+ * @description A minimal, reusable permit-list keyed by an abstract `Bytes<32>`
+ * account identifier. It answers a single question: is this account a member?
+ *
+ * This module is a pure membership MECHANISM, not a policy. It holds no notion
+ * of "when" the list is enforced and no access control. The mutators (`_allow`,
+ * `_disallow`) are deliberately UNGATED composable primitives: a composing
+ * contract gates them behind its access-control module (the same way a token
+ * gates `_mint` / `_burn`), and decides when to call `assertAllowed` (for
+ * example, only while an "eligibility required" flag is set). Keeping the
+ * enforce-or-not decision in the composing layer leaves this module general
+ * enough to back KYC allowlists, permissioned mints, membership gates, and the
+ * like.
+ *
+ * @dev Eligibility seam. `assertAllowed` is the single check a composing
+ * contract calls. This is the swap point: a confidential / graph-private asset
+ * replaces the public-map backend here with a zero-knowledge membership proof
+ * against a published root, WITHOUT changing the composition. See the
+ * confidential graph-privacy findings doc.
+ *
+ * @notice This is the PUBLIC (transparent) allowlist: membership status is
+ * publicly readable and the approved set is publicly enumerable. That is the
+ * right posture for transparent or public-graph assets. A graph-private asset
+ * should instead use a shielded allowlist (a public allowlist would re-expose
+ * the candidate set and bound the anonymity set).
+ */
+module Allowlist {
+  import CompactStandardLibrary;
+
+  // ---------------------------------------------------------------------------
+  // State
+  // ---------------------------------------------------------------------------
+
+  /** @description Approved accounts. Absent or `false` means not allowed. */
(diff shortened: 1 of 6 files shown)


contracts/src/security/Allowlist.compact
```diff
@@ -0,0 +1,101 @@
...
+ * enough to back KYC allowlists, permissioned mints, membership gates, and the
+ * like.
+ *
+ * @notice This is the PUBLIC (transparent) allowlist: membership status is
+ * publicly readable and the approved set is publicly enumerable. That is the
+ * right posture for transparent or public-graph assets. A graph-private asset
+ * should instead use a shielded allowlist (a public allowlist would re-expose
+ * the candidate set and bound the anonymity set).
+ */
+module Allowlist {
```
0xisk, line 27: ❔ **question** (non-blocking): Is `security/` the final home, or should this and `Blocklist` move to `accesscontrol/`? Your own PR description flags it ("there is an argument that this belongs in `accesscontrol/`"). Worth deciding once for both modules so they stay together.

_added by claude (dev3-midnight-basic-review skill)_

contracts/src/security/Allowlist.compact
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * This module is a pure membership MECHANISM, not a policy. It holds no notion
+ * of "when" the list is enforced and no access control. The mutators (`_allow`,
+ * `_disallow`) are deliberately UNGATED composable primitives: a composing
+ * contract gates them behind its access-control module (the same way a token
+ * gates `_mint` / `_burn`), and decides when to call `assertAllowed` (for
+ * example, only while an "eligibility required" flag is set). Keeping the
+ * enforce-or-not decision in the composing layer leaves this module general
+ * enough to back KYC allowlists, permissioned mints, membership gates, and the
+ * like.
```
0xisk, line 19: 🔵 **followup:** State the fail-closed lockout consequence explicitly. An empty allowlist rejects everyone, so a contract that enforces `assertAllowed` before the set is populated bricks access for all accounts (the operator included). The doc already mitigates this well with the eligibility-flag example, so this is one sentence of polish: note that the operator must populate and/or gate enforcement first. (Fails safe, so lower stakes than the inverse note on Blocklist.)

_added by claude (dev3-midnight-basic-review skill)_

contracts/src/security/Allowlist.compact
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * @description Returns whether `account` is currently allowed.
+   *
+   * @circuitInfo k=9, rows=305
+   *
+   * @param {Bytes<32>} account - The account to query.
+   * @return {Boolean} - True if `account` is a member of the allowlist.
+   */
+  export circuit isAllowed(account: Bytes<32>): Boolean {
+    return _allowed.member(disclose(account));
```
0xisk, line 50: 🔵 **followup:** Add a one-line `@notice`/`@dev` on `isAllowed` (inherited by `assertAllowed`) that calling it discloses `account` to the public transcript. The module `@notice` covers the posture, but a dev reading only the circuit's API-ref entry sees `@param account` with no disclosure hint, and could leak an identifier when composing this into an otherwise-private contract.

_added by claude (dev3-midnight-basic-review skill)_

contracts/src/security/test/Allowlist.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      expect(
+        (await sim.getPublicState()).Allowlist__allowed.member(ALICE),
+      ).toBe(false);
+
+      await sim.allow(ALICE);
+
+      expect(
+        (await sim.getPublicState()).Allowlist__allowed.member(ALICE),
+      ).toBe(true);
+    });
```
0xisk, line 125: 🔵 **followup (test):** This wiring test asserts an `allow` shows up in the raw `Allowlist__allowed` ledger, but never asserts a `disallow` removes it. Close the asymmetry so the state invariant is checked against the resulting ledger output (the field indexers subscribe to), not only via the `isAllowed` circuit:

```ts
it('should reflect disallow in the public ledger', async () => {
  const sim = await AllowlistSimulator.create();
  await sim.allow(ALICE);
  expect((await sim.getPublicState()).Allowlist__allowed.member(ALICE)).toBe(true);
  await sim.disallow(ALICE);
  expect((await sim.getPublicState()).Allowlist__allowed.member(ALICE)).toBe(false);
});
```

_added by claude (dev3-midnight-basic-review skill)_

contracts/src/security/test/Allowlist.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+// Deterministic 32-byte account identifier seeded from a label.
+const account = (label: string): Uint8Array => {
+  const a = new Uint8Array(32);
+  a.set(new TextEncoder().encode(label).slice(0, 32));
+  return a;
+};
+
+const ALICE = account('ALICE');
+const BOB = account('BOB');
```
0xisk, line 12: 🔵 **followup (test):** No test pins the all-zero `Bytes<32>` account. OZ Solidity special-cases `address(0)`; this module deliberately does not, so a test guards against a future regression that adds an implicit exemption:

```ts
const ZERO = new Uint8Array(32);
it('should treat the all-zero account as an ordinary member', async () => {
  expect(await allowlist.isAllowed(ZERO)).toBe(false);
  await allowlist.allow(ZERO);
  expect(await allowlist.isAllowed(ZERO)).toBe(true);
  await allowlist.assertAllowed(ZERO);
  await allowlist.disallow(ZERO);
  await expect(allowlist.assertAllowed(ZERO)).rejects.toThrow('Allowlist: account not allowed');
});
```

_added by claude (dev3-midnight-basic-review skill)_

contracts/src/security/test/Allowlist.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    it('clears with a single disallow after being allowed multiple times', async () => {
+      await allowlist.allow(ALICE);
+      await allowlist.allow(ALICE);
+      expect(await allowlist.isAllowed(ALICE)).toBe(true);
+      await allowlist.disallow(ALICE);
+      // Membership is binary, not a counter: one disallow clears it regardless
+      // of how many times it was allowed.
+      expect(await allowlist.isAllowed(ALICE)).toBe(false);
+    });
```
0xisk, line 64: 🔵 **followup (test):** The allow→allow→disallow case is asserted only through `isAllowed`. Exercise the check seam (`assertAllowed` throwing) on the same path so both seams agree:

```ts
it('should throw from assertAllowed after a single disallow of a multiply-allowed account', async () => {
  await allowlist.allow(ALICE);
  await allowlist.allow(ALICE);
  await allowlist.disallow(ALICE);
  await expect(allowlist.assertAllowed(ALICE)).rejects.toThrow('Allowlist: account not allowed');
});
```

_added by claude (dev3-midnight-basic-review skill)_

contracts/src/security/test/Allowlist.test.ts
```diff
@@ -0,0 +1,127 @@
...
+describe('Allowlist', () => {
+  beforeEach(async () => {
+    allowlist = await AllowlistSimulator.create();
+  });
+
+  describe('default state', () => {
+    it('is empty: no account is allowed', async () => {
+      expect(await allowlist.isAllowed(ALICE)).toBe(false);
+      expect(await allowlist.isAllowed(BOB)).toBe(false);
+    });
```
0xisk, line 25: 🔵 **followup (test):** A view shouldn't mutate. Cheap assertion that `isAllowed` leaves the set untouched:

```ts
it('should not mutate the set when querying isAllowed', async () => {
  const sim = await AllowlistSimulator.create();
  await sim.isAllowed(ALICE);
  expect((await sim.getPublicState()).Allowlist__allowed.member(ALICE)).toBe(false);
});
```

_added by claude (dev3-midnight-basic-review skill)_

0xisk:
Non-blocking follow-ups and one open question from the basic review are collected in #635 (7 items: 1 hardening, 1 docs, 4 test coverage, 1 placement question). None block this PR.

_added by claude (dev3-midnight-basic-review skill)_

0xisk (approved):
LGTM! thank you @andrew-fleming!

0xisk (approved):

PR #638, opened 2026-06-29
feat(token): native shielded token supply extension

### LINKED ISSUE #640: Add Native Shielded Token supply extension
Sub-issue of #544.

**🧐 Motivation**
On-chain supply accounting (`totalSupply`) is a privacy trade-off, not core to the native shielded token: composing it makes contract-mediated burn amounts public. So it ships as an optional extension that consumers opt into only when they need auditable supply.

**📝 Details**
- `NativeShieldedTokenSupply` (scalar) and `NativeShieldedTokenFamilySupply` (per-domain) as thin wrappers over a shared `NativeShieldedTokenSupplyCore`.
- `_addMinted` / `_addBurned` building blocks the consumer pairs with the matching mint/burn op; `totalMinted` / `totalBurned` / `totalSupply` getters.
- `_addBurned` enforces the per-domain `burned <= minted` invariant.

Implemented by #638 (stacked on #621).

0xisk:
## Types of changes

Stacked on #621 (core native shielded token). **Base this PR's review on the diff against `feat/native-shielded-token-supply`, not `main`.**

Optional on-chain supply accounting for the native shielded token:

- `NativeShieldedTokenSupply` (scalar) and `NativeShieldedTokenFamily Supply` (per-domain) are thin wrappers over a shared `NativeShieldedTokenSupplyCore`.
- `_addMinted` / `_addBurned` building blocks the consumer pairs with the matching mint/burn op; `totalMinted` / `totalBurned` / `totalSupply` getters.
- `_addBurned` enforces the per-domain `burned <= minted` invariant: a readable error that also makes the add-overflow and the `totalSupply` underflow unreachable.
- `UINT128_MAX` comes from `Utils`.
- Per-module mocks, simulators, and unit suites (extensions import no token module, so they are driven standalone).

> Privacy note: composing supply accounting makes contract-mediated burn amounts public (the `_addBurned` write is the disclosure). The bare core burns stay amount-private.

## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added supply-tracking support for native shielded tokens, including total minted, total burned, and total supply views.
  * Added per-domain supply tracking for token families, keeping balances isolated by domain.
  * Added test simulators and mock contracts to exercise the new supply features.

* **Bug Fixes**
  * Enforced that burned amounts cannot exceed minted amounts.
  * Added coverage for zero-value and multi-domain supply behavior.


contracts/src/token/extensions/NativeShieldedTokenSupply.compact
```diff
@@ -0,0 +1,129 @@
...
+ * burn-amount privacy.
+ *
+ * @dev Mint amounts are already public via the protocol's `shieldedMints`
+ * effect, so `_addMinted` adds accounting, not disclosure. See
+ * `NativeShieldedTokenSupplyCore` for the exact / lower-bound / upper-bound
+ * guarantees, the `burned <= minted` invariant, and the mint-side drift caveat.
+ */
+module NativeShieldedTokenSupply {
+  import CompactStandardLibrary;
+  import "./NativeShieldedTokenSupplyCore" prefix Core_;
```
andrew-fleming, line 61: Should we apply the named import/re-export pattern to expose the core supply ledger keys?

andrew-fleming: If yes, same with the supply family module + tests

contracts/src/token/extensions/NativeShieldedTokenSupplyCore.compact
```diff
@@ -0,0 +1,142 @@
...
+   * burns that bypass the contract are invisible. See the module notes.
+   *
+   * @circuitInfo k=9, rows=440
+   *
+   * @param {Bytes<32>} domain - 32-byte token discriminator.
+   * @return {Uint<128>} - The upper bound on tokens in existence.
+   */
+  export circuit totalSupply(domain: Bytes<32>): Uint<128> {
+    return (totalMinted(domain) - totalBurned(domain)) as Uint<128>;
+  }
```
andrew-fleming, line 162: Hypothetically, if totalBurned > totalMinted, would it be better to return 0 as opposed to returning an error? In other words, should we check that totalMinted > totalBurned and return 0 otherwise?

andrew-fleming: If you agree, we might have to get a bit creative with how to test this though...maybe have an `unsafeCorruptState` circuit or something in the mock

contracts/src/token/test/extensions/NativeShieldedTokenSupply.property.test.ts
andrew-fleming, line 1: Very nice to include this

andrew-fleming: **followup** I'd target the family variant as well...instead of having separate files for each, since they're quite small, maybe we can get away with just having one file that fuzzes all three modules (including core)

andrew-fleming (commented):
Looking good, @0xisk! You mind fixing conflicts? Also, I didn't review the integration files. I assume these are meant to be removed for now since we can't run the integration tests yet

contracts/src/token/extensions/NativeShieldedTokenSupply.compact
0xisk: Done in `ec1804e`. Re-exported `_totalMinted` / `_totalBurned` from `SupplyCore` in both `NativeShieldedTokenSupply` and `NativeShieldedTokenFamilySupply` (same pattern as the core token modules), surfaced them in both mocks, and added getters-match-ledger wiring tests to each suite.

contracts/src/token/extensions/NativeShieldedTokenSupplyCore.compact
0xisk: Agreed, done in `5770f4a`. `totalSupply` now returns `minted >= burned ? minted - burned : 0`, so the getter reads cleanly instead of reverting on an out-of-bound state. Tested via a test-only `unsafeSetBurned` on the mocks that writes `_totalBurned` directly (bypassing the invariant) to reach `burned > minted`. The guarded subtraction is also 1 row cheaper.

contracts/src/token/test/extensions/NativeShieldedTokenSupply.property.test.ts
0xisk: Done in `5f47727`. One property file now fuzzes both the scalar and per-domain flavors; both delegate to `SupplyCore`, so core is covered transitively (no standalone core mock).

contracts/src/token/extensions/NativeShieldedTokenSupplyCore.compact
```diff
@@ -0,0 +1,149 @@
...
+   *
+   * @param {Bytes<32>} domain - 32-byte token discriminator.
+   * @param {Uint<128>} amount - The burned amount.
+   * @return {[]} - Empty tuple.
+   */
+  export circuit _addBurned(domain: Bytes<32>, amount: Uint<128>): [] {
+    const minted = totalMinted(domain);
+    const current = totalBurned(domain);
+    assert(minted >= current, "NativeShieldedTokenSupply: burned exceeds minted");
+    assert(minted - current >= amount, "NativeShieldedTokenSupply: burned exceeds minted");
```
andrew-fleming: Do we need both checks? If so, should the error msgs reflect it? Right now, it's difficult to confirm which is triggered because the msgs are the same

andrew-fleming: This line in particular is interesting. This works assuming the minted side is correct. A potential concern is if the mint side undercounts, users may not be able to legitimately burn

There's an argument for creating a token + supply module so that there's a guarantee that accounting won't drift and we can safely assume the relationship is sound. This will be a reoccurring question with compositional design and I think both sides have a point. Flexibility + required user correctness || Rigid with more lib maintenance + harder for users to misuse

andrew-fleming: If we keep this design as-is, let's document that failure to pair these circuits correctly with mint and burn can have serious consequences

andrew-fleming (commented):
Improvements look good, @0xisk!

contracts/src/token/extensions/NativeShieldedTokenSupplyCore.compact
0xisk: Yes, both are needed and now carry distinct messages (e00ef37).

- `minted >= current` → `"burned total exceeds minted"`: guards the subtraction below from underflowing if stored `burned` is somehow already past `minted`. Unreachable through the accounting API itself, so it's the corrupt-state guard.
- `minted - current >= amount` → `"burn exceeds available supply"`: the reachable per-call check that this burn doesn't push the domain past its minted total.

Added a comment on the split and a corrupt-state test in each suite (`unsafeSetBurned` above minted, then `_addBurned`) so the first message is actually asserted — the existing five reverting cases all hit the second.

0xisk: Kept the compositional design and documented the risk as a module-level `@warning` in all three supply files (2dcd26e). Core carries the authoritative version; the two flavors point back to it.

It calls the mint/burn ↔ accounting pairing a security-critical invariant the extension can't enforce or detect, and spells out the case you flagged: a mint whose `_addMinted` is missing/wrong-domain leaves `totalMinted` too low, so `_addBurned` then rejects burns that are actually legitimate — holders can't burn coins that genuinely exist. Also covers the overcount (over-burn allowed) and unpaired-burn (`totalSupply` overstated) directions.

andrew-fleming (approved):
LGTM!

PR #639, opened 2026-06-29
feat(token): native shielded token derived-nonce extension

### LINKED ISSUE #641: Add Native Shielded Token derived-nonce extension
Sub-issue of #544.

**🧐 Motivation**
Some mint flows should not require the caller to manage coin nonces. An optional derived-nonce extension supplies deterministic, recipient-public nonces for those flows; callers needing recipient privacy keep using the base `_mint` with a secret random nonce.

**📝 Details**
- Counter-only: a monotonic `_counter`; `_deriveNonce` returns `evolveNonce(_counter, tag)` in a single hash. Unique via the counter, namespaced by the fixed tag. No seed, no init.
- Deterministic and recipient-public by design.
- Standalone (imports no token module).

Implemented by #639 (stacked on #621).

0xisk:
## Types of changes

Stacked on #621 (core native shielded token). **Base this PR's review on the diff against `feat/native-shielded-token-supply`, not `main`.**

Optional derived coin nonces for native shielded token mints, so callers need not manage nonces themselves:

- Counter-only: owns a monotonic `_counter` and derives `evolveNonce(_counter, tag)` in a single hash. Uniqueness comes from the counter; the fixed tag namespaces derived nonces away from raw caller-supplied ones. No seed, no init.
- Deterministic and recipient-public by design. For recipient privacy, call the base `_mint` with a secret, random nonce instead.
- Standalone mock, simulator, and unit suite (imports no token module).

## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added deterministic nonce derivation for shielded token flows, with a counter-based 32-byte nonce that advances on each use.
  * Included simulator support so the new behavior can be exercised in test environments.

* **Tests**
  * Added coverage for nonce length, counter progression, uniqueness across repeated derivations, and consistent results across separate runs.

### DIFF
--- a/contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
+++ b/contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
@@ -40,7 +40,9 @@ pragma language_version >= 0.23.0;
  * - Access-control both mint paths, and avoid exposing them at different trust
  *   levels.
  * - Do not rely on retrying. The nonce is a pure function of the counter, and a
- *   reverted mint may not advance it, so the retry can collide again.
+ *   reverted mint will not advance it, so an identical retry collides again.
+ *   Recovery is a mint with different inputs (recipient or amount): it advances
+ *   the counter, so the original tuple then mints against a fresh nonce.
  */
 module NativeShieldedTokenDerivedNonce {
   import CompactStandardLibrary;
--- a/contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
+++ b/contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-// OpenZeppelin Compact Contracts v0.2.0 (token/extensions/NativeShieldedTokenDerivedNonce.compact)
+// OpenZeppelin Compact Contracts v0.3.0-alpha (token/extensions/NativeShieldedTokenDerivedNonce.compact)
 
 pragma language_version >= 0.23.0;
 
@@ -10,20 +10,21 @@ pragma language_version >= 0.23.0;
  * owns a monotonic `_counter` and exposes `_deriveNonce`; it imports no token
  * module and needs no initialization.
  *
- * @notice Pairs with `NativeShieldedToken`, `NativeShieldedTokenFamily`, and
- * `NativeTokenConverter`. The consuming contract composes the pieces:
+ * @notice Pairs with `NativeShieldedToken` and `NativeShieldedTokenFamily`. The
+ * consuming contract composes the pieces:
  *
  *   export circuit mint(recipient: ..., amount: Uint<64>): ShieldedCoinInfo {
  *     return Token__mint(recipient, amount, Derived__deriveNonce());
  *   }
  *
  * @dev Derivation
- * Each call derives `evolveNonce(_counter, pad(32, "NativeShieldedToken:nonce"))`:
- * the monotonic counter guarantees uniqueness, and the fixed tag pushes derived
- * nonces out of the small-integer space an honest caller of the base `_mint`
- * might pick, avoiding accidental cross-path collisions. One counter serves the
- * whole contract, so derived nonces never repeat across token types (domains)
- * or across the shielded/converter mint paths of one contract.
+ * Each call derives `evolveNonce(_counter, pad(32, "NativeShieldedToken:nonce"))`.
+ * `evolveNonce` hashes the counter into a full 32-byte nonce, so uniqueness comes
+ * from the monotonic counter, not the tag. The fixed tag is a domain separator:
+ * it keeps these derived nonces from clashing with any other `evolveNonce` call
+ * the implementing contract makes on the same counter for a different purpose.
+ * One counter serves the whole contract, so derived nonces never repeat across
+ * token types (domains) of one contract.
  *
  * @notice This variant is deterministic and recipient-PUBLIC by design: derived
  * nonces are predictable from public state, so anyone can recompute the coin
@@ -38,8 +39,9 @@ pragma language_version >= 0.23.0;
  * collides with a future derived-nonce mint of the same tuple, making that mint
  * fail on duplicate-commitment rejection. Gate both mint paths behind access
  * control, and prefer not exposing both to distinct trust levels. A failed mint
- * is recoverable: any subsequent derived mint advances the counter past the
- * collision.
+ * is recoverable with the same recipient and amount: the counter has already
+ * advanced, so the retry derives a fresh nonce and therefore a different
+ * commitment, which no longer collides.
  */
 module NativeShieldedTokenDerivedNonce {
   import CompactStandardLibrary;
@@ -51,8 +53,7 @@ module NativeShieldedTokenDerivedNonce {
 
   /**
    * @description Advances the counter and returns the next derived coin nonce,
-   * for use as the `nonce` argument of a native token mint or a converter
-   * shield.
+   * for use as the `nonce` argument of a native token mint.
    *
    * @notice The derivation input is public ledger state, so mints using this
    * nonce are recipient-public; see the module notes.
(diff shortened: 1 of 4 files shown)


contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * @module NativeShieldedTokenDerivedNonce
+ * @description Optional standalone extension providing derived coin nonces for
+ * native shielded token mints, so callers need not manage nonces themselves. It
+ * owns a monotonic `_counter` and exposes `_deriveNonce`; it imports no token
+ * module and needs no initialization.
+ *
+ * @notice Pairs with `NativeShieldedToken`, `NativeShieldedTokenFamily`, and
+ * `NativeTokenConverter`. The consuming contract composes the pieces:
```
andrew-fleming: **question** Is there an existing `NativeTokenConverter`? Or is this an anticipated feature?

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
```diff
@@ -0,0 +1,68 @@
...
+ * nothing here because the seed and chain are disclosed to public state, so it
+ * was removed.)
+ *
+ * @notice The predictability also enables deliberate collision-griefing: an
+ * actor with access to a caller-nonce mint can pre-mint a commitment that
+ * collides with a future derived-nonce mint of the same tuple, making that mint
+ * fail on duplicate-commitment rejection. Gate both mint paths behind access
+ * control, and prefer not exposing both to distinct trust levels. A failed mint
+ * is recoverable: any subsequent derived mint advances the counter past the
+ * collision.
```
andrew-fleming: **followup** Nit-ish: Wouldn't a subsequent mint with identical values from the failed mint continue to fail? At least one of the values would need to be different, no?

contracts/src/token/test/extensions/NativeShieldedTokenDerivedNonce.test.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  it('should never repeat a derived nonce across N calls', async () => {
+    const N = 25;
+    const seen = new Set<string>();
+    for (let i = 0; i < N; i++) {
+      seen.add(toHex(await nonce._deriveNonce()));
+    }
+    expect(seen.size).toBe(N);
+  });
```
andrew-fleming, line 38: ::followup:: It might be a good idea to add a deterministic test so there's a loud error if some sort of randomness were introduced

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
```diff
@@ -0,0 +1,68 @@
...
+ * `NativeTokenConverter`. The consuming contract composes the pieces:
+ *
+ *   export circuit mint(recipient: ..., amount: Uint<64>): ShieldedCoinInfo {
+ *     return Token__mint(recipient, amount, Derived__deriveNonce());
+ *   }
+ *
+ * @dev Derivation
+ * Each call derives `evolveNonce(_counter, pad(32, "NativeShieldedToken:nonce"))`:
+ * the monotonic counter guarantees uniqueness, and the fixed tag pushes derived
+ * nonces out of the small-integer space an honest caller of the base `_mint`
```
andrew-fleming: **followup** nit: isn't it the evolveNonce circuit that expands the width? AFAICT the domain prevents clashes if the implementing contract uses `evolveNonce` elsewhere

andrew-fleming (commented):
Looking good, @0xisk! Would you mind fixing conflicts?

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
0xisk: That was part of the exploreing contracts here: [https://github.com/OpenZeppelin/compact-contracts/blob/feat/expr-native-shielded-token/contracts/src/token/extensions/NativeTokenConverter.compact](https://github.com/OpenZeppelin/compact-contracts/blob/feat/expr-native-shielded-token/contracts/src/token/extensions/NativeTokenConverter.compact) But I have not included it into the NST design as it is out of scope. So that is doc bug will fix it.

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
0xisk: Correct, reworded in a7a1434: `evolveNonce` does the width expansion (hashes the counter into a full 32-byte nonce), so uniqueness comes from the monotonic counter, not the tag. The fixed tag is a domain separator, keeping these nonces from clashing with any other `evolveNonce` call the implementing contract makes on the same counter.

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
0xisk: ~~The retry succeeds precisely because the counter already advanced: the derived nonce is one of the commitment inputs, so a fresh nonce yields a different commitment even with identical recipient and amount. No value needs to change by hand. Clarified the wording in a7a1434.~~

contracts/src/token/test/extensions/NativeShieldedTokenDerivedNonce.test.ts
0xisk: Added in f1f9e02: two independent deployments must derive byte-identical sequences (the derivation is a pure function of the public counter), so any randomness introduced into `_deriveNonce` makes them diverge and fails loudly.

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
andrew-fleming: > The retry succeeds precisely because the counter already advanced

When does it advance? The counter advances within the same tx, no? Therefore, if mint with `counter = 1` fails, the counter doesn't advance and the subsequent mint would still have `counter = 1`. What am I missing?

0xisk: You're right, and sorry for the confusion on my end. The failed mint reverts the counter bump, so a same-tuple retry re-derives the same nonce and collides again. I've corrected the note to drop the retry-recovery claim and lean on access control as the mitigation. Thanks for catching it.

0xisk: You're right, and sorry for the confusion on my end. Ive corrected the note to drop the retry-recovery claim and lean on access control as the mitigation. Thanks for catching it.

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
```diff
@@ -34,14 +34,13 @@ pragma language_version >= 0.23.0;
...
- * is recoverable with the same recipient and amount: the counter has already
- * advanced, so the retry derives a fresh nonce and therefore a different
- * commitment, which no longer collides.
+ * @notice Collision-griefing. Derived nonces are predictable, so an actor with a
+ * caller-nonce mint can pre-mint a colliding commitment and make the matching
+ * derived-nonce mint fail. Mitigations:
+ * - Access-control both mint paths, and avoid exposing them at different trust
+ *   levels.
+ * - Do not rely on retrying. The nonce is a pure function of the counter, and a
+ *   reverted mint may not advance it, so the retry can collide again.
```
andrew-fleming: **followup** the docs are fine as-is tbh. Just some tiny suggestions that might be helpful

>  ...a reverted mint may not advance it...

The only time the counter could advance is if it's decoupled from _mint. The whole idea of this module is to remove nonce management responsibility. IMO a definitive statement is better: "a reverted mint will not advance it..."

> Do not rely on retrying.

This is fine; however, it might be useful to offer guidance if this case does occur. Maybe something like "Recovery is a mint with different inputs." The original failed mint would then work bc the nonce advanced

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
andrew-fleming: n/p!

andrew-fleming (approved):
@0xisk LGTM!

contracts/src/token/extensions/NativeShieldedTokenDerivedNonce.compact
0xisk: good suggestions! will add those fixes.

andrew-fleming (approved):
LGTM!

PR #653, opened 2026-07-02
Add Confidential Fungible Token

### LINKED ISSUE #569: Add Confidential Contract Tokens
**🧐 Motivation**

**📝 Details**

andrew-fleming:
Resolves #569

## Summary by CodeRabbit

* **New Features**
  * Added confidential fungible token support with encrypted balances, private transfers, approvals, escrow, memos, and account registration.
  * Added optional public supply management, including minting, burning, and total supply tracking.
  * Added enhanced confidential decryption verification capabilities.
* **Tests**
  * Added comprehensive coverage for transfers, balances, allowances, supply operations, memo handling, registration, and security validations.
* **Documentation**
  * Documented the confidential fungible token feature in the unreleased changelog.


0xisk (commented):
## Review (midnight-basic-review, dev3)

**Scope:** all 12 files (+4,275) — Core + PublicSupply modules, the two new ElGamal scalar variants, witnesses, mocks, simulators, 3 test suites. Checked against the midnight-basic-review checklist (privacy/disclosure, circuit soundness, authorization/replay, state integrity, economic, composability).

**Verified locally:** SKIP_ZK compile + all 80 PR tests pass (mock backend). A 3-test simulator repro was written for the escrow finding; see the inline comment on `_spendEscrow`.

🔴 **blocking:** CI never ran this PR's tests: "Run Test Suite" hit `timeout-minutes: 30` during `yarn compact` (step cancelled, type checks and tests skipped). CI compiles without `SKIP_ZK`, so it runs full key generation, and the two CFT mocks add ~6 k=15/16 circuits (`transfer` ~44k rows, `transferFrom` ~52k rows, in both mocks). Unblocks: a green run — bump the timeout, cache compiled artifacts, or compile with `SKIP_ZK` in the test job. FWIW the suite is green locally (80/80 in ~7s with SKIP_ZK).

**Checklist summary:**

| Category | Result |
|---|---|
| Privacy & disclosure | Pass, conditional on wallet seed freshness (documented); counterparty graph public by design |
| Circuit soundness | Pass (bounded Uint<128> amounts, valid-scalar derivations, codegen workaround documented) |
| Authorization & replay | Balance paths sound (SK+EK binding); escrow control gap — see blocking inline comment |
| State integrity | Dual-balance/sweep sound; same-block conflicts documented |
| Economic | k=16 circuits near row ceiling; memo-spam grief documented as deferred |
| Composability | Caller-id return pattern enables wrapper gating; `_spendEscrow` re-export hazard — see inline followup |
| Upgradeability / CMA | N/A (library modules; deploy decisions live with consumers) |

**Verdict: needs changes — 3 blocking (1 here, 2 inline), 2 questions.**

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -0,0 +1,1027 @@
...
+   *
+   * @notice Only the spender's escrow copy is decrypted and checked
+   * `>= value`. The owner's copy is reduced by the same `value`, preserving the
+   * approve-time invariant that both copies encrypt the same amount, so the
+   * owner's copy needs no separate bound check.
+   *
+   * @param fromAddress - The owner whose escrow the caller (spender) draws on.
+   * @param value - The amount to consume from the allowance.
+   */
+  export circuit _spendEscrow(fromAddress: Bytes<32>, value: Uint<128>): Bytes<32> {
```
0xisk: 🔴 **blocking:** After a partial `transferFrom`/`burnFrom`, the owner can never re-approve or revoke this spender, and the escrow remainder is stranded. `_spendEscrow` reduces the owner's escrow copy by a value delivered only to the spender and the recipient (no owner memo). The next `approve` homomorphically refunds that unknown remainder and then requires the owner to prove the exact post-refund plaintext, which the owner cannot know. The header's claim that the owner "observes it by initiating that transaction" only holds when nothing was spent in between. A spender can also grief deliberately: spend 1, and the owner permanently loses control of the remaining allowance.

Verified with a simulator repro: re-approve passes only when the wallet is given the spent amount out of band; with the owner's plausible belief (escrow untouched), `approve(bob, 10)` and `approve(bob, 0)` both fail with `ElGamal: plaintext mismatch`.

Unblocks: push an EcdhMask memo to the owner in `_spendEscrow` carrying the spent value (mirror of `_credit`'s recipient memo), or document the spender's off-chain disclosure obligation as a hard protocol requirement and add the partial-spend re-approve test.

contracts/src/crypto/ElGamal.compact
```diff
@@ -470,4 +470,59 @@ module ElGamal {
...
+   *
+   * - `ekScalar` is the discrete log of `pk` (`g^ekScalar == pk`).
+   * - `ct` decrypts under `(pk, ekScalar)` to `m`.
+   *
+   * @param ct - The ciphertext to verify.
+   * @param pk - The public key `ct` is encrypted under (`g^ekScalar`).
+   * @param ekScalar - The secret scalar.
+   * @param m - The message point `ct` is claimed to encrypt.
+   */
+  export pure circuit assertDecryptsToPointScalar(
```
0xisk: 🔴 **blocking:** `assertDecryptsToPointScalar` / `assertDecryptsToScalar` are referenced nowhere outside their own definitions and have no test coverage (`ElGamal.test.ts` does not exercise them). They also lack the weak-input hygiene of their siblings: `ekScalar = 0` with `pk = identity` passes vacuously (claims `m = c2` for any ciphertext). The hashed-ek path cannot reach scalar 0, but the computed stealth scalars these are built for can. Unblocks: move these to the stealth branch that needs them, or add mirror tests of the `assertDecryptsTo` suite plus an `ekScalar != 0` (or `pk != identity`) assert.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -0,0 +1,1027 @@
...
+   * @param {Bytes<32>} to - The recipient account.
+   * @param {Uint<128>} value - The amount to transfer.
+   * @return {Bytes<32>} - The caller's (sender's) accountId.
+   */
+  export circuit transfer(to: Bytes<32>, value: Uint<128>): Bytes<32> {
+    assertInitialized();
+    // Derive the sender from the same authentication `_debit` uses, so the
+    // returned id is exactly the account whose funds moved.
+    const senderId = _debit(value);
+    assert(senderId != to, "ConfidentialFungibleToken: self-transfer");
```
0xisk, line 739: ❔ **question:** `transfer` rejects self-transfer but `_move` allows it, and the flow itself is self-transfer-safe (debit spendable, credit own pending). What does this guard protect? If it is purely a semantic choice, a sentence in `_move`'s doc would keep composers from being surprised by the asymmetry.

contracts/src/token/ConfidentialFungibleTokenCore.compact
```diff
@@ -0,0 +1,1027 @@
...
+
+    // Randomness must be a VALID Jubjub scalar. `transientHash`'s raw Field output
+    // can exceed the Jubjub scalar-field order and fault `ecMulGenerator`, so each
+    // scalar is derived via `degradeToTransient(persistentHash(...))`, which
+    // truncates into the scalar field. The per-recipient count, folded into a
+    // Bytes<32> nonce, keeps each credit's randomness unique even if the seed
+    // repeats (essential for the memo OTP). This inlines the tag + expansion into
+    // a single hash per scalar. The residual persistentHash cost is poseidon-gated
+    // (a stable cheap hasher would cheapen it), same as the memo KDF.
+    const countHash = persistentHash<JubjubPoint>(ecMulGenerator(count));
```
0xisk: ❔ **question:** Is `ecMulGenerator(count)` here only to make a `Field` hashable with `persistentHash`? A cheaper Bytes/struct encoding of the count would trim a scalar-mul from the hottest circuits (`transfer`/`transferFrom` already sit near the k=16 row ceiling).

contracts/src/token/test/simulators/ConfidentialFungibleTokenSimulator.ts
```diff
@@ -0,0 +1,366 @@
...
+      const updated = { ...current, secretKey: newSK };
+      this.setPrivateState(updated);
+      return updated;
+    },
+
+    /**
+     * @description Replaces EK in the private state. Used in tests to inject
+     * a wrong EK and verify the decryption-consistency assertion catches it.
+     */
+    injectEncryptionKey: async (
```
0xisk, line 224: 🔵 **followup:** No test exercises `ElGamal: ek/pk mismatch` (wrong EK), even though `injectEncryptionKey` exists expressly for it; the wrong-plaintext path is covered but the EK-to-pk binding assert is not. An `approve(spender, 0)` revoke test is also missing.

contracts/src/token/ConfidentialFungibleTokenCore.compact
```diff
@@ -0,0 +1,1027 @@
...
+  // equal credit), which holds the invariant by construction. This base is
+  // supply-free; supply changes live in the layer that imports it.
+  //
+  // DANGER: `export circuit` means callable on the deployed contract, and
+  // Compact cannot restrict a re-export. A consumer that re-exports `_credit`
+  // turns it into a permissionless mint, and `_debit` into a permissionless
+  // burn. This base is a module, not independently deployable; a consuming
+  // contract MUST expose only the conserving surface (`transfer`, `_move`,
+  // `approve`/`transferFrom`, `sweep`) plus a supply layer's gated `mint`/`burn`,
+  // and MUST NEVER re-export `_credit` or `_debit`.
```
0xisk: 🔵 **followup:** This note names only `_credit`/`_debit`, but `_spendEscrow` is exported too and is similarly hazardous to re-export raw: a spender could destroy escrowed value with no supply decrement, so a supply layer's `totalSupply` would overstate circulation. Add it to the never-re-export list.

contracts/src/token/test/witnesses/ConfidentialFungibleTokenWitnesses.ts
```diff
@@ -0,0 +1,321 @@
...
+ *   (a) skip the recovery and cache the plaintext directly via
+ *       `ConfidentialFungibleTokenPrivateState.cachePlaintext` if the
+ *       value is known by construction, or
+ *   (b) implement a small brute-force recovery limited to small values
+ *       for use in tests only.
+ *
+ * @throws Always. Replace with a real implementation when memo decryption
+ * is needed end-to-end.
+ */
+export function recoverMemoPlaintext(
```
0xisk: 🔵 **followup:** Stale pre-ECDH remnants: `recoverMemoPlaintext` is typed for an ElGamal `Ciphertext` and documents BSGS over `[0, 2^48)`, but memos are now EcdhMask ciphertexts that the tests decrypt directly (no discrete log). The private-state doc above also still says memo plaintexts are "recovered via discrete-log search". Delete the stub or retype it for the EcdhMask memo and update the comments.

contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -0,0 +1,855 @@
...
+
+    await cft.transferFrom(ALICE.accountId, CHARLIE.accountId, 25n);
+
+    const charlieMemos = (await cft.getPublicState()).CFT__memos.lookup(
+      CHARLIE.accountId,
+    );
+    expect(charlieMemos.length()).toBe(1n);
+  });
+
+  it('reduces the allowance by the spent amount on a partial transferFrom', async () => {
```
0xisk: 🔵 **followup:** The invariant this suite leans on but never checks: both escrow copies stay equal (`_spendEscrow`'s "no separate bound check" note relies on it). Add a test of the shape approve 40 → spend 25 → re-approve, verifying the refund is exactly 15.

contracts/src/token/ConfidentialFungibleTokenCore.compact
```diff
@@ -0,0 +1,1027 @@
...
+    return disclose(accountId);
+  }
+
+  /**
+   * @description Credits `value` to `account`: homomorphically adds it to the
+   * account's balance ciphertext and pushes an ECDH one-time-pad memo that
+   * delivers `value` directly to the recipient (no discrete-log recovery, so no
+   * `2^48` bound). All per-credit randomness is derived internally from the
+   * witness seed and a per-recipient monotonic nonce (the memo-list length), so
+   * a reused seed cannot cause a mask/randomness collision (critical for the
```
0xisk: ⚪ **nitpick:** "a reused seed cannot cause a mask/randomness collision" overclaims: it holds only within a memo epoch, since `clearMemos` resets the count (the header and the section note both scope it that way). Align this sentence.

contracts/src/token/ConfidentialFungibleTokenCore.compact
```diff
@@ -0,0 +1,1027 @@
...
+    const rOwner   = ElGamal_expandRandomness(seed, pad(32, "approve_owner_ct"));
+
+    const newOwnerCt = ElGamal_subEncrypted(ownerCt, ownerPk, value, rDebit);
+    _balances.insert(disclose(ownerId), disclose(newOwnerCt));
+
+    const spenderPk = _encryptionKeys.lookup(disclose(spender));
+    const spenderCt        = ElGamal_encrypt(spenderPk, value, rSpender);
+    const ownerCtForRefund = ElGamal_encrypt(ownerPk,   value, rOwner);
+
+    if (!_escrow.member(disclose(ownerId))) {
```
0xisk: ⚪ **nitpick:** Dead branch: `_refundPriorEscrow` has already ensured `_escrow.member(ownerId)` by this point, so this re-init can be dropped (or the init kept in one place).

contracts/src/crypto/ElGamal.compact
```diff
@@ -470,4 +470,59 @@ module ElGamal {
...
+                   ct: Ciphertext,
+                   pk: JubjubPoint,
+                   ekScalar: Field,
+                   m: JubjubPoint
+                   ): [] {
+    assert(ecMulGenerator(ekScalar) == pk,
```
0xisk: ⚪ **nitpick:** If these variants stay, `assertDecryptsToPoint` can delegate here as `assertDecryptsToPointScalar(ct, pk, secretToScalar(ek), m)`, removing its duplicated body while keeping the single-hash optimization its comment describes.

contracts/src/token/ConfidentialFungibleTokenCore.compact
```diff
@@ -0,0 +1,1027 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Compact Contracts v0.3.0-alpha (token/ConfidentialFungibleToken.compact)
```
0xisk, line 2: ⚪ **nitpick:** Header says `token/ConfidentialFungibleToken.compact`; the file is `ConfidentialFungibleTokenCore.compact`.

contracts/src/token/ConfidentialFungibleTokenPublicSupply.compact
```diff
@@ -0,0 +1,243 @@
...
+
+  export circuit pendingOf(account: Bytes<32>): ElGamal_Ciphertext {
+    return CFT_pendingOf(account);
+  }
+
+  export circuit sweep(): Bytes<32> {
+    return CFT_sweep();
+  }
+
+  export circuit allowance(account: Bytes<32>, spender: Bytes<32>): EscrowEntry {
```
0xisk: ⚪ **nitpick:** Param is named `account` here but `owner` in the base module and the simulators. Align on `owner`.

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: Great catch! The proposed fix could work, but I think the cleaner fix is to embed `ownerMemo` in `EscrowEntry` to avoid any ambiguity on the allowance: [db5e2e7](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/db5e2e7a7ec905369295e9fdfba1ca484919f12a)

contracts/src/crypto/ElGamal.compact
andrew-fleming: Indeed, these are stale from the experimental branch. Removed [0677054](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/0677054b7518b570b9255cf6df2c310b46b18b56)

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: Good question. My thinking is that if users want to (for whatever reason) perform a self transfer, they can through `_move` if a contract wants to expose it with whatever additional logic. The cost of a self-transfer though is that the user needs to sweep the funds in order to make the tokens spendable again. The idea is to disallow it with normal transfers to prevent accidents (forcing users to sweep to spend again). The asymmetry is documented [3213368](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/32133688aef3c70dba25abc003b2b3dac3434a1b)

Happy to hear opposing arguments on this

contracts/src/token/ConfidentialFungibleTokenCore.compact
andrew-fleming: Correct and good call [8c34bbd](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/8c34bbd3d662ab88bd25a7426396af5557976e51)

contracts/src/token/test/simulators/ConfidentialFungibleTokenSimulator.ts
andrew-fleming: Done [f3b1c4f](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/f3b1c4f6591dffd698069bfc44c2a53d117913e0)

contracts/src/token/ConfidentialFungibleTokenCore.compact
andrew-fleming: Good catch [c7ad04f](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/c7ad04fa6f6320c0de8e57c955838f220f692717)

contracts/src/token/test/witnesses/ConfidentialFungibleTokenWitnesses.ts
andrew-fleming: Fixed [35785ce](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/35785ce14ddabb708984fb7f5a47c0ea291227f8)

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: Added [964c702](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/964c70253a12b42f3ac8951dc43aafc9be0af7ff)

contracts/src/token/ConfidentialFungibleTokenCore.compact
andrew-fleming: Agreed [23c3d57](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/23c3d57c3b99f0e1988b9c58766ac3647d305867)

contracts/src/token/ConfidentialFungibleTokenCore.compact
andrew-fleming: Removed [6c8ad14](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/6c8ad14dca6bd49c0227073750f0506264b299f3)

contracts/src/token/ConfidentialFungibleTokenCore.compact
andrew-fleming: [1bc0d7e](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/1bc0d7ef96d5f9d48cb458351dd083174457aaaf)

contracts/src/token/ConfidentialFungibleTokenPublicSupply.compact
andrew-fleming: Fixed [8b4eba8](https://github.com/OpenZeppelin/compact-contracts/pull/653/commits/8b4eba84ba617f5eae2800b9b20fcce1753b9004)

contracts/src/crypto/ElGamal.compact
andrew-fleming: removed `assertDecryptsToPointScalar` 👍

andrew-fleming:
> blocking: CI never ran this PR's tests: "Run Test Suite" hit timeout-minutes: 30 during yarn compact (step cancelled, type checks and tests skipped). CI compiles without SKIP_ZK, so it runs full key generation, and the two CFT mocks add ~6 k=15/16 circuits (transfer ~44k rows, transferFrom ~52k rows, in both mocks). Unblocks: a green run — bump the timeout, cache compiled artifacts, or compile with SKIP_ZK in the test job. FWIW the suite is green locally (80/80 in ~7s with SKIP_ZK).

@0xisk bumped the timeout to 60. IMO we should not skip zk during tests so that we don't get surprised with a "failed to generate keys/compile". And personally, I don't trust caching in the CI

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -0,0 +1,1027 @@
...
+   * @param {Bytes<32>} owner - The account that granted the allowance.
+   * @param {Bytes<32>} spender - The approved spender.
+   * @return {EscrowEntry} - The encrypted escrow entry.
+   */
+  export circuit allowance(owner: Bytes<32>, spender: Bytes<32>): EscrowEntry {
+    if (!_escrow.member(disclose(owner)) ||
+        !_escrow.lookup(disclose(owner)).member(disclose(spender))) {
+      return default<EscrowEntry>;
+    }
+    return _escrow.lookup(disclose(owner)).lookup(disclose(spender));
```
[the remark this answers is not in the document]
0xisk: +1 @andrew-fleming lets cover this please.

contracts/src/token/ConfidentialFungibleToken.compact
0xisk: sounds good :+1:

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -0,0 +1,1137 @@
...
+ * abstract `Bytes<32>` accountId) is already account-model-agnostic: it does
+ * not assume EOAs. If a contract key-custody scheme ever lands alongside c2c
+ * (e.g. a designated operator, threshold/MPC decryption, or a viewing-key
+ * arrangement), a contract obtains a `Bytes<32>` accountId via that scheme and
+ * the value logic is unchanged; only registration/authentication would. Until
+ * c2c exists AND such a custody design is defined, c2c is intentionally
+ * deferred. Interim: a protocol that needs to "hold" these tokens uses an
+ * EOA-custodied account it controls (supported today via `approve` /
+ * `transferFrom`), not the contract itself.
+ */
```
0xisk, line 193:
```suggestion
/**
 * @module ConfidentialFungibleToken
 * @description An account-based confidential fungible token module.
 *
 * Balances are stored as ElGamal ciphertexts on Jubjub. Account identity
 * follows the same pattern as `FungibleToken`: a witness-derived
 * `accountId = persistentHash(secretKey)`. Encryption uses a separate witness
 * `wit_ConfidentialTokenEK` to derive an ElGamal keypair `(ek, pk)` via
 * `ElGamal_derivePk`, where `pk = g^degradeToTransient(persistentHash(EK))`.
 *
 * @notice Supply. The mint/burn building blocks (`_mint`/`_burn`/`_burnFrom`)
 * live here; total-supply tracking is an optional add-on
 * (`extensions/ConfidentialFungibleTokenPublicSupply`) a composing contract pairs
 * with them if it wants a tracked supply.
 *
 * @notice Single-receiver. Auditor support, freeze, and admin-style seize are
 * deliberately left to companion modules or a future multi-receiver variant.
 * The escrow-spend primitive here (`_spendEscrow`) enables user-consenting
 * redemption (`_burnFrom`); admin-style seize against a non-cooperating owner
 * requires authority visibility into balances and is therefore out of scope
 * for the single-receiver base.
 *
 * @dev Trust model for witnesses.
 *
 * Witness implementations live in the user's wallet (TypeScript frontend) and
 * must NEVER be trusted to behave correctly. Every witness in this module is
 * either bound to publicly verifiable state (and the binding is checked by the
 * circuit) or used in a way where a hostile value cannot break the security
 * property the module claims.
 *
 * Witness audit:
 *
 *  - `wit_ConfidentialTokenSK`: verified by binding
 *    `accountId = persistentHash(SK)` to the on-chain account identifier used
 *    elsewhere in the circuit. A wrong SK produces a different accountId and
 *    the transaction fails at lookup.
 *
 *  - `wit_ConfidentialTokenEK`: verified by `ElGamal_assertDecryptsTo`,
 *    which re-derives `pk` and asserts it equals the on-chain stored pk.
 *
 *  - `wit_PlaintextBalance(ct)`: verified by `ElGamal_assertDecryptsTo`,
 *    which asserts `Dec(ct, EK) == claimedValue`. The ciphertext is passed
 *    as an argument so the wallet can look up the right cached plaintext
 *    for the specific ciphertext being verified, rather than relying on
 *    call-order conventions.
 *
 *  - `wit_RandomnessSeed`: supplies the single 32-byte seed all per-operation
 *    randomness is derived from. Unlike the witnesses above, it is NOT verified
 *    in-circuit; its freshness is load-bearing for confidentiality. See the
 *    Seed freshness @warning below.
 *
 * @warning Seed freshness. Seed freshness is load-bearing for confidentiality
 * on EVERY path. The seed-to-randomness expansion is deterministic and public,
 * so an observer who knows the seed can strip a memo's mask and read the exact
 * delivered amount (of any size), and can recover a balance or escrow amount
 * from its public ciphertext by discrete log. The seed MUST be fresh,
 * high-entropy, and secret for every invocation.
 *
 * Seed REUSE leaks plaintext differences, and the circuit does NOT prevent
 * it. The debit and escrow ciphertexts (`_debit`, `approve`, `_spendEscrow`,
 * and the re-approve refund) derive randomness from a static domain tag with
 * no per-operation nonce, so two operations under one seed share randomness
 * and their public ciphertext deltas reveal the amount difference. The credit
 * path folds the recipient's memo-list length in as a nonce (defense in
 * depth), but that counter RESETS on `clearMemos`, so it is an in-epoch
 * backstop only, never a substitute for freshness.
 *
 * @dev Plaintext disclosures (what reaches the public ledger):
 *  - This base discloses no amounts; supply accounting lives in a layer.
 *  - A public-supply layer reveals each mint/burn amount via the totalSupply
 *    delta (a confidential-supply layer would not).
 *  - transfer / transferFrom reveal no amounts.
 *  - approve does not disclose the cap (private-cap escrow design).
 *  - Counterparty graph (sender_id, recipient_id) is public on every
 *    transfer.
 *
 * @dev Per-transfer value bound. Value-bearing circuits check
 * `value <= MAX_TRANSFER_VALUE()`, which is now the `Uint<128>` maximum: the ECDH
 * memo delivers the amount directly (no discrete-log recovery), so the former
 * `2^48` cap that kept wallet-side BSGS tractable is gone. Amounts stay well under
 * the Jubjub scalar-field order, so the homomorphic balance sums (and the
 * `g^value` encoding) do not wrap. Balances remain Uint<128>.
 *
 * @dev Value delivery via memo channel. Every credit made through `_credit`
 * (mint / transfer / transferFrom) pushes an ECDH one-time-pad memo (see
 * `crypto/EcdhMask`) to the recipient's per-account memo list. The recipient's
 * wallet scans only its own list and decrypts each new entry DIRECTLY with its
 * encryption secret (no discrete-log search, no bounded-value table), updating
 * its local plaintext balance cache.
 *
 * Two caveats:
 *
 *  (1) Memos record CREDITS only: debits (sends, approvals, burns) move
 *      value with no memo and a hidden amount, so summing memos OVERCOUNTS
 *      a balance that has ever been spent from. The wallet's running
 *      plaintext cache, not memo replay, is the authoritative balance;
 *      replay-from-chain alone only reconstructs a never-spent account.
 *
 *  (2) A spender drawing on an escrow (`transferFrom`/`burnFrom`) reduces
 *      the owner's escrow copy with no balance memo; the owner instead
 *      learns the new remaining from the escrow entry's `ownerMemo`
 *      (refreshed on every spend), which is what lets them prove the
 *      post-refund balance on a later re-approve or revoke. Without that
 *      memo a partial spend would strand the escrow, since the owner cannot
 *      know the remainder folded back into their balance.
 *
 * @dev Hash function discipline. Every hash is built on `persistentHash` (no
 * direct use of `transientHash`). `accountId` is a raw `persistentHash` map key.
 * Deriving Jubjub scalars (pk derivation, the decryption check, and per-credit
 * randomness) applies `degradeToTransient` to a `persistentHash` output so the
 * result lands in the Jubjub scalar field. The memo mask KDF (`crypto/EcdhMask`) is
 * also `persistentHash`, and MUST stay persistent: the recipient reproduces it to
 * decrypt, possibly across a platform upgrade. These `persistentHash` (SHA256)
 * calls dominate the credit-path row counts; Midnight's roadmap includes a
 * cheaper stable (Poseidon-class) hasher, which would cut those counts
 * substantially with no source changes.
 *
 * @dev Caller-identity gating (composability). This module is unopinionated
 * about compliance policy (freeze, KYC, pause); a composing contract layers
 * that on top. Gating a circuit on an EXPLICIT argument (e.g. the recipient
 * `to`) is straightforward; the wrapper and this module see the same value.
 * Gating on the CALLER, however, is not: the caller is derived from
 * `wit_ConfidentialTokenSK`, and a separate witness call in the wrapper could
 * return a different secret than the one this module authenticates with,
 * letting a frozen account slip a clean id past the gate. To make caller-side
 * gates sound, every caller-authenticating circuit (`register`, `_debit`,
 * `transfer`, `_move`, `approve`, `transferFrom`, `sweep`, `_spendEscrow`)
 * RETURNS the caller's `accountId`, the principal it authenticated from the
 * witness. A
 * wrapper gates on that returned value; since an assert failing anywhere in a
 * Compact circuit invalidates the whole transaction, a post-call check still
 * prevents the operation.
 *
 * Invariant: the returned id is ALWAYS the caller (the sender / approver /
 * spender / registrant, depending on the circuit); it is never the recipient
 * `to` nor the `fromAddress` argument. Those are explicit arguments a wrapper
 * gates directly. A supply layer's `mint` is not caller-authenticating (its
 * recipient is an explicit argument), so it returns nothing.
 *
 * @warning Known limitations.
 *
 * Concurrency. Two transfers to the same recipient in the same block
 * conflict. Documented limitation; acceptable for the v1 target use case
 * (tokenized deposits).
 *
 * Block weight. A large composite deployment (notably the public-supply
 * variant, which layers the supply extension on the full token surface) can
 * exceed the node's per-block weight limit and fail to deploy with a
 * `BlockLimitExceeded` error. Keep the deployed circuit surface small (drop
 * unused getters, split oversized composites); raising the limit is a
 * node-side change, not a contract one. Tracked upstream:
 * https://github.com/midnightntwrk/midnight-node/issues/1202.
 *
 * Residual griefing. The dual-balance split (see `sweep`) stops dust
 * credits from churning a victim's spendable ciphertext, but every `_credit`
 * still appends a memo, and only the recipient can `clearMemos`. An attacker
 * can spam value-1 credits to grow a victim's memo list without bound; the
 * victim pays the storage, the wallet scan cost, and the clear gas. Rejecting
 * value == 0 does not help (dust at value == 1 defeats it). A real fix is
 * architectural (event-based delivery, a memo cap, or a bond) and is deferred.
 *
 * EOA-only. Contract-to-contract (c2c) is out of scope. Accounts here are
 * externally-owned: identity is `accountId = persistentHash(SK)` and spending
 * requires the holder to prove a balance plaintext under a secret encryption
 * key (`EK`) supplied via witness. A contract has no private state, so it
 * cannot custody such a key; it could be blindly credited (the homomorphic
 * add needs no key) but could never spend, decrypt its balance, or scan memos.
 * This is a cryptographic limitation, NOT a missing interface: adding a
 * `ContractAddress` account variant would not make a contract able to hold a
 * confidential balance, and clear contract addresses would also leak into the
 * otherwise-pseudonymous counterparty graph.
 *
 * Importantly, the value layer (balances / escrow / supply, all keyed by an
 * abstract `Bytes<32>` accountId) is already account-model-agnostic: it does
 * not assume EOAs. If a contract key-custody scheme ever lands alongside c2c
 * (e.g. a designated operator, threshold/MPC decryption, or a viewing-key
 * arrangement), a contract obtains a `Bytes<32>` accountId via that scheme and
 * the value logic is unchanged; only registration/authentication would. Until
 * c2c exists AND such a custody design is defined, c2c is intentionally
 * deferred. Interim: a protocol that needs to "hold" these tokens uses an
 * EOA-custodied account it controls (supported today via `approve` /
 * `transferFrom`), not the contract itself.
 */
```

0xisk (approved):
Great work! Thank you @andrew-fleming! 🚀

PR #657, opened 2026-07-03
ci(release): automate publishing with an approval gate

0xisk:
## What

Reworks the release pipeline into a dispatch-driven, two-line flow and gates npm publish behind a manual approval.

- **`prepare-release.yml`** — now `workflow_dispatch` with a `version` input; opens a signed `chore/bump-<version>` PR into the chosen branch (`main` for alphas, `release/X.Y.x` for rc/stable). Replaces the old `release-v*` branch trigger.
- **`create-release.yml`** (new) — on a version bump merged to `main` or `release/**`, cuts `v<version>` + a GitHub Release when the tag has no release yet (idempotent retries, no manual tag surgery). Uses the GitHub App token so the release event chains to `release.yml`.
- **`release.yml`** — publish job gated behind the `release` Environment.
- **`RELEASING.md`** — rewritten for the new flow with generic `X.Y.x` versioning.

## Flow

`prepare-release` (dispatch) → merge bump PR → `create-release` cuts the tag/release → `release` publishes to npm, paused at the approval gate. dist-tags: stable → `latest`, `-alpha`/`-rc` → `beta`.

## Closes

Closes #551
Closes #561
Closes #645

## Repo settings (verified present)

- [x] `release` Environment with required reviewers — approval gate is live
- [x] `GH_APP_ID` var + `GH_APP_PRIVATE_KEY` secret — bump + create→publish chain

## Summary by CodeRabbit

* **New Features**
  * Added automated release workflows to prepare version bumps, create GitHub Releases, and publish with an approval step.
  * Release creation now supports stable and pre-release versions, with safer reruns if a release already exists.

* **Documentation**
  * Updated release guidance to reflect the new workflow-based process, versioning rules, and required repository setup.

PR #658, opened 2026-07-03
ci(release): adopt npm OIDC trusted publishing

### LINKED ISSUE #560: Adopt npm OIDC trusted publishing for releases
**🧐 Motivation**

Cutting `v0.1.0` was blocked because `NPM_TOKEN` lacked publish/create rights on the `@openzeppelin` scope (`404 PUT .../@openzeppelin%2fcompact-contracts`). Long-lived npm tokens recur this failure class: they expire, need manual scope grants, and are a standing secret to rotate. npm OIDC trusted publishing removes the token entirely — Actions mints short-lived, repo-scoped credentials per run and npm verifies provenance.

**📝 Details**

- **npm side (DevOps):** configure OIDC trusted publishing for `@openzeppelin/compact-contracts`, bound to this repo + the publish workflow.
- **GitHub Actions side:** add `permissions: id-token: write` to the publish job; bump `actions/setup-node` + npm to versions supporting trusted publishing; remove `NPM_TOKEN` usage; gain provenance attestation. Ref: [`openzeppelin-ui/.github/workflows/publish.yml`](https://github.com/OpenZeppelin/openzeppelin-ui/blob/main/.github/workflows/publish.yml).

Outcome: a release publishes with no `NPM_TOKEN` secret, the package shows provenance on npm, and token expiry/scope failures can no longer block a release.

Relates to #551.

0xisk:
## What

Switches npm publishing to **OIDC trusted publishing** — no `NPM_TOKEN`.

- Remove `NODE_AUTH_TOKEN` / `NPM_TOKEN` from the publish step; npm mints short-lived, repo-scoped credentials per run via the OIDC `id-token` (`id-token: write` was already set).
- Ensure npm ≥ 11.5.1 (`npm install -g npm@^11.5.1`) — trusted publishing requires it, and the `.nvmrc` Node may bundle an older npm. Pinned to the vetted 11.x major so a release run can't pull an unvetted npm major (the rest of the workflow pins every tool).
- Single `setup-node` (the `.nvmrc` one) with `registry-url` set, so there is one Node source for build and publish.
- Attest SLSA build provenance over the packed tarball (`attest-build-provenance`) into the GitHub attestation store; complements the npm-registry provenance.
- Keep provenance (`NPM_CONFIG_PROVENANCE: true`).

Result: no standing npm secret to rotate, no token expiry/scope failures, provenance on npm.

## Notes

- Independent of [#657](https://github.com/OpenZeppelin/compact-contracts/issues/657) (release automation + approval gate). Both touch `release.yml` but in different regions (job-level gate vs. publish-step auth), so they merge cleanly in either order.
- Reference pattern: `OpenZeppelin/openzeppelin-ui/.github/workflows/publish.yml`.

Closes [#560](https://github.com/OpenZeppelin/compact-contracts/issues/560).

## Summary by CodeRabbit

* **Chores**
  * Updated package publishing to use secure OIDC-based authentication.
  * Added build provenance attestations to published packages.
  * Updated the release workflow configuration for the latest Node.js publishing environment.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,6 +14,7 @@ jobs:
     permissions:
       contents: read
       id-token: write
+      attestations: write # Write SLSA build provenance to the attestation store
 
     steps:
       - name: Harden Runner
@@ -99,6 +100,14 @@ jobs:
             echo "❌ Tarball contains private: true - cannot publish"
             exit 1
           fi
+
+      - name: Attest build provenance
+        # Signed SLSA v1 provenance over the exact tarball, stored in the
+        # GitHub attestation store (verifiable with `gh attestation verify`).
+        # Complements the npm-registry provenance from NPM_CONFIG_PROVENANCE.
+        uses: actions/attest-build-provenance@0f67c3f4856b2e3261c31976d6725780e5e4c373 # v4.1.1
+        with:
+          subject-path: ${{ steps.pack.outputs.tarball }}
 
       - name: Publish to npm
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,14 +28,21 @@ jobs:
       - name: Checkout
         uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
-          # No git writes here (publish goes through the npm token), so don't persist creds.
+          # No git writes here (publish uses npm OIDC, not git creds), so don't persist creds.
           persist-credentials: false
 
       - name: Setup Node.js
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version-file: ".nvmrc"
+          registry-url: https://registry.npmjs.org
           package-manager-cache: false # Prevent cache poisoning issues
+
+      - name: Ensure npm supports trusted publishing
+        # OIDC trusted publishing requires npm >= 11.5.1. The .nvmrc Node may
+        # bundle an older npm, so pin the floor explicitly: stays on the
+        # vetted 11.x major, avoids a surprise npm major in a release run.
+        run: npm install -g npm@^11.5.1
 
       - name: Enable Corepack
         run: corepack enable
@@ -60,13 +67,6 @@ jobs:
             exit 1
           fi
           echo "✅ Version consistency validated: $RELEASE_VERSION"
-
-      - name: Setup npm registry
-        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
-        with:
-          node-version: "24"
-          registry-url: https://registry.npmjs.org
-          package-manager-cache: false
 
       - name: Pack tarball
         id: pack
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -55,6 +59,11 @@ jobs:
         with:
           registry-url: "https://registry.npmjs.org"
 
+      - name: Ensure npm supports trusted publishing
+        # OIDC trusted publishing needs npm >= 11.5.1; the version bundled with
+        # the .nvmrc Node may be older.
+        run: npm install -g npm@latest
+
```
tirumerla: You can combine this 
```suggestion
         with:
             node-version: "24"
             registry-url: https://registry.npmjs.org
             package-manager-cache: false

```

tirumerla (approved):
This looks ok, i have couple comments
- We should add slsa actions to make sure this follows deterministic builds and push provenance to npm. You can see how it's being used https://github.com/OpenZeppelin/openzeppelin-ui/blob/main/.github/workflows/publish.yml

son-oz (approved):
LGTM.

Please keep in mind that the "Setup npm registry" step runs `actions/setup-node` with `registry-url`, which unconditionally writes `//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}` into `.npmrc`.

According to npmjs [docs](https://docs.npmjs.com/trusted-publishers/): "The npm CLI automatically detects OIDC environments and uses them for authentication before falling back to traditional tokens.", OIDC would still take precedence. However when `_authToken` is empty after we remove `NODE_AUTH_TOKEN`, the step may fail.

See https://github.com/OpenZeppelin/contracts-wizard/pull/797

son-oz:
I've added this workflow as a Trusted Publisher for `@openzeppelin/compact-contracts` on npmjs.

.github/workflows/release.yml
```diff
@@ -59,7 +64,9 @@ jobs:
       - name: Setup npm registry
         uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
-          registry-url: "https://registry.npmjs.org"
+          node-version: "24"
+          registry-url: https://registry.npmjs.org
+          package-manager-cache: false
```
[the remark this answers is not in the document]
0xisk: Good catch — valid concern. Fixed in 7a16ff8d, going a bit beyond the suggestion:

- Consolidated the two `setup-node` steps into the single `.nvmrc`-based one and moved `registry-url` onto it, so build and publish share one Node source. The hardcoded `node-version: "24"` step is gone.
- Reusing `.nvmrc` alone still would not *guarantee* npm ≥ 11.5.1 (it depends on the bundled npm), so I added an explicit `npm install -g npm@^11.5.1` before publishing. Pinned the floor to the vetted 11.x major rather than `latest`, consistent with the rest of the workflow pinning every tool by version/SHA.

pepebndc (approved):

son-oz (approved):

.github/workflows/release.yml
tirumerla, line 91: You don't have to separately pack tarball, you should use https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/builder_nodejs_slsa3.yml which will automatically runs builds, tests, packs and generates tarball and provenance. You then later download the output artifacts from this workflow which will be tarball and provenance that you will upload when publishing.


You can see it here as well how we did ithttps://github.com/OpenZeppelin/defender-sdk/blob/main/.github/workflows/publish.yml


Does this makes sense?

PR #661, opened 2026-07-07
Fix multisig partial spends

### LINKED ISSUE #656: [Bug]: Change coin persisted after sendImmediateShielded is unspendable (NullifierAlreadyPresent) — ShieldedTreasury, ShieldedTreasuryStateless, ForwarderPrivate
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/OpenZeppelin/midnight-contracts/CONTRIBUTING.md#creating-an-issue).
- [x] I agree to follow the [Code of Conduct](https://github.com/OpenZeppelin/midnight-contracts/CODE_OF_CONDUCT.md).
- [x] I have searched for [existing issues](https://github.com/OpenZeppelin/midnight-contracts/issues) that already report this problem, without success.

### What are the steps to reproduce this issue?

1. Run a real node (local standalone devnet: midnight-node:1.0.0, indexer-standalone:4.3.3, proof-server:8.1.0). A simulator does not reproduce this: the failure lives in the ledger's nullifier set, which only a node enforces.
2. Deploy a contract using the ShieldedTreasury receive/send pattern. A verbatim port of `_send` (minus the accounting maps) is sufficient.
3. Deposit a shielded coin of value 400 (receiveShielded + insertCoin).
4. Call the send circuit for 150. The transaction lands and the ledger map now holds a 250 "change" entry, qualified with a valid mt_index.
5. Call the send circuit for the remaining 250. The proof succeeds; the node rejects the transaction as an attempted double spend.

Runnable end-to-end harness (single script brings up the devnet and runs the exact sequence, including the fixed variant): https://github.com/midnightntwrk/passport/pull/89, directory experiments/stateless-shielded-custody, command ./run-all.sh --tests w3. The treasury-pattern reproduction is phase 3 of that probe.

### What happens?

The change produced by a partial spend is unspendable. sendImmediateShielded does not simply re-own the coin it is given: it transiently spends it (the coin's nullifier is revealed in that same transaction) and creates a replacement coin with an evolved nonce, returned as the `sent` field of its ShieldedSendResult. The contracts discard that return value and persist or hand back result.change.value, which is the coin that was just burnt.

Because the burnt coin's commitment is genuinely in the tree, a later spend of it proves successfully; the node then rejects the transaction with Zswap(NullifierAlreadyPresent). Through midnight-js this surfaces as substrate "Custom error: 103" (the node maps the variant to ZswapInvalidErrorCode::Unknown); the true error is only visible in the node logs.

Affected sites on main @ c096e317e8 (v0.3.0-alpha):
- ShieldedTreasury.compact `_send`: persists the burnt coin on-chain via insertCoin — https://github.com/OpenZeppelin/compact-contracts/blob/c096e317e8/contracts/src/multisig/ShieldedTreasury.compact#L120-L132
- ShieldedTreasuryStateless.compact `_sendFromCoin`: returns `result`, whose .change is the burnt coin, to the caller whose documented job is to store it for the next spend — https://github.com/OpenZeppelin/compact-contracts/blob/c096e317e8/contracts/src/multisig/ShieldedTreasuryStateless.compact#L60-L75
- ForwarderPrivate.compact: same shape — https://github.com/OpenZeppelin/compact-contracts/blob/c096e317e8/contracts/src/multisig/ForwarderPrivate.compact#L157-L171

As far as we can tell, no code in the repository uses the return value of sendImmediateShielded. For ShieldedTreasury the consequence is that after the first partial spend of a colour, the remaining funds are stuck behind a map entry no circuit can spend.

### What were you expecting to happen?

When the contract pays out part of a coin and keeps the change, the change
should be spendable in a later transaction.

Inside the send, the change coin produced by sendShielded (coin A) is
re-registered under the contract by sendImmediateShielded — which works by
SPENDING coin A and creating a brand-new coin B with a different nonce.
Coin A is consumed on the spot; coin B is what the contract actually owns
afterwards, and sendImmediateShielded returns coin B as the `.sent` field
of its result.

The contracts store coin A (result.change.value) instead of coin B. Since
coin A was already consumed when it was re-registered, every later spend
of the stored coin is rejected by the node as a double spend, while coin B
sits on-chain unspendable because its description was never recorded.

Expected: the persisted (or returned) coin is coin B, i.e. the `.sent`
field of the sendImmediateShielded result.

The one-line fix validates on the same harness: replacing the persisted coin with `redeposit.sent` makes the identical deposit / partial-spend / change-spend sequence pass.

### Paste any relevant logs, error output, etc.

```shell
2026-07-03 10:37:11 attempted double spend inp.nullifier=Nullifier(0385a4aecce1464a3236ba668001cf1ad84e912e40c121c73ab3d7baeb143646)
2026-07-03 10:37:11 🚫 Rejected transaction d0701efd2b71d41bbba27282a68c824e9a7a5b9ba7fc409833b92bc6dce99ec1 from mempool: guaranteed execution would fail: Zswap(NullifierAlreadyPresent(Nullifier(0385a4aecce1464a3236ba668001cf1ad84e912e40c121c73ab3d7baeb143646)))

SDK side (midnight-js): SubmissionError: Transaction submission error ... 1010: Invalid Transaction: Custom error: 103
```

### Additional context

Suggested fix, treasury:

```
  if (disclose(result.change.is_some)) {
    const redeposit = sendImmediateShielded(
      disclose(result.change.value),
      Utils_selfAsRecipient(),
      disclose(result.change.value.value)
    );
    _coins.insertCoin(disclose(color), disclose(redeposit.sent), Utils_selfAsRecipient());
  }
```

For _sendFromCoin and ForwarderPrivate, return a result whose change carries the re-owned coin:

```
  return ShieldedSendResult { sent: result.sent, change: some<ShieldedCoinInfo>(disclose(redeposit.sent)) };
```

Why the test suite misses it: the simulator does not enforce the ledger's nullifier set, and the existing tests stop one step short (they verify the change entry exists but never spend it). A regression test needs deposit, partial spend, then spend of the change, against a node or any nullifier-enforcing harness.

Client-side note: the re-owned coin is also observable in the transaction's private nextZswapLocalState.outputs, which may be worth documenting alongside the ShieldedMultiSigV3 discovery flow.

### Code Reproduction URL

https://github.com/midnightntwrk/passport/pull/89

### Version

0.31.0 (Default)

andrew-fleming:
Resolves #656.

This PR proposes to remove the change routing logic because `sendShielded` already routes the change coin to the contract. Implementing contracts can use the `ShieldedSendResult` to decide what to do with the change: persist it and pass it back on the next spend, or spend it onward to a different recipient in the same tx via `sendImmediateShielded`

The redundant re-send is dropped at all three sites:
- `ShieldedTreasury._send` records the live change coin
- `ShieldedTreasuryStateless._send` and `ForwarderPrivate._drain` return it untouched for the caller to handle

This PR also adds simulator regression tests that assert on the recorded zswap i/o. The retained change coin's nullifier must not appear among the tx inputs since that's the one thing observable without a node. Note that this only can catch double spends within the same tx and should not be used by itself as a standalone testing pattern

## Summary by CodeRabbit

* **New Features**
  * Improved shielded transfer flows so any leftover “change” is now handled safely and can be used in later transactions.
  * Added support for routing returned change to a different recipient in supported multisig flows.

* **Bug Fixes**
  * Fixed change handling to avoid accidental double-spends when sending partial amounts.
  * Sending the full balance now correctly produces no change output.

### DIFF
--- a/contracts/src/multisig/ForwarderPrivate.compact
+++ b/contracts/src/multisig/ForwarderPrivate.compact
@@ -99,8 +99,10 @@ module ForwarderPrivate {
    * matching the stored commitment; the coin is sent as a plain shielded note,
    * so the parent stays hidden on-chain. If `coin.value` exceeds `value`,
    * `sendShielded` routes the change back to this contract as a self-owned
-   * output (dwelling at the contract address for future drains); it is
-   * returned untouched via `result.change` and must not be re-spent here.
+   * output; `_drain` returns it untouched via `result.change`. The
+   * implementing contract chooses what to do with it: leave it dwelling at
+   * the contract for a future drain, or spend it onward to a different
+   * recipient in the same transaction via `sendImmediateShielded`.
    *
    * The parent is a `ZswapCoinPublicKey`, never a `ContractAddress`: a shielded
    * send to a contract publishes that contract's address in cleartext on the
--- a/contracts/src/multisig/ForwarderPrivate.compact
+++ b/contracts/src/multisig/ForwarderPrivate.compact
@@ -97,8 +97,10 @@ module ForwarderPrivate {
    * @description Spends a previously-deposited shielded coin to the parent,
    * a coin public key. The caller proves knowledge of `(parent, opSecret)`
    * matching the stored commitment; the coin is sent as a plain shielded note,
-   * so the parent stays hidden on-chain. If `coin.value` exceeds `value`, the
-   * change is re-emitted back to the contract for future drains.
+   * so the parent stays hidden on-chain. If `coin.value` exceeds `value`,
+   * `sendShielded` routes the change back to this contract as a self-owned
+   * output (dwelling at the contract address for future drains); it is
+   * returned untouched via `result.change` and must not be re-spent here.
    *
    * The parent is a `ZswapCoinPublicKey`, never a `ContractAddress`: a shielded
    * send to a contract publishes that contract's address in cleartext on the
@@ -160,14 +162,14 @@ module ForwarderPrivate {
       disclose(value)
     );
 
-    if (disclose(result.change.is_some)) {
-      sendImmediateShielded(
-        disclose(result.change.value),
-        Utils_selfAsRecipient(),
-        disclose(result.change.value.value)
-      );
-    }
-
+    // `sendShielded` already emits any change as a fresh output owned by this
+    // contract (dwelling at the contract address for future drains). `_drain`
+    // returns it untouched: it must not re-spend the change here and then hand
+    // that same (now-nullified) coin back via `result.change` — that is the
+    // double spend a node rejects on the next drain. A caller is free to spend
+    // the returned change onward itself (e.g. route it to another recipient);
+    // the rule is only that the change must never be re-spent AND still treated
+    // as held.
     return result;
   }
 
--- a/contracts/src/multisig/ShieldedTreasury.compact
+++ b/contracts/src/multisig/ShieldedTreasury.compact
@@ -85,8 +85,9 @@ module ShieldedTreasury {
    * @description Sends shielded tokens from the treasury.
    *
    * Looks up the stored coin by color, verifies sufficient value,
-   * and executes the shielded send. If the send produces change,
-   * it is sent back to the contract via `sendImmediateShielded`.
+   * and executes the shielded send. `sendShielded` already routes any
+   * change back to this contract as a self-owned output; the change coin
+   * is simply recorded in the UTXO map for the next spend.
    *
    * @notice Access control is NOT enforced here.
    * The consuming contract must gate this behind its own
@@ -119,14 +120,13 @@ module ShieldedTreasury {
 
     const result = sendShielded(coin, disclose(recipient), disclose(amount));
 
+    // `sendShielded` spends the stored coin (revealing its nullifier) and,
+    // when there is change, emits it as a fresh output owned by this
+    // contract. That change coin is already spendable. Do NOT re-spend it
+    // with `sendImmediateShielded`, which would reveal its nullifier in this
+    // same transaction and leave the recorded coin permanently double-spent.
     if (disclose(result.change.is_some)) {
-      const changeCoin = result.change.value;
-      sendImmediateShielded(
-        changeCoin,
-        Utils_selfAsRecipient(),
-        changeCoin.value
-      );
-      _coins.insertCoin(disclose(color), changeCoin, Utils_selfAsRecipient());
+      _coins.insertCoin(disclose(color), result.change.value, Utils_selfAsRecipient());
     } else {
       _coins.remove(disclose(color));
     }
--- a/contracts/src/multisig/ShieldedTreasury.compact
+++ b/contracts/src/multisig/ShieldedTreasury.compact
@@ -86,8 +86,11 @@ module ShieldedTreasury {
    *
    * Looks up the stored coin by color, verifies sufficient value,
    * and executes the shielded send. `sendShielded` already routes any
-   * change back to this contract as a self-owned output; the change coin
-   * is simply recorded in the UTXO map for the next spend.
+   * change back to this contract as a self-owned output. Unlike the
+   * stateless variant, this treasury RETAINS the change: it is recorded in
+   * the UTXO map for the next spend. The returned `result.change` therefore
+   * belongs to the treasury and must not be re-spent by the caller (doing
+   * so would double-spend the recorded coin).
    *
    * @notice Access control is NOT enforced here.
    * The consuming contract must gate this behind its own
--- a/contracts/test-utils/zswap.ts
+++ b/contracts/test-utils/zswap.ts
@@ -11,7 +11,7 @@
  * in particular, a circuit that emits a change coin, immediately re-spends it
  * (revealing its nullifier), and then persists/returns that same coin as if it
  * were still spendable. See the `ShieldedTreasury` / `ForwarderPrivate`
- * change-handling regression tests.
+ * "change coin is spendable" tests.
  */
 
 type ShieldedCoinInfo = { nonce: Uint8Array; color: Uint8Array; value: bigint };
--- a/contracts/test-utils/zswap.ts
+++ b/contracts/test-utils/zswap.ts
@@ -1,0 +1,122 @@
+/**
+ * @description Test helpers for inspecting the Zswap inputs (coin spends /
+ * nullifiers) and outputs (coin commitments) a circuit produces in the
+ * simulator.
+ *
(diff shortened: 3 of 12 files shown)


contracts/src/multisig/ShieldedTreasury.compact
```diff
@@ -119,14 +123,13 @@ module ShieldedTreasury {
...
+    // same transaction and leave the recorded coin permanently double-spent.
     if (disclose(result.change.is_some)) {
-      const changeCoin = result.change.value;
-      sendImmediateShielded(
-        changeCoin,
-        Utils_selfAsRecipient(),
-        changeCoin.value
-      );
-      _coins.insertCoin(disclose(color), changeCoin, Utils_selfAsRecipient());
+      _coins.insertCoin(disclose(color), result.change.value, Utils_selfAsRecipient());
```
0xisk, line 132: ❔ **question (non-blocking):** This stores `result.change.value` with no preceding `receiveShielded` (unlike `_deposit`). That's correct only if `sendShielded` auto-claims its self-routed change, so `insertCoin` yields a spendable `QualifiedShieldedCoinInfo` with a valid `mt_index`. Confirmed on a node? A live spend of the retained change would settle it (see the `zswap.ts` note).

_sent by claude (dev3-midnight-basic-review)_

contracts/test-utils/zswap.ts
```diff
@@ -0,0 +1,122 @@
+/**
+ * @description Test helpers for inspecting the Zswap inputs (coin spends /
+ * nullifiers) and outputs (coin commitments) a circuit produces in the
+ * simulator.
+ *
+ * The dry (in-memory) simulator does NOT enforce the ledger's nullifier set,
```
0xisk, line 6: 🔵 **followup (non-blocking):** These helpers read the **dry** backend, so the new assertions prove only same-tx non-re-spend. The actual #656 failure was the *next* tx spending the retained change against a node's nullifier set. A live-backend regression — deposit → partial `_send` → spend the retained change on a fresh node — is what fully closes #656. (The PR body already flags this pattern isn't sufficient standalone; noting so the live test isn't lost.)

_sent by claude (dev3-midnight-basic-review)_

contracts/test-utils/zswap.ts
```diff
@@ -0,0 +1,122 @@
...
+export function zswapLocalState(sim: unknown): ZswapLocalState {
+  // The circuit context lives on the wrapped synchronous simulator inside the
+  // backend; the public simulator surface intentionally does not re-export it.
+  const state = (
+    sim as {
+      _backend?: {
+        sim?: { circuitContext?: { currentZswapLocalState?: ZswapLocalState } };
+      };
+    }
+  )?._backend?.sim?.circuitContext?.currentZswapLocalState;
```
0xisk, line 58: ⚪ **nitpick (if-minor):** Reaches through two undocumented simulator layers — `_backend.sim` (`sim` is `private` on `DryBackend`) then `.circuitContext`. It fails loud (throws with a clear message) so it won't rot silently, but consider an upstream ask for a supported zswap-local-state accessor now that the simulator is dual-backend.

_sent by claude (dev3-midnight-basic-review)_

contracts/test-utils/zswap.ts
```diff
@@ -0,0 +1,122 @@
...
+    );
+  }
+  return state as ZswapLocalState;
+}
+
+/** Hex-encodes a byte string — handy for comparing coin/recipient bytes. */
+export const bytesToHex = (b: Uint8Array): string =>
+  Buffer.from(b).toString('hex');
+
+const toHex = bytesToHex;
```
0xisk, line 72: ⚪ **nitpick (if-minor):** `const toHex = bytesToHex` is a second name for the same function in one small file — use `bytesToHex` directly in `isNonceSpent` and drop the alias.

_sent by claude (dev3-midnight-basic-review)_

contracts/src/multisig/ForwarderPrivate.compact
```diff
@@ -97,8 +97,12 @@ module ForwarderPrivate {
    * @description Spends a previously-deposited shielded coin to the parent,
    * a coin public key. The caller proves knowledge of `(parent, opSecret)`
    * matching the stored commitment; the coin is sent as a plain shielded note,
-   * so the parent stays hidden on-chain. If `coin.value` exceeds `value`, the
-   * change is re-emitted back to the contract for future drains.
+   * so the parent stays hidden on-chain. If `coin.value` exceeds `value`,
+   * `sendShielded` routes the change back to this contract as a self-owned
+   * output; `_drain` returns it untouched via `result.change`. The
+   * implementing contract chooses what to do with it: leave it dwelling at
+   * the contract for a future drain, or spend it onward to a different
```
0xisk, line 104: 🔵 **followup (non-blocking):** This guidance ("spend it onward to a different recipient … via `sendImmediateShielded`") carries no recipient-type caveat, but for *this* module routing change to a `ContractAddress` republishes a contract address in cleartext — the exact leak the module forbids for `parent`. Add: onward change routing here must stay to a `ZswapCoinPublicKey`, same reason as the parent. (The generic treasuries don't carry the parent-privacy guarantee, so their identical wording is fine.)

_sent by claude (dev3-midnight-basic-review)_

0xisk (commented):
> **Basic review — PR #661 (Fix multisig partial spends)**
> 
> Inline comments cover one mechanism question, two followups, and two nitpicks. One more followup that can't be inline (the file isn't in this diff):
> 
> 🔵 **followup (non-blocking):** preset docs. `ShieldedMultiSigV2.execute` and the forwarder preset `drain` now surface a *live* `result.change` to dApp integrators, but their `@returns` still say only "including any change." Mirror the base-module doc improvement: state that the caller now owns the returned change (persist-and-replay, or route onward) and must not treat it as auto-retained.
> 
> **Verdict: ok — 0 blocking, 1 question, 3 followups, 2 nitpicks.** Ready to merge; the live-backend regression (noted inline on `zswap.ts`) is the item that truly closes #656.
> 
> _sent by claude (dev3-midnight-basic-review)_


Thank you @andrew-fleming! LGTM! I will just wait to verify all on local live infra.

contracts/src/multisig/ShieldedTreasury.compact
andrew-fleming: Confirmed on the local node...if preferred, I can create a new repo or something with the scripts to run so it's verifiable and not "just trust me, bro"

0xisk (approved):
LGTM! thank you @andrew-fleming! There will be a following PR that will have all the live infra testing.

contracts/src/multisig/ShieldedTreasury.compact
0xisk: No need we just need to push the live harness for running tests on local node.

Issue #672, opened 2026-07-10
dev: run utils specs on the live harness


0xisk:
Parent: #665
Depends on the core harness (#666).

## Scope

Run the `src/utils` specs (1 spec file) against the live harness (`unit-live`,
`MIDNIGHT_BACKEND=live`) where they add value over dry.

- [ ] Audit whether the utils circuits touch on-chain state; if they are pure,
      document why this category stays dry-only rather than adding a live run.

Not started.

Issue #668, opened 2026-07-10
dev: run token specs on the live harness


0xisk:
Parent: #665
Depends on the core harness (#666).

## Scope

Run the `src/token` specs (10 spec files) against the live harness (`unit-live`,
`MIDNIGHT_BACKEND=live`) and make the on-chain-relevant ones pass end to end.

- [ ] Audit which specs exercise real coin flows: mint, burn, transfer, and
      balance reads for shielded and unshielded tokens (vs pure logic that adds
      no live value).
- [ ] Wire the fixtures / pooled signers the coin paths need.
- [ ] Green on `make env-up`.

Not started.

Issue #665, opened 2026-07-10
Live-backend test harness for the contract suite


0xisk:
## Motivation

The contract suite's specs run against the dry simulator only, so live-only
failure modes go uncaught: real deploys, ZK proofs, on-chain state, per-signer
wallets, dust fees, and indexer lag.

This umbrella tracks a **live-backend test harness** that reruns the existing unit specs against a local node (`make env-up`) under `MIDNIGHT_BACKEND=live` (the `unit-live` vitest project), rolled out one module category at a time.

PR #673, opened 2026-07-10
test: live-backend harness for the contract suite (core + multisig)

### LINKED ISSUE #666: dev: core live-backend harness infrastructure
Parent: #665

## Scope

The reusable live-backend harness that lets any unit spec run against a local
node, plus the generic fixes found standing it up.

- [x] Live backend for `@openzeppelin/compact-simulator` (`registerLiveBackend`)
      + `LiveSimulatorBackend` (deploy-per-`create()`, per-alias providers).
- [x] `WalletPool` + `FundedWallet`: one funded wallet per signer alias, with a
      published `MIDNIGHT_<ALIAS>_COIN_PK`.
- [x] `funding` / `dust` / `network` helpers; `live.setup.ts` composition root;
      the `unit-live` vitest project.
- [x] Share one in-memory private-state store across aliases (was per-alias
      LevelDB lock contention + coin-PK scoping).
- [x] Top up a signer the dev preset leaves empty (transfer from the deployer).
- [x] Re-sync the wallet before balancing a tx (stale-UTXO race that surfaced
      as a bare "Transaction submission error" on consecutive same-signer txs).

Delivered on `test/multisig-live-harness`.

### LINKED ISSUE #667: dev: run multisig shielded specs on the live harness
Parent: #665
Depends on the core harness.

## Scope

Run the multisig shielded preset specs against the live harness and make them
pass end to end.

- [x] Rename shielded presets by capability.
- [x] `ShieldedProposalMultisig` passes live: deploy, deposit, propose,
      approve / revoke, execute; multi-signer authorization across pooled
      signer wallets.
- [x] Raise the live `testTimeout` to 600s for the ~10-tx lifecycle specs.

Out of scope (known later cleanup): move the heavy `full lifecycle` specs to the
integration project.

Delivered on `test/multisig-live-harness`. Relates to #470.

0xisk:
## Types of changes

Closes #666
Closes #667

Stands up a **live-backend test harness** so the existing unit specs can run
against a local Midnight node (`make env-up`, `MIDNIGHT_BACKEND=live`, the
`unit-live` vitest project), and brings the multisig shielded specs green on it.
First two sub-issues of the umbrella #665.

### What's here

- **Live backend + harness** (`contracts/test-utils/harness/`): `WalletPool`,
  `FundedWallet`, `LiveSimulatorBackend` (deploy-per-`create()`, per-alias
  providers), plus `funding` / `dust` / `network` / `ledgerEvents` /
  `NativeShieldedTokenTracker` and the `live.setup.ts` composition root.
- **test-utils reorg**: `fixtures/` (address, zswap, shieldedKey,
  nativeShieldedToken) split from `harness/`; vitest configs consolidated into
  one `vitest.config.ts` with `unit` / `unit-live` / `integration` / `harness`
  projects.
- **Preset rename by capability**: `ShieldedMultiSig` → `ShieldedProposalMultisig`,
  `ShieldedMultiSigV2` → `ShieldedStatelessMultisig`,
  `ShieldedMultiSigV3` → `NativeShieldedTokenMultisig` (CHANGELOG updated).
- **Fixes found standing it up**:
  - share one in-memory private-state store across signer aliases (was
    per-alias LevelDB lock contention + coin-PK scoping);
  - fund the 4th signer the dev preset leaves empty (top-up from the deployer);
  - re-sync the wallet before balancing a tx, fixing a stale-UTXO race that
    surfaced as a bare "Transaction submission error" on consecutive
    same-signer submissions.
- Raise the live `testTimeout` to 600s for the ~10-tx lifecycle specs.

### Scope note

Large diff, but most of it is mechanical: test-utils import-path updates across
suites, `yarn.lock`, and the vitest config merge. The substantive new code is
`test-utils/harness/` + `test-utils/fixtures/`. No production contract behavior
changes beyond the preset renames.

## PR Checklist

- [x] I have read the Contributing Guide
- [ ] I have added documentation for new methods or changes to existing method behavior.

#### Further comments

Opened as a **draft**: dry suite, tsc, and biome are green, and the heaviest
live subset (`full lifecycle`, 3 tests) passes with the sync fix. Flipping to
ready once the full `ShieldedProposalMultisig` live suite (and the rest of the
multisig live specs) are confirmed green on a fresh node. The remaining category
sub-issues (#668–#672) roll out one at a time under #665.

## Summary by CodeRabbit

* **New Features**
  * Added native shielded-token fixtures and support for dry and live test environments.
  * Added live test tooling with wallet funding, progress reporting, retry handling, and environment safeguards.
  * Added expanded multisig coverage for proposal-based and stateless variants.

* **Updates**
  * Replaced Compact workflows with Compile workflows across build and test commands.
  * Updated multisig naming and simplified proposal and treasury access patterns.

* **Documentation**
  * Added comprehensive test-running guidance and refreshed multisig changelog details.

* **Bug Fixes**
  * Corrected bug-report instructions and disabled an unreliable proof-server health check.


contracts/src/multisig/test/ShieldedTreasury.test.ts
```diff
@@ -18,27 +27,38 @@ const AMOUNT = 1000n;
 // THIS contract, not merely to "some contract arm".
 const TREASURY_ADDRESS = '7a'.repeat(32);
 
-const Z_RECIPIENT = utils.createEitherTestUser('RECIPIENT');
+// Assigned in `beforeEach` after `create()` syncs the wallet: on live the harness
+// then publishes MIDNIGHT_DEPLOYER_COIN_PK, so this resolves to the deployer's own
+// coin public key (an encryption key the node can resolve); dry → a synthetic user.
+let Z_RECIPIENT: ReturnType<typeof shieldedTestRecipient>;
```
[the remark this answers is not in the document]
0xisk: Fixed in 10f7d9c7. Comment now reads `beforeAll` to match the hook that assigns `Z_RECIPIENT`.

contracts/src/multisig/test/simulators/ShieldedMultiSigSimulator.ts
```diff
@@ -139,8 +139,14 @@ export class ShieldedMultiSigSimulator extends ShieldedMultiSigSimulatorBase {
...
-  public getReceivedMinusSent(color: Uint8Array): Promise<bigint> {
-    return this.circuits.impure.getReceivedMinusSent(color);
+  // getReceivedMinusSent was dropped from the contract (redundant; removed to fit
+  // the deploy block limit). Derive it from the two tracked totals.
+  public async getReceivedMinusSent(color: Uint8Array): Promise<bigint> {
+    const [received, sent] = await Promise.all([
+      this.getReceivedTotal(color),
+      this.getSentTotal(color),
+    ]);
+    return received - sent;
```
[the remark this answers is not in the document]
0xisk: Fixed in 170fc903. `getReceivedMinusSent` awaits `getReceivedTotal` then `getSentTotal` sequentially; on live each impure getter submits a tx, so the concurrent version raced against one wallet snapshot.

contracts/test-utils/harness/FundedWallet.ts
```diff
@@ -0,0 +1,125 @@
...
+    // unit-test mock provider may omit it.
+    if (typeof provider.balanceTx === 'function') {
+      const balanceTx = provider.balanceTx.bind(provider);
+      provider.balanceTx = (async (...args: Parameters<typeof balanceTx>) => {
+        await syncWallet(wallet);
+        return balanceTx(...args);
+      }) as typeof provider.balanceTx;
+    }
+
+    return new FundedWallet(alias, provider, nightBalance, dustBalance);
```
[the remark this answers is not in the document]
0xisk: Fixed in 129f9f0a. Wrapped init in try/catch and call `provider.stop()` on any failure, so a rejected build no longer leaks the started provider.

contracts/test-utils/harness/network.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Endpoints of the local stack (`local-env.yml`), shared by the live setup and
+ * the live harness smoke. `LocalTestConfiguration` builds the `127.0.0.1`
+ * `/api/v4/graphql` URLs from these ports; override per port via the
+ * `MIDNIGHT_*_PORT` env vars for a relocated stack.
+ */
+export const PORTS = {
+  indexer: Number(process.env.MIDNIGHT_INDEXER_PORT ?? 8088),
+  node: Number(process.env.MIDNIGHT_NODE_PORT ?? 9944),
+  proofServer: Number(process.env.MIDNIGHT_PROOF_SERVER_PORT ?? 6300),
+};
```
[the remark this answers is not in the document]
0xisk: Fixed in a0824d9d. Added a `port()` helper that rejects non-integer or out-of-range overrides at load time and names the responsible env var.

contracts/test-utils/harness/test/network.test.ts
```diff
@@ -0,0 +1,43 @@
...
+  });
+
+  describe('PORTS', () => {
+    it('should default to the local stack ports', () => {
+      expect(PORTS).toStrictEqual({
+        indexer: 8088,
+        node: 9944,
+        proofServer: 6300,
+      });
+    });
```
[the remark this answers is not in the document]
0xisk: Fixed in e7fa25c3. The default-port test now clears the `MIDNIGHT_*_PORT` vars and re-imports `network.js` before asserting the defaults.

contracts/test-utils/harness/WalletPool.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (!this.ready) {
+      this.ready = Promise.all(
+        Object.entries(this.seeds).map(async ([alias, walletSeed]) => {
+          const wallet = await this.buildWallet(alias, walletSeed);
+          process.env[coinPkEnv(alias)] = wallet.coinPublicKey;
+          this.wallets.set(alias, wallet);
+        }),
+      ).then(() => undefined);
+    }
+    return this.ready;
```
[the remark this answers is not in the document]
0xisk: Fixed in d3d038c1. `ensureReady` now builds wallets sequentially, so the funding txs no longer balance against the same deployer snapshot.

turbo.json
```diff
@@ -1,78 +1,103 @@
...
+    "compile": {
+      "dependsOn": [
+        "compile:crypto",
+        "compile:security",
+        "compile:utils",
+        "compile:access",
+        "compile:multisig",
+        "compile:token"
+      ]
     },
```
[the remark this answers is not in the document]
0xisk: Fixed in 62fc4e81. Added `compile:archive` to the `compile` aggregate so `yarn test` rebuilds stale `src/archive` artifacts.

0xisk:
@andrew-fleming added docs on how to run live test per specific test file. [2bdab60](https://github.com/OpenZeppelin/compact-contracts/pull/673/commits/2bdab60dbdf1306a59344d3d01efc1a6e86bb855)

0xisk:
### CI for the live suite: `live.yml`

The last 9 commits add a CI workflow for the live tests. It is separate from the normal checks and never required, since a full category takes hours, not minutes.

**Triggers**
- **Nightly on `main`** (02:00 UTC). A failed nightly opens or comments on a `live-nightly` tracking issue, which closes automatically on the next green run.
- **On demand**: `workflow_dispatch` with a `category` dropdown (`all` / `multisig`) and an optional vitest file `filter`.
- **On a PR**: apply the `live-tests` label (re-apply to re-run). Since only maintainers can label, this also gates fork code.

**Shape**
- A small `plan` job derives the live-ready list from the runner itself (`yarn test:live --list`, backed by `LIVE_READY` in `scripts/test-live.ts`), so there is no second list to maintain.
- A matrix fans out **one parallel job per category** (`live-multisig`, ...). Each gets its own runner, docker stack, compile, and 6-hour job budget; `fail-fast: false`, so one category cannot cancel the rest. Wall clock is the slowest category, not the sum.
- Each job runs the same `yarn test:live` runner as a local run, scoped to its category (`yarn test:live <category>`, the matrix value as the first arg): compile with ZK-key verification, stack reset via `make env-up`, harness smoke, then the two-round flake classification (flaky = exit 0, reported loudly). Only a bare local `yarn test:live` runs every category in one invocation; in CI the all-categories view is the sum of the parallel jobs.

**Observability**
- Each job writes its verdict (with flaky/real file lists) to the GitHub job summary, and every flaky file gets a `::warning::` annotation, so a green-but-flaky run stays visible.
- Artifacts: `live-reports-<category>` (JSON verdicts, uploaded on every run, 14 days) and `live-logs-<category>` (worker + service logs, failure-only, 7 days).

**Update (1bc30b3f, 6980a49e)**
- Dropped the unused `test:live` turbo task and the contracts `test:live` script: nothing invoked them since the unified runner, and they were a second entry point that skipped the stack reset, lock, and flake verdict. The runner is the single entry point for live tests.
- The `test-live.ts` header now documents why the runner is a script rather than turbo tasks: turbo models a DAG of stateless, cacheable tasks, while a live run needs stateful orchestration (two-round flake verdict, docker lifecycle against one shared node, ZK-key self-heal, infra-vs-test exit codes). Turbo still handles the DAG-shaped steps inside the run (cached compile, harness smoke).

**Notes**
- `schedule` and `workflow_dispatch` only activate once this lands on `main`; the `live-tests` label already works on this PR.
- The dispatch dropdown mirrors `LIVE_READY`; a stale entry fails loudly (runner exit 2). New categories join the nightly and label runs automatically via `--list`.
- Sizing: one file (`ShieldedProposalMultisig`) took ~63 min locally, so a category is a multi-hour job. If a category ever nears the 6-hour cap, the levers are sharding within the category or a larger runner.

.github/ISSUE_TEMPLATE/01_bug_report.yml
```diff
@@ -31,7 +31,7 @@ body:
       label: What are the steps to reproduce this issue?
       placeholder: |
         1. Clone repo
-        2. Run `yarn compact`
+        2. Run `yarn compile`
```
andrew-fleming: I agree with the change but let's keep this PR focused on just the harness for testing

.github/workflows/checks.yml
```diff
@@ -33,4 +33,4 @@ jobs:
           skip-compact: "true"
 
       - name: Format & Lint
-        run: yarn fmt-and-lint:ci
+        run: yarn check:ci
```
andrew-fleming: Ditto here with making these changes in a separate PR (also, there are lingering `fmt-and-lint` snippets in the README and CONTRIBUTING)

.github/workflows/live.yml
andrew-fleming, line 1: I think this is a little too big and complex to be included in this PR with the already-big and complex harness logic. I vote we get the harness working locally and then have a separate PR to integrate it into the CI

contracts/src/multisig/presets/NativeShieldedTokenMultisig.compact
```diff
@@ -1,10 +1,10 @@
 // SPDX-License-Identifier: MIT
-// OpenZeppelin Compact Contracts v0.0.1-alpha.1 (multisig/presets/ShieldedMultiSigV3.compact)
+// OpenZeppelin Compact Contracts v0.0.1-alpha.1 (multisig/presets/NativeShieldedTokenMultisig.compact)
```
andrew-fleming: I agree with the change but this doesn't have anything to do with the test harness. Remember these are user-facing contracts (or meant to be) so renaming contracts should have their own PR + changelog entry so that users can see the clean change (and not all the harness noise)

contracts/src/multisig/test/ShieldedMultiSigV3.test.ts
```diff
@@ -296,28 +330,47 @@ describe('ShieldedMultiSigV3', () => {
...
 
-      it('should burn with signers 0 and 2', async () => {
-        const coin = makeQualifiedCoin(await multisig.getTokenType(), 100n);
-        await multisig.burn(coin, 100n, [PK1, PK3], [DUMMY_SIG, DUMMY_SIG]);
-      });
+      it.skipIf(isLiveBackend())(
+        'should burn with valid coin and signers 0 and 1',
+        async () => {
+          const coin = makeQualifiedCoin(await multisig.getTokenType(), 100n);
+          await multisig.burn(coin, 100n, [PK1, PK2], [DUMMY_SIG, DUMMY_SIG]);
```
andrew-fleming: **nit**: we can group these tests in a describe block so we only need to use one `isLiveBackend()`. Not sure if it's worth it, just an idea

contracts/src/multisig/test/ShieldedMultiSig.test.ts
andrew-fleming, line 1: Do you mind using `git mv` so we have accurate diff here?

contracts/test-utils/fixtures/address.ts
```diff
@@ -66,6 +66,23 @@ export const createEitherTestUser = (
...
+ * @param coinPublicKey 64-char hex coin public key.
+ * @returns Defined Either object for the given ZswapCoinPublicKey.
+ */
+export const eitherUserFromCoinPublicKey = (
+  coinPublicKey: string,
+): Either<ZswapCoinPublicKey, ContractAddress> => ({
+  is_left: true,
+  left: { bytes: encodeCoinPublicKey(coinPublicKey) },
+  right: encodeToAddress(''),
+});
```
andrew-fleming, line 84: **followup** out of scope, but it probably makes sense to just use `eitherUserFromCoinPublicKey` and remove `createEitherTestUser` so there's no ambiguity

contracts/package.json
```diff
@@ -21,24 +21,23 @@
...
-    "compact:archive": "compact-compiler --dir archive",
-    "compact:crypto": "compact-compiler --dir crypto",
-    "compact:multisig": "compact-compiler --dir multisig",
-    "compact:security": "compact-compiler --dir security",
-    "compact:token": "compact-compiler --dir token",
-    "compact:utils": "compact-compiler --dir utils",
+    "compile:access": "compact-compiler --dir access",
+    "compile:archive": "compact-compiler --dir archive",
+    "compile:crypto": "compact-compiler --dir crypto",
+    "compile:multisig": "compact-compiler --dir multisig",
```
andrew-fleming: As mentioned in a previous comment, let's not muddy up the test harness with development api changes if we don't need to

turbo.json
```diff
@@ -1,78 +1,99 @@
...
-      "outputs": ["artifacts/**"]
+      "outputs": ["artifacts/**/"]
+    },
+    "compile": {
+      "dependsOn": [
+        "compile:crypto",
+        "compile:security",
+        "compile:utils",
+        "compile:access",
+        "compile:archive",
```
andrew-fleming: Why are we reintroducing archive to the global compile?

package.json
```diff
@@ -8,18 +8,31 @@
...
+    "test": "SKIP_ZK=true turbo run test --filter=@openzeppelin/compact-contracts --env-mode=loose --log-prefix=none",
+    "test:coverage": "SKIP_ZK=true turbo run test:coverage --filter=@openzeppelin/compact-contracts --env-mode=loose --log-prefix=none",
+    "test:live": "node --disable-warning=MODULE_TYPELESS_PACKAGE_JSON scripts/test-live.ts",
+    "test:live:multisig": "node --disable-warning=MODULE_TYPELESS_PACKAGE_JSON scripts/test-live.ts multisig",
+    "test:integration": "SKIP_ZK=true turbo run test:integration --filter=@openzeppelin/compact-contracts --env-mode=loose --log-prefix=none",
+    "test:harness": "turbo run test:harness --filter=@openzeppelin/compact-contracts --env-mode=loose --log-prefix=none",
+    "test:harness:live": "turbo run test:harness:live --filter=@openzeppelin/compact-contracts --env-mode=loose --log-prefix=none",
+    "check": "biome check .",
+    "check:fix": "biome check . --write",
+    "check:ci": "biome ci . --no-errors-on-unmatched",
```
andrew-fleming: Personally, I don't like `check`...check what exactly? Why not use `lint` like in compact-tools?

package.json
```diff
@@ -8,18 +8,31 @@
...
+    "compile:integration": "turbo run compile:integration --filter=@openzeppelin/compact-contracts --log-prefix=none",
     "build": "turbo run build --log-prefix=none",
-    "test": "turbo run test --filter=@openzeppelin/compact-contracts --log-prefix=none",
-    "fmt-and-lint": "biome check .",
-    "fmt-and-lint:fix": "biome check . --write",
-    "fmt-and-lint:ci": "biome ci . --no-errors-on-unmatched",
+    "test": "SKIP_ZK=true turbo run test --filter=@openzeppelin/compact-contracts --env-mode=loose --log-prefix=none",
+    "test:coverage": "SKIP_ZK=true turbo run test:coverage --filter=@openzeppelin/compact-contracts --env-mode=loose --log-prefix=none",
+    "test:live": "node --disable-warning=MODULE_TYPELESS_PACKAGE_JSON scripts/test-live.ts",
+    "test:live:multisig": "node --disable-warning=MODULE_TYPELESS_PACKAGE_JSON scripts/test-live.ts multisig",
```
andrew-fleming: This seems odd. Can't we just use `test:live <foo>`

.github/ISSUE_TEMPLATE/01_bug_report.yml
andrew-fleming: Case in point, the README still has snippets of `yarn compact` that are easy to miss

CONTRIBUTING.md
```diff
@@ -152,6 +157,85 @@ A maintainer will re-run the status check for you. If we conclude that the failu
 
 While the prerequisites above must be satisfied prior to having your pull request reviewed, the reviewer(s) may ask you to complete additional design work, tests, or other changes before your pull request can be ultimately accepted.
 
+## Running Tests
+
+Run all commands from the repository root, and always invoke Yarn through `corepack`.
```
andrew-fleming: This reads a little awkward. Consider that the CI and this CONTRIBUTING doc (lines 237 and 245) have yarn commands that don't explicitly run through corepack. Why not just "enable corepack" as a prerequisite instead of starting every command with `corepack` ?

andrew-fleming: If you agree, let's replace all `corepack yarn ...` with just `yarn ...`

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add blocklist (#626)
 - Add allowList (#625)
 - Add ElGamal module (#617)
-- Multisig contract suite under `contracts/src/multisig/`: configurable M-of-N `Signer` / `SignerManager` registry, `ProposalManager`, stateful `ShieldedTreasury` and `ShieldedTreasuryStateless`, `UnshieldedTreasury`, `Forwarder` + `ForwarderPrivate` modules with per-recipient presets, and the `ShieldedMultiSig` / `ShieldedMultiSigV2` presets. Signature verification is stubbed pending ECDSA + Keccak primitives (#475). (#378, #424, #526)
+- Multisig contract suite under `contracts/src/multisig/`: configurable M-of-N `Signer` / `SignerManager` registry, `ProposalManager`, stateful `ShieldedTreasury` and `ShieldedTreasuryStateless`, `UnshieldedTreasury`, `Forwarder` + `ForwarderPrivate` modules with per-recipient presets, and the `ShieldedProposalMultisig` / `ShieldedStatelessMultisig` presets. Signature verification is stubbed pending ECDSA + Keccak primitives (#475). (#378, #424, #526)
```
andrew-fleming: This would undermine the changelog's purpose

contracts/src/multisig/test/ShieldedTreasuryStateless.test.ts
```diff
@@ -1,57 +1,70 @@
...
+  contractOwner,
+  getQualifiedShieldedCoinInfo,
+} from '#test-utils/harness/NativeShieldedTokenTracker.js';
 import { MockShieldedTreasuryStatelessSimulator } from './simulators/MockShieldedTreasuryStatelessSimulator.js';
 
-const COLOR = new Uint8Array(32).fill(1);
+// Genesis-funded shielded color (`0x00…01`): on the live backend a `_deposit` /
+// `_send` can only draw a color the deployer wallet holds. `fill(1)` (`0x0101…01`)
+// is a different, unfunded type on live; on dry any color mints freely. See
+// nativeShieldedToken.ts.
```
andrew-fleming:
```suggestion
// `_send` can only draw a color the deployer wallet holds.
```
The last two sentences explain the justification for the change, but it doesn't make sense without that context

contracts/src/multisig/test/ShieldedTreasuryStateless.test.ts
```diff
@@ -87,118 +92,183 @@ describe('ShieldedTreasuryStateless', () => {
...
-        TREASURY_ADDRESS,
-      );
-      expect(isNonceSpent(inputs, result.change.value.nonce)).toBe(false);
-    });
+  //
+  // Dry-only: `zswapSnapshot`/`zswapDelta` read the dry sim's Zswap local state,
+  // which does not exist on the live backend. The live counterpart follows.
+  describe.skipIf(isLiveBackend())(
+    '_send — change coin is spendable, via Zswap I/O (dry only, no double spend)',
+    () => {
```
andrew-fleming, line 98: **followup**: Why not just insert `dry` as a description in the describe block with `isLiveBackend` and have the dry-only tests inside of `it` blocks?

turbo.json
andrew-fleming: Trying to compile is failing on my machine bc of this. When you delete it, it compiles fine

scripts/test-live.ts
```diff
@@ -0,0 +1,455 @@
...
+ * run may therefore pass in isolation on a fresh node. Hence two rounds:
+ *
+ *   Round 1: compile + harness smoke once, then per category: reset the stack
+ *            and run that category's files (parallel workers). Collect the
+ *            files that failed from the JSON reporter.
+ *   Round 2: reset the stack again, re-run only those files, one worker.
+ *
+ * A file that fails round 1 but passes round 2 is FLAKY (an environment
+ * artifact); one that fails both — or never reports in round 2 — is a REAL
+ * failure. Exit 0 unless there is a real failure, so an env flake never turns
```
andrew-fleming, line 34: Can't a an order-dependent flake re-fail in round 2 bc an earlier round-2 file sullied the node?

andrew-fleming (commented):
Excellent work, @0xisk! This is a partial review, but I left some comments and questions

Also, `make env-up` is not portable across Docker Compose versions (I can't highlight the makefile here). `yarn test:live` aborts during startup on my machine. My quick fix to get it working:

```make
docker compose -f $(COMPOSE_FILE) up -d
docker compose -f $(COMPOSE_FILE) up -d --wait node indexer
```

.github/ISSUE_TEMPLATE/01_bug_report.yml
0xisk: Agreed. Reverted here; the `compact`→`compile` rename (including the README `yarn compact` snippets) now lives in #680.

.github/workflows/checks.yml
0xisk: Agreed. Reverted here; `fmt-and-lint`→`lint` and the lingering README/CONTRIBUTING snippets moved to #680.

contracts/package.json
0xisk: Agreed. The dev-api rename is backed out of this PR and now lives in #680.

package.json
0xisk: Agreed, went with `lint` (matching compact-tools). Moved to #680, out of this PR.

.github/workflows/live.yml
0xisk: Agreed. Removed `live.yml` (and its "Live tests in CI" CONTRIBUTING subsection) from this PR. It now lives in #681, stacked on this branch so it retargets to `main` once this merges.

turbo.json
0xisk: Agreed, that was the bug. Dropped `compact:archive` from the aggregate `compact` task; archive is excluded from the full compile on `main` too, and the standalone `compact:archive` task stays for targeted runs. Fixed in 221692da.

0xisk:
Fixed the `make env-up` portability issue in 8cce7053. Root cause: `proof-server` has no healthcheck, so the blanket `docker compose up -d --wait` is not portable across Compose versions. It now starts all services, then waits only on `node` and `indexer` (the two with healthchecks), matching your suggested fix.

contracts/src/multisig/presets/NativeShieldedTokenMultisig.compact
0xisk: Agreed. Reverted the whole preset rename out of this PR (presets back to `ShieldedMultiSig*`) and moved it to #682: a dedicated PR, stacked on this branch, with its own changelog entry. Presets, tests, and simulators rename together there as clean git renames.

contracts/src/multisig/test/ShieldedMultiSig.test.ts
0xisk: Handled in #682. Here the file was only ~17% similar (git showed add/delete) because the harness rewrite landed in the same commit as the rename. #682 separates them: the harness edits stay here under `ShieldedMultiSig.test.ts`, and the pure rename in #682 shows as a clean rename (R096).

CHANGELOG.md
0xisk: Agreed. That edit rewrote the already-released 0.3.0-alpha entry. Reverting the preset rename here restored that line to `ShieldedMultiSig` / `ShieldedMultiSigV2`, so the CHANGELOG now matches main (empty diff). #682 records the rename as a new Unreleased `Changed` entry instead of editing the shipped one, so release history stays intact.

CONTRIBUTING.md
0xisk: Agreed. Done in 44aef934: enabled Corepack as a one-time prerequisite (`corepack enable`, already run in CI) and dropped the `corepack` prefix from every command in CONTRIBUTING and the runner. Plain `yarn` now resolves to the pinned version.

package.json
0xisk: Agreed, done in e8e490b4. The runner already scopes by a positional arg, so dropped `test:live:multisig` in favor of `yarn test:live multisig [filter]`. This also matches how CI already invokes it (`yarn test:live "$CATEGORY"`), and avoids one script per category as more become live-ready.

scripts/test-live.ts
0xisk: Good catch. Fixed in d4afe49b. Round 2 now resets the node before each failed file and re-runs just that file alone, so state left by an earlier round-2 file can no longer fail a later one. Each file gets the fresh node the "passes in isolation" check assumes. Round 2 only re-runs the failed files, so the extra resets are cheap.

contracts/src/multisig/presets/ShieldedMultiSig.compact
```diff
@@ -187,17 +187,12 @@ export circuit getProposal(id: Uint<64>): Proposal_Proposal {
...
-
-export circuit getProposalColor(id: Uint<64>): Bytes<32> {
-  return Proposal_getProposalColor(id);
-}
+// NOTE: the per-field proposal getters (getProposalRecipient / getProposalAmount
+// / getProposalColor) and getReceivedMinusSent were dropped to bring the deploy
+// transaction under the block-weight limit (the full 19-circuit deploy is
+// rejected with "Transaction would exhaust the block limits"). They are all
+// redundant: read the proposal fields via `getProposal(id).to / .amount / .color`,
+// and the net balance via `getReceivedTotal(color) - getSentTotal(color)`.
```
andrew-fleming, line 195: 👍

contracts/src/multisig/test/simulators/ShieldedMultiSigSimulator.ts
```diff
@@ -110,16 +107,19 @@ export class ShieldedMultiSigSimulator extends ShieldedMultiSigSimulatorBase {
     return this.circuits.impure.getProposal(id);
   }
 
-  public getProposalRecipient(id: bigint): Promise<Recipient> {
-    return this.circuits.impure.getProposalRecipient(id);
+  // getProposalRecipient / getProposalAmount / getProposalColor were dropped from
+  // the contract (redundant with getProposal; removed to fit the deploy block
+  // limit). Derive them here so specs are unchanged.
+  public async getProposalRecipient(id: bigint): Promise<Recipient> {
+    return (await this.getProposal(id)).to;
```
andrew-fleming: I'm a little uneasy about this bc it's dishonest. I think the these particular circuits should just be ledger reads in the tests. Furthermore, we should remove the getters from the multisig contract since it's undeployable as of now

contracts/src/multisig/test/simulators/ShieldedMultiSigSimulator.ts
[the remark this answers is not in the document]
andrew-fleming: Similar to my previous comment, I'd write this out in the test file (or maybe create a helper there or something). We shouldn't wrap logic like this as if it's a circuit in the module

contracts/src/multisig/test/ShieldedTreasury.test.ts
```diff
@@ -146,93 +171,161 @@ describe('ShieldedTreasury', () => {
...
-      expect(outputs[0].coinInfo.value).toBe(400n);
-      expect(outputs[0].coinInfo).toStrictEqual(result.sent);
+  describe('accounting consistency', () => {
+    beforeEach(async () => {
+      treasury = await freshTreasury();
     });
-  });
 
-  describe('accounting consistency', () => {
     it('should keep receivedMinusSent equal to balance', async () => {
```
andrew-fleming, line 329: If we agree on not creating a wrapper for a non-circuit, we can remove this section

contracts/test-utils/harness/NativeShieldedTokenTracker.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  /**
+   * Syncs (retrying to absorb indexer lag) until `coin`'s output is indexed, then
+   * returns it with its `mt_index`. Call after the transaction that created the
+   * coin. Throws if it never appears (never created on-chain, or wrong fields).
+   */
+  async resolve(
+    owner: ShieldedOwner,
+    coin: EncodedShieldedCoinInfo,
+    {
```
andrew-fleming, line 91: It's probably worth testing happy resolved cases and that it indeed fails if the coin doesn't appear

contracts/test-utils/harness/NativeShieldedTokenTracker.ts
```diff
@@ -0,0 +1,150 @@
...
+  private commitmentFor(
+    owner: ShieldedOwner,
+    coin: EncodedShieldedCoinInfo,
+  ): string {
+    const runtimeCoin = decodeShieldedCoinInfo(coin);
+    return owner.kind === 'contract'
+      ? ZswapOutput.newContractOwned(runtimeCoin, undefined, owner.address)
+          .commitment
+      : coinCommitment(runtimeCoin, owner.coinPublicKey);
+  }
```
andrew-fleming, line 81: Worth testing, no?

contracts/test-utils/harness/NativeShieldedTokenTracker.ts
```diff
@@ -0,0 +1,150 @@
...
+   * Pulls new ledger events up to the current head into the index. Re-reads the
+   * anchor block (idempotent) so a coin inserted in the same block the tracker
+   * was created in is never missed.
+   */
+  private async sync(): Promise<void> {
+    const head = await indexerHead(this.url);
+    if (head < this.lastHeight) return;
+    for (const event of await fetchCoinEvents(
+      this.url,
+      this.lastHeight,
```
andrew-fleming, line 60: Also worth testing

contracts/test-utils/harness/NativeShieldedTokenTracker.ts
```diff
@@ -0,0 +1,150 @@
...
+ */
+export function getShieldedCoinTracker(): Promise<NativeShieldedTokenTracker> {
+  if (!trackerPromise) {
+    const port = process.env.MIDNIGHT_INDEXER_PORT ?? '8088';
+    trackerPromise = NativeShieldedTokenTracker.create(
+      `http://127.0.0.1:${port}/api/v4/graphql`,
+    );
+  }
+  return trackerPromise;
+}
```
andrew-fleming, line 126: Also worth testing to improve test integrity

contracts/test-utils/harness/network.ts
```diff
@@ -0,0 +1,29 @@
+import { LocalTestConfiguration } from '@midnight-ntwrk/testkit-js';
+
+/** Parse a port override, failing with the responsible env var name so a
+ * malformed value (empty, non-numeric, out of range) is caught at load time
+ * instead of surfacing later as a bad `NaN`/`0` URL. */
```
andrew-fleming, line 5: **followup**: Can we confirm the comment's claim? Only defaults and happy overrides are covered by the tests

contracts/test-utils/harness/test/WalletPool.test.ts
```diff
@@ -0,0 +1,301 @@
...
+  for (const alias of Object.keys(SEEDS)) delete process.env[coinPkEnv(alias)];
+});
+
+describe('WalletPool', () => {
+  describe('ensureReady', () => {
+    it('should build every seed once and publish each coin public key', async () => {
+      const { build, built } = recordingBuilder();
+      await new WalletPool(SEEDS, build).ensureReady();
+
+      expect([...built].sort()).toEqual(['SIGNER1', 'SIGNER2', 'deployer']);
```
andrew-fleming: Correct me if I'm wrong, but doesn't `sort` hide the difference in ordering? Shouldn't the test prove that these were built one at a time? I imagine that there should be some sort of overlap detection

contracts/package.json
```diff
@@ -21,24 +21,23 @@
   "homepage": "https://docs.openzeppelin.com/contracts-compact/",
   "type": "module",
   "imports": {
-    "#test-utils/address.js": "./test-utils/address.js",
-    "#test-utils/zswap.js": "./test-utils/zswap.js"
+    "#test-utils/*": "./test-utils/*"
   },
   "scripts": {
-    "compact": "compact-compiler --exclude '*/archive/*'",
```
andrew-fleming: Why are we deleting this?

CONTRIBUTING.md
```diff
@@ -152,6 +157,76 @@ A maintainer will re-run the status check for you. If we conclude that the failu
...
+Scope the same mechanism to one category, or to a single file within it. The
+first argument names the category; any further argument is a filename substring
+(vitest matches it), so pass a spec name to run just that file on the live
+backend — the fast loop while iterating on one feature, instead of waiting for
+the whole category:
+
+```bash
+yarn test:live multisig                  # the whole category
+yarn test:live multisig ShieldedTreasury # just that one file
+yarn test:live multisig Forwarder        # any file matching "Forwarder"
```
andrew-fleming: I think we should find a better way to do this. What happens when I want to run `yarn test:live token NativeShieldedToken`? Wouldn't this also run `NativeShieldedTokenFamily` and `NSTCore`?

andrew-fleming: Or the path of least resistance: Don't advertise that you can run just the file name. Just "any file matching that name"

Makefile
```diff
@@ -6,7 +6,10 @@ SERVICES := proof-server indexer node
 
 ## Start local environment and stream logs to logs/
 env-up: env-down
-	docker compose -f $(COMPOSE_FILE) up -d --wait
+	docker compose -f $(COMPOSE_FILE) up -d
+	@# proof-server has no healthcheck, so a blanket `--wait` is not portable
+	@# across Docker Compose versions; wait only on the services that expose one.
+	docker compose -f $(COMPOSE_FILE) up -d --wait node indexer
```
andrew-fleming, line 12: 👍

turbo.json
```diff
@@ -58,21 +66,33 @@
         "compact:access",
         "compact:multisig",
         "compact:token"
-      ],
-      "env": ["COMPACT_HOME", "SKIP_ZK"],
-      "inputs": ["src/**/*.compact", "test/**/*.compact"],
-      "outputLogs": "new-only",
-      "outputs": ["artifacts/**"]
```
andrew-fleming, line 65: Why are these removed from the `compact` task?

andrew-fleming (commented):
Looking good, @0xisk! Thank you for moving out the unrelated changes. I left some comments and questions

contracts/src/multisig/test/simulators/ShieldedMultiSigSimulator.ts
andrew-fleming: If you agree with removing the getters from the preset, let's open an issue so we can focus on the test harness

contracts/src/multisig/test/ShieldedMultiSigV3.test.ts
0xisk: Done in 0283ba16. Grouped the happy-path (dry-only) burn tests under one `describe.skipIf(isLiveBackend())('happy path (dry only)')`, matching the describe split in `ShieldedTreasuryStateless`. The reject tests stay backend-agnostic as direct children; mint keeps its lone `skipIf` (a single test needs no wrapper). (The file is `ShieldedMultiSigV3.test.ts` on this branch now — the preset rename moved to #682.)

contracts/src/multisig/test/ShieldedTreasuryStateless.test.ts
0xisk: Could you elaborate more? Because I this is already in place here: `ShieldedTreasuryStateless.test.ts` uses `describe.skipIf(isLiveBackend())('… (dry only)')` paired with `describe.runIf(isLiveBackend())('… on live')`, with the dry/live split right in the describe labels.

contracts/test-utils/fixtures/address.ts
0xisk: Agreed. Since it spans ~10 files across access/token/utils/multisig plus the fixtures, I moved it to #688, stacked on this branch (it depends on `eitherUserFromCoinPublicKey`, which this PR introduces). The two helpers were byte-identical, so every call site becomes `eitherUserFromCoinPublicKey(toHexPadded(label))` with no change to test values. The `archive` package keeps its own separate copy.

contracts/src/multisig/test/simulators/ShieldedMultiSigSimulator.ts
0xisk: Agreed. Removed the faked getters from the preset simulator (`getProposalRecipient/Amount/Color`) in 783595be; the preset tests now read those fields off `getProposal` directly. The contract-side getter trim (making the preset deployable, and dropping `getReceivedMinusSent` as a circuit) is tracked in #689 so this PR stays focused on the harness.

contracts/src/multisig/test/simulators/ShieldedMultiSigSimulator.ts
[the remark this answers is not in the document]
0xisk: Same here: dropped the `getReceivedMinusSent` wrapper, and the tests now derive `getReceivedTotal - getSentTotal` inline. Removing it as a circuit is part of #689.

contracts/src/multisig/test/ShieldedTreasury.test.ts
0xisk: The `ShieldedTreasury` module still exposes `getReceivedMinusSent` as a real circuit (only the preset dropped it), so this section exercises an actual circuit, not a wrapper. Kept it for now; it's removed together with the module circuit in #689.

contracts/test-utils/harness/NativeShieldedTokenTracker.ts
0xisk: Added in d235550c: `resolve` happy path for a wallet- and a contract-owned coin (returns the coin with its `mt_index`), plus the not-found case (throws, naming the synced height). Only the indexer event stream is mocked; the commitment math runs for real.

contracts/test-utils/harness/NativeShieldedTokenTracker.ts
0xisk: Covered by the resolve tests in d235550c: both owner arms (wallet via `coinCommitment`, contract via `ZswapOutput.newContractOwned`) are exercised end-to-end with the real commitment computation; only `ledgerEvents` is mocked.

contracts/test-utils/harness/NativeShieldedTokenTracker.ts
0xisk: Added an anchor-block test in d235550c: it asserts the first sync re-reads the creation height (`from === to === head`), so a coin inserted in the block the tracker was created in is not missed.

contracts/test-utils/harness/NativeShieldedTokenTracker.ts
0xisk: Added a memoization test in d235550c: `getShieldedCoinTracker()` returns the same process-wide promise across calls.

contracts/test-utils/harness/network.ts
0xisk: Good catch, the guard's claim had no test. Added the three throw cases in 48d84b8a: empty, non-numeric, and out-of-range overrides each throw at load time naming the responsible env var.

contracts/test-utils/harness/test/WalletPool.test.ts
0xisk: You're right, `sort` hid the ordering. Replaced it in 0f90aa37 with an exact-order assertion plus a max-concurrency check (a counting builder proves no two builds ever overlap). Also fixed the class doc that wrongly said "concurrently": the build is serial on purpose, since empty signers are funded from the shared deployer in turn.

CONTRIBUTING.md
0xisk: Took the path of least resistance in 0492f4fa: the doc now states the argument is a filename substring, so a name that prefixes others runs them all (with `NativeShieldedToken` as the example). It no longer advertises a single-file run.

contracts/package.json
0xisk: `compact` is now a pure turbo aggregator task: it `dependsOn` the per-dir `compact:*` tasks and runs no command of its own, so the top-level `compact` script (which re-compiled the whole tree in one pass) is redundant. `turbo run compact` still fans out to all six per-dir compiles, verified via `--dry`. See the related note on `turbo.json`.

turbo.json
0xisk: Same refactor: since `compact` now only orchestrates (no command), each per-dir `compact:*` task carries its own `env`/`inputs`/`outputs`, so the aggregator declaring them again was redundant and would misreport its fingerprint (it produces no artifacts itself). `turbo run compact --dry` confirms it still schedules all six per-dir compiles.

0xisk:
Follow-ups split out of this PR:

**Stacked on this PR** (retarget to `main` automatically once it merges):
- #681 — live test suite workflow (draft)
- #682 — rename presets by capability
- #688 — remove `createEitherTestUser`

**Independent** (base `main`):
- #680 — rename dev scripts to `compile`/`lint`

**Tracking issue:** #689 — trim the `ShieldedMultiSig` preset getters to a deployable set.

contracts/src/multisig/test/ShieldedTreasury.test.ts
andrew-fleming: Ah my bad, sounds good

CONTRIBUTING.md
```diff
@@ -152,6 +157,77 @@ A maintainer will re-run the status check for you. If we conclude that the failu
...
+argument names the category; any further argument is a filename substring
+(vitest matches it), so pass a spec name to run every file whose name matches it
+on the live backend — the fast loop while iterating on one feature, instead of
+waiting for the whole category. The match is a substring, not an exact file, so
+a name that prefixes others runs all of them:
+
+```bash
+yarn test:live multisig                  # the whole category
+yarn test:live multisig ShieldedTreasury # any file matching "ShieldedTreasury"
+yarn test:live token NativeShieldedToken # every "NativeShieldedToken*" spec
```
andrew-fleming:
```suggestion
yarn test:live multisig ShieldedTreasury # any file matching "ShieldedTreasury"
```
**nit**: Forgive me for being annoying, they're saying the same thing, no? Why not just keep the one?

andrew-fleming (commented):
Very nice, @0xisk! You mind resolving conflicts? Then we should be good to go

andrew-fleming (commented):
@0xisk just a heads up, there are still conflicts. And I assume the extra commits after the last review are from force pushing something? The conflicts were (and are) in yarn.lock

0xisk:
> @0xisk just a heads up, there are still conflicts. And I assume the extra commits after the last review are from force pushing something? The conflicts were (and are) in yarn.lock

@andrew-fleming I fixed the conflicts with rebase and force push with the main.

andrew-fleming (approved):
LGTM!

PR #676, opened 2026-07-14
fix(multisig): reject non-adjacent duplicate signers in EcdsaSignerManager

### LINKED ISSUE #629: [Bug]: EcdsaSignerManager.verify only detects adjacent duplicate signers (incorrect for 3+ approvals)
### Prerequisites

- [x] I have read the Contributing Guidelines.
- [x] I agree to follow the Code of Conduct.
- [x] I have searched for existing issues that already report this problem, without success.

### What are the steps to reproduce this issue?

1. Use `EcdsaSignerManager` (from #628) with a registry of 3+ signers and a threshold of 3.
2. Call `verify<n>` with `n >= 3`, presenting a commitment vector that contains a **non-adjacent** duplicate, e.g. `[A, B, A]`.
3. Observe that verification passes with `validCount == 3`.

### What happens?

`verifySignature` rejects duplicates by comparing each commitment only against the **immediately previous** one:

```
assert(commitment != state.prevCommitment, "EcdsaSignerManager: duplicate signer");
```

This catches adjacent duplicates only. With 3+ presented signers, a non-adjacent duplicate such as `[A, B, A]` passes every check, so the same registered signer is counted more than once toward the threshold — a forged quorum (e.g. "3 approvals" from only 2 distinct signers).

The module already documents this in its notice ("sufficient for at most 2 signers"), and every current call site uses `verify<2>` (where the only possible duplicate is adjacent), so it is **not exploitable in shipped usage today**. It becomes a live correctness/authorization bug the moment `verify` is called with 3+ presented signers.

### What were you expecting to happen?

`verify` should count only **distinct** registered signers toward the threshold, for any `n`. A duplicate at any position — adjacent or not — must be rejected.

### Paste any relevant logs, error output, etc.

```shell
need 3 approvals; present commitments [A, B, A]
  step 1: A vs (zero) -> different  OK  count=1
  step 2: B vs A      -> different  OK  count=2
  step 3: A vs B      -> different  OK  count=3   <-- A was already counted in step 1
result: validCount = 3 >= threshold 3  -> PASS
        (should FAIL: only 2 distinct signers)
```

### Additional context

- Source: `contracts/src/multisig/EcdsaSignerManager.compact`, `verifySignature` (refactored in #628).
- Fix options noted in the module itself: require **strictly-increasing (sorted) commitments** so a strict total order forbids any repeat anywhere, or use a **bitmap** of registered-signer indices. The sorted approach keeps O(1) fold state but requires callers to present `pubkeys`/`signatures` sorted by ascending commitment, plus a `Bytes<32>` total-order helper (Compact has no native ordering on `Bytes<32>`: `Field` has no relational operator and `Uint` caps at 248 bits, so a byte-wise lexicographic comparison is needed).
- Scope: latent — affects any consumer that presents 3+ approvals. Filing to track for a later fix; not blocking current 2-of-N usage.

### Version

0.31.0

Yonkoo11:
## Types of changes

Fixes #629

## Summary

`EcdsaSignerManager.verify` only rejected **adjacent** duplicate commitments:

```compact
assert(commitment != state.prevCommitment, "EcdsaSignerManager: duplicate signer");
```

For `n >= 3`, a presentation like `[A, B, A]` passed every check and counted the same signer twice toward the threshold (forged quorum). The module already documented that the adjacent check was only sufficient for at most 2 signers; this PR implements the sorted / strictly-increasing fix noted in #629 and in that notice.

### Approach

Require presented commitments to be **strictly increasing** under Compact's `Bytes`→integer embedding (first byte is LSB — the same order as `Bytes as Field`):

1. Zero sentinel as `prevCommitment` for the first iteration.
2. Each next commitment must be strictly greater than the previous.
3. Equals (adjacent or not) fail with `EcdsaSignerManager: duplicate or unsorted signer`.

**Why not `as Field as Uint<248>`?** Full 32-byte hash commitments often exceed Compact's max `Uint` (`2^248 - 1`). That cast compiles but panics at runtime. Comparison uses two 16-byte little-endian limbs as `Uint<128>` (always in range):

```compact
(aHigh > bHigh) || ((aHigh == bHigh) && (aLow > bLow))
```

### Call-site impact (why "breaking" is checked)

- **Semantics for distinct unsorted pairs change:** for `n >= 2`, callers must present `pubkeys`/`signatures` sorted by ascending commitment (first-byte-LSB order). Previously any order of two distinct keys worked.
- **Current shipped call sites** on this branch use `verify<2>` only; they remain safe if they sort (or already present in ascending order).
- Latent bug for any `verify` with `n >= 3` is fixed.

### Tests

- Mock exposes `verify3` for a 3-of-N presentation path.
- New cases: three distinct sorted signers pass; `[A, B, A]` rejects; adjacent duplicate in a triple rejects.
- Existing adjacent-duplicate / membership tests updated to sort when presenting distinct keys.
- Verified locally: `SKIP_ZK=true yarn vitest run src/multisig/test/EcdsaSignerManager.test.ts` → **11/11 passed**.

## PR Checklist

#### Further comments

**Base branch:** `refactor/multisig-signature-verifier` (#628). `EcdsaSignerManager` is not on `main`; targeting `main` would be incorrect.

**Alternatives considered:** bitmap of registered-signer indices (also noted in #629) — stronger without a sort requirement, but needs more fold state / registry coupling. Strict order keeps O(1) fold state and matches the module's documented direction.

### DIFF
--- a/contracts/src/multisig/EcdsaSignerManager.compact
+++ b/contracts/src/multisig/EcdsaSignerManager.compact
@@ -31,9 +31,17 @@ pragma language_version >= 0.23.0;
  * true). Replace it with `ecdsaVerify` once the Compact ECDSA primitive is
  * available.
  *
- * @notice Duplicate detection compares each commitment against the previous one
- * only, which is sufficient for at most 2 signers. Larger signer sets need a
- * different uniqueness mechanism (sorted commitments or a bitmap).
+ * @notice Duplicate detection requires that presented signer commitments form a
+ * **strictly increasing** sequence under Compact's `Bytes`→integer order
+ * (first byte is the least-significant byte — the same embedding as
+ * `Bytes as Field`). Equal commitments therefore fail, including non-adjacent
+ * duplicates such as `[A, B, A]`. Callers that present `n >= 2` public
+ * keys/signatures must sort them by ascending commitment under that order
+ * before calling `verify`.
+ *
+ * Order is total on 32-byte values. Collision of distinct hash commitments is
+ * negligible. Comparison is implemented as two `Uint<128>` limbs (bytes 0–15
+ * and 16–31) so no value overflows Compact's 248-bit `Uint` cap.
  */
 module EcdsaSignerManager {
   import CompactStandardLibrary;
@@ -43,8 +51,8 @@ module EcdsaSignerManager {
 
   /**
    * @description Accumulator for fold-based signature verification. Threads the
-   * valid count, previous commitment (for duplicate detection), and message hash
-   * through each iteration.
+   * valid count, previous commitment (strictly-increasing uniqueness for any
+   * `n`), and message hash through each iteration.
    */
   export struct VerificationState {
     validCount: Uint<8>,
@@ -95,19 +103,22 @@ module EcdsaSignerManager {
    * for duplicates and registry membership, and its signature validated against
    * `msgHash`; the valid count is then checked against the threshold.
    *
-   * @notice Duplicate detection is correct for at most 2 signers (see module
-   * notice).
+   * @notice Presented commitments must be **strictly increasing** under Compact
+   * `Bytes`→integer order (first byte LSB; see module notice). Sort
+   * `pubkeys`/`signatures` by ascending commitment before calling when `n >= 2`.
    *
    * Requirements:
    *
    * - Every public key must hash to a registered signer commitment.
    * - Every signature must be valid over `msgHash`.
-   * - Signers must not be duplicates.
+   * - Signers must be distinct: each commitment must be strictly greater than
+   *   the previous one in presentation order (total order on `Bytes<32>`).
    * - Valid count must meet the threshold.
    *
    * @param {Bytes<32>} msgHash - The message hash signers signed off-chain.
-   * @param {Vector<n, Bytes<64>>} pubkeys - ECDSA public keys of approving signers.
-   * @param {Vector<n, Bytes<64>>} signatures - Signatures over `msgHash`.
+   * @param {Vector<n, Bytes<64>>} pubkeys - ECDSA public keys of approving signers
+   *   (ordered by ascending commitment; first byte LSB).
+   * @param {Vector<n, Bytes<64>>} signatures - Signatures over `msgHash` (same order).
    * @returns {[]} Empty tuple.
    */
   export circuit verify<#n>(
@@ -117,6 +128,9 @@ module EcdsaSignerManager {
   ): [] {
     const initialState = VerificationState {
       validCount: 0 as Uint<8>,
+      // Zero sentinel so the first commitment need only be non-zero (hash
+      // commitments are effectively never zero) and later ones must strictly
+      // increase.
       prevCommitment: pad(32, ""),
       msgHash: msgHash
     };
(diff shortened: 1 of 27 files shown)


0xisk:
🔵 **followup:** No `CHANGELOG.md` entry in the PR. The repo uses Keep-a-Changelog (`## Unreleased` → `### Fixed`); add an entry for the #629 fix, or confirm it is intentionally folded into the #628 line since `EcdsaSignerManager` is not on `main` yet.

contracts/src/multisig/EcdsaSignerManager.compact
```diff
@@ -217,4 +236,30 @@ module EcdsaSignerManager {
...
+   * @returns {Boolean} True if a > b.
+   */
+  pure circuit isStrictlyGreaterCommitment(a: Bytes<32>, b: Bytes<32>): Boolean {
+    // Bytes→Uint: first byte of the slice is the least-significant byte.
+    // 16 bytes always fit in Uint<128> (max 2^128-1).
+    const aLow = slice<16>(a, 0) as Uint<128>;
+    const aHigh = slice<16>(a, 16) as Uint<128>;
+    const bLow = slice<16>(b, 0) as Uint<128>;
+    const bHigh = slice<16>(b, 16) as Uint<128>;
+    return (aHigh > bHigh) || ((aHigh == bHigh) && (aLow > bLow));
```
0xisk, line 262: 🔵 **followup:** The low-limb branch (`aHigh == bHigh && aLow > bLow`) and the little-endian `Bytes<16> as Uint<128>` assumption it rests on are not exercised — all test commitments are `persistentHash` outputs whose high limbs differ, so this branch never runs. Distinctness (the #629 fix) holds regardless of endianness, but if the cast were not LSB-first the shipped `sortByCommitmentField` would mis-order callers and reject honest `n>=3` quorums (liveness, not a security break). Add a unit test on crafted 32-byte values: a pair equal in bytes 16–31 differing only in bytes 0–15, and a pair crossing the limb boundary; assert the TS sort order matches on-chain accept/reject.

contracts/src/multisig/test/EcdsaSignerManager.test.ts
```diff
@@ -69,20 +117,42 @@ describe('EcdsaSignerManager', () => {
...
       await expect(
         verifier.verify(MSG, [PK1, PK1], [SIG, SIG]),
-      ).rejects.toThrow('EcdsaSignerManager: duplicate signer');
+      ).rejects.toThrow('EcdsaSignerManager: duplicate or unsorted signer');
     });
 
     it('should reject an unregistered signer', async () => {
+      const { pubkeys, signatures } = sortedPair(PK_UNKNOWN, PK2);
+      // Order may place unknown first or second; either fails membership or order
+      await expect(verifier.verify(MSG, pubkeys, signatures)).rejects.toThrow();
```
0xisk, line 134: ⚪ **nitpick:** This lost its specific-message assertion because sort order decides whether the unknown key fails on membership or on ordering. Make it deterministic — present a registered key that sorts *before* the unknown one — so you can assert `'SignerManager: not a signer'` again instead of "any throw".

contracts/src/multisig/EcdsaSignerManager.compact
```diff
@@ -217,4 +236,30 @@ module EcdsaSignerManager {
...
+  }
 }
```
0xisk, line 264: ⚪ **nitpick (if-minor):** Stray trailing blank line after the module closing brace; `compact fmt` / biome will likely flag it.

0xisk (commented):
Thank you @Yonkoo11 LGTM! Howver, I'm sorry you will need to sign your commit. Also for the followups, nitpicks you can ignore them. We can take care of that.

Yonkoo11:
Thanks @0xisk — commits are now signed (SSH). Ready for re-review.

PR #682, opened 2026-07-15
refactor(multisig): rename presets by capability

0xisk:
## Types of changes

Split out of #673 per review (comments on the preset `.compact` rename and the
`git mv` diff): the multisig preset rename is a user-facing contract change, so
it lands as its own clean, reviewable PR with a changelog entry instead of being
buried in the harness diff.

Renames the shielded multisig presets to describe their capability:

| Before | After |
| --- | --- |
| `ShieldedMultiSig` | `ShieldedProposalMultisig` (on-chain proposal lifecycle) |
| `ShieldedMultiSigV2` | `ShieldedStatelessMultisig` (off-chain signatures, stateless treasury) |
| `ShieldedMultiSigV3` | `NativeShieldedTokenMultisig` (threshold mint/burn native token) |

Presets, tests, and simulators are renamed together, so every file shows as a
clean git rename (95-99% for the `.compact` presets and their tests). Every
identifier, artifact path, and reference is updated.

**Stacked PR.** Based on `test/multisig-live-harness` (#673), so the diff here is
only the rename. Once #673 merges to `main`, this retargets to `main`
automatically.

## PR Checklist

- [x] I have added tests that prove my fix is effective or that my feature works. Existing suites move with the rename; compile + type-check pass.


0xisk:
outdated

Issue #716, opened 2026-07-27
dev: live-harness follow-ups (node-rejection constants, orchestrator tests, env-down pkill)


0xisk:
Follow-ups found while building the `integration-live` harness (branch
`test/integration-live-harness`). None of these block that branch; they were
deliberately left out to keep it scoped. Design notes live in
`docs/design/integration-live-harness*.md`.

Filed from the **Codebase improvement** template.

## 1. Centralize the node-rejection strings

Node error strings are hardcoded in 8 files, in three different roles:

| Role | Where |
|---|---|
| Live assertion | `contracts/test/integration/specs/confidentialFungibleToken.spec.ts` (exact phrase) |
| Live assertion | `contracts/src/token/test/ConfidentialFungibleToken.test.ts:1188` (loose `/block limits\|exhaust the block/i`) |
| Retry logic | `contracts/test-utils/harness/LiveSimulatorBackend.ts:170` — `DETERMINISTIC_REJECTION = /1010: Invalid Transaction/` |
| Prose only | 6 doc comments in `live.globalSetup.ts`, `scripts/live/LiveOrchestrator.ts`, `test-utils/fixtures/nativeShieldedToken.ts` |

The retry logic and the canary now encode **different** notions of the same
rejection: the canary demands the full phrase, `isDeterministicRejection` matches
only the `1010: Invalid Transaction` prefix. That is how these drift apart.

Proposed `contracts/test-utils/harness/nodeRejections.ts`:

```ts
/** Verified against ledger-v8 8.1.0 / midnight-js 4.1.1. These are the NODE's
 *  words, not ours — a ledger bump can reword them. */
export const VERIFIED_AGAINST = { '@midnight-ntwrk/ledger-v8': '8.1.0' } as const;

export const RPC_INVALID_TRANSACTION = '1010: Invalid Transaction';

export const NODE_REJECTION = {
  /** Tx IR exceeds the per-tx block byte budget. */
  blockLimitExceeded: `${RPC_INVALID_TRANSACTION}: Transaction would exhaust the block limits`,
  /** Coin spent against stale node state (re-spend / stale UTXO). */
  staleCoinState: `${RPC_INVALID_TRANSACTION}: Custom error: 103`,
} as const;
```

On tying this to the ledger version: a `Record<ledgerVersion, ...>` map would be
speculative, since only one ledger version is pinned at a time and the unused
branches would rot. Prefer a `VERIFIED_AGAINST` constant plus a harness unit test
[...]

Issue #693, opened 2026-07-17
dev: run the supply property/fuzz spec on the live harness


0xisk:
Parent: #665
Depends on the core harness (#666).

## Scope

The token supply property/fuzz spec (`NativeShieldedTokenSupply.property.test.ts`) is currently dry-only (`describe.skipIf(isLiveBackend())`), left out of the live run on a **feasibility** ground, not a coverage one:

- fast-check drives 15 randomized runs of up to 6–8 mint/burn ops each, and every op is its own deploy/tx on the live backend (~18s apiece), so the file would run for hours; and
- the random op sequences are non-deterministic, so the two-round flake classifier in `scripts/test-live.ts` (re-run a failed file to tell a real failure from an environment flake) cannot reproduce a failing case.

The deterministic scalar/family supply specs already exercise the same `NativeShieldedTokenSupplyCore` accounting live, so the invariant is covered on the node. This issue tracks the optional extra of fuzzing it live.

- [ ] Pin a fixed fast-check `seed` so runs are reproducible (the flake classifier needs determinism).
- [ ] Reduce `numRuns` on live (env-gated) to keep the job within its budget.
- [ ] Un-gate the spec and confirm green on `make env-up`.

Introduced alongside the token/access/security/crypto/utils live rollout.

0xisk:
Introduced alongside #694 (the live-harness rollout for the remaining categories); the fuzz spec is gated `skipIf(isLiveBackend())` there.

PR #717, opened 2026-07-27
test(integration): drive the integration specs against the live stack

### LINKED ISSUE #718: dev: run integration specs on the live harness
Parent: #665
Depends on the core harness (#666).

## Scope

Run the composed-contract specs in `test/integration/specs` against the live
harness, via a new `integration-live` vitest project that reuses the same
globalSetup, wallet pool, backend, and orchestrator as `unit-live`.

Unlike the per-category unit children, the composed contract **cannot deploy
live**: the base `ConfidentialFungibleToken` already bundles four k=16 circuits'
IR into one deploy tx and overruns the per-tx block byte budget, and this
composition is strictly larger. The deliverable is therefore the capability plus
a **block-limit canary** that asserts the rejection, rather than green functional
coverage. The canary flips red the day a smaller composition or a staged deploy
lets the contract through, which is precisely when functional live coverage
becomes possible.

The functional flows are also blocked independently: they mutate private state
mid-test via `switchIdentity` / `cachePlaintext`, which the live backend throws
on (`private-state mutation unsupported on live backend`). Going green needs both
a deploy that fits and a deploy-seeded, memo-decrypt rewrite.

- [x] `integration-live` vitest project, reusing `unit-live`'s globalSetup and setup
- [x] `test:live integration` orchestrator target, emitted by `--list` so CI can
      spawn a job for it
- [x] Full-key mock compile — `compile:integration` takes `SKIP_ZK` from the
      environment so one script serves the dry and live paths
- [x] Truncated-key scan widened to cover the integration mocks
- [x] Block-limit canary asserting the node's exact rejection, verified on a local
      stack: `1010: Invalid Transaction: Transaction would exhaust the block limits`
- [x] Backend guards on both integration specs
- [ ] Green functional live coverage — blocked by the block limit. Needs a smaller
      composed mock or a staged-deploy / CMA path, each its own design track
- [ ] Wallet-partitioned parallel `integration-live` — only needed once multiple
      green live specs exist; single worker is sufficient for one rejected deploy
- [ ] CI gating parity for the `integration-live` matrix job, which needs the live
      workflow in #681 to merge first

Related: #716 tracks the follow-ups left out of the implementing PR.

Also noted while implementing: the dry `integration` project is currently red on
`main`, because `test/integration/fixtures/*` are written against an older
`@openzeppelin/compact-simulator` API — the installed 0.2.0 constructor takes
`(deps)`, so `new ComposedTokensSimulator(args, {})` hits `deps.backend.kind` on
undefined. That is adjacent to #712 but not listed there.

0xisk:
## Types of changes

Fixes #718

Parent: #665 (live-backend test harness umbrella). Sibling of the per-category
children #668–#672 and #693, for the `test/integration/specs` tree.
Refs #716, which tracks the follow-ups deliberately left out.

Lets the existing live backend drive the `integration` vitest project, the same way
it already drives `unit-live`. The backend is already backend-generic, so this is
additive wiring rather than new harness logic.

**The deliverable is the capability plus a block-limit canary, not green functional
live coverage.** The composed contract cannot deploy live: the base token already
bundles four k=16 circuits' IR into one deploy tx and overruns the per-tx block byte
budget, and this composition is strictly larger. Rather than skip and hide that, the
canary asserts the rejection, so it fails loudly the day a staged deploy or a smaller
composition lets it through.

### What's here

| Change | Why |
|---|---|
| `integration-live` vitest project + `test:live integration` target | The capability. Reuses `unit-live`'s globalSetup and setup verbatim |
| Block-limit canary in `confidentialFungibleToken.spec.ts` | Proves the "cannot deploy live" scope claim instead of assuming it |
| `scripts/live/` — orchestrator split into 8 services | `test-live.ts` had reached 478 lines and eleven responsibilities |
| Stack teardown on every exit path | Containers and log streamers previously outlived every run |
| One-live-project-per-run guard | Two live projects would spend the same genesis deployer's coins |
| Composed CFT mock renamed | Two contracts shared one artifact directory; last compile won |
| `compile` turbo inputs scoped | Any `.ts` edit forced a full 70-contract keygen |
| `scripts` vitest project + tests | The orchestrator had no coverage at all |

Dry behaviour is unchanged: `--project integration` produces the same pass set as
`main`, plus the canary skipped.

### Verified

| Suite | Result |
|---|---|
| `unit` (CI-gated) | 44 files, 1483 passed / 10 skipped |
| `harness` | 11 files, 110 passed / 4 skipped |
| `scripts` (new) | 14 passed |
| `integration` (dry) | 7 passed / 1 skipped |
| `integration-live` | canary green, `VERDICT: PASSED` |

Live-verified on a local stack: the canary, the exact node rejection, the
one-live-project guard, teardown on both the normal and signal paths, and the mock
rename resolving through `artifactName` for a real deploy.

The canary asserts the node's exact wording, confirmed on-stack:

```
1010: Invalid Transaction: Transaction would exhaust the block limits
```

A loose `/block limits/` match could be satisfied by an unrelated deploy failure
(funding, proving, a submission bounce) and report a false green, which would
silently void the scope claim the canary exists to prove.

## PR Checklist

- [x] CI Workflows Are Passing

#### Further comments

**The `integration-live` job will not run in CI until #681 merges.** That PR adds the
workflow which reads `test:live --list`; this branch only makes `--list` emit
`integration`. So green CI here has not exercised the live path. It was verified
locally instead.

**Two fixes here predate this branch** and are self-contained commits, splittable if
you'd rather keep the PR narrow:

- `02533780` — the `compile` turbo task declared only `dependsOn`, so turbo hashed
  every tracked file in the package. The six per-category tasks were cache hits while
  the top-level task rebuilt all 70 contracts anyway. Resolved input set drops from
  214 files to 71, measured with `--dry=json`.
- `f453342d` — `ConfidentialFungibleTokenPublicSupply.compact` existed under both
  `src/token/extensions` and `test/integration/_mocks`, and the compiler names each
  artifact directory after the source basename. It surfaced as a wrong-contract arity
  error, invisible to lint, typecheck, and the truncated-key scan.

**Why the orchestrator was split rather than extended.** Adding the integration target
to a 478-line file carrying eleven concerns would have made it worse. The split is
behaviour-preserving: every console string, all exit codes, and the order of
operations are unchanged. It also made the two silent-failure paths testable, which is
where the new tests went. The spawn wrappers are deliberately untested, since
asserting the arguments passed to `spawnSync` restates the implementation.

**One interrupt bug worth flagging.** Ctrl-C during a compile used to be ignored:
`spawnSync` blocked the event loop for a whole compile phase, so the SIGINT handler
could not run, the interrupted compile looked successful because turbo exits 0 on its
own graceful shutdown, and the truncated keys the interruption had just created were
then read as a poisoned cache — draining it and starting a pointless serial recompile.
`run` is now asynchronous, so the handler pre-empts that.

**Known limitation.** `unit-live` was not re-run on this branch. Two changes touch it
(the per-project worker total, and the reporter's worker tag for skipped tests), and
`yarn test:live multisig` is a long run. Worth one before merge if you want that path
covered.


contracts/package.json
```diff
@@ -32,7 +32,7 @@
     "compile:security": "compact-compiler --dir security",
     "compile:token": "compact-compiler --dir token",
     "compile:utils": "compact-compiler --dir utils",
-    "compile:integration": "SKIP_ZK=true compact-compiler --src test/integration/_mocks",
+    "compile:integration": "compact-compiler --src test/integration/_mocks",
```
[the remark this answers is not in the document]
0xisk: Not applied. `ArtifactCompiler.#compileAll` clears `SKIP_ZK` on purpose so a live run gets real proving keys; a hardcoded prefix in the script overrides that and would deploy keyless artifacts. The dry path still exports it from the root `test:integration`, and turbo keys the task on it so the two cache apart. Every sibling `compile:*` leaves it to the caller too. Intentional in a8e872ac.

scripts/live/RunLock.ts
```diff
@@ -0,0 +1,75 @@
...
+      if ((e as NodeJS.ErrnoException).code !== 'EEXIST') throw e;
+      const info = this.#read();
+      if (info && pidAlive(info.pid)) {
+        throw new Error(
+          `another test:live run is already in progress (pid ${info.pid}, ` +
+            `started ${info.startedAt}). Wait for it, or remove ${this.#path}.`,
+        );
+      }
+      writeFileSync(this.#path, stamp); // stale — reclaim
+    }
```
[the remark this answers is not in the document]
0xisk: Fixed in 2d7cb828, via `rename` rather than `unlink` + `wx`. The unlink version still races: A unlinks and creates, then B (holding the earlier stale read) unlinks A's fresh lock. POSIX `rename` is atomic, so exactly one process can claim a stale lock and the loser gets the "already in progress" error. Also narrowed the catch to ENOENT so a permissions fault on `logs/` reports itself. Adds the lock's first 6 tests.

scripts/live/VitestRunner.ts
```diff
@@ -0,0 +1,67 @@
...
+   *   (dirty node / lock) or crashed before writing one, which callers must treat
+   *   as an infrastructure abort rather than a test failure.
+   */
+  fileStatuses(reportPath: string): Map<string, string> | undefined {
+    if (!existsSync(reportPath)) return undefined;
+    const report = JSON.parse(readFileSync(reportPath, 'utf8')) as JsonReport;
+    return new Map(
+      (report.testResults ?? []).map((r) => [r.name, r.status] as const),
+    );
+  }
```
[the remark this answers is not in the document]
0xisk: Fixed in ce118341. Parsing now falls into the same no-result path and names the unreadable file, since the caller's "produced no results file" would otherwise misdescribe it. Adds 4 tests for the reader.

scripts/test-live.ts
```diff
@@ -1,478 +1,99 @@
...
+    cleanup('run finished');
   }
 }
 
 main()
   .then((code) => process.exit(code))
   .catch((e) => {
     console.log(e instanceof Error ? e.message : String(e));
-    process.exit(2);
+    process.exit(INFRA_ABORT);
```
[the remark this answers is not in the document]
0xisk: Fixed in d0885c8f. Verified piped: `--list` still exits 0 and a non-live-ready category exits 2, output intact.

contracts/test/integration/specs/confidentialFungibleToken.spec.ts
```diff
@@ -21,6 +22,14 @@ import { ConfidentialFungibleTokenPublicSupplySimulator } from '../fixtures/conf
  *
  * so `totalSupply` must move in lockstep with the confidential balance change.
  * Balances stay hidden; `totalSupply` (the public aggregate) is what we assert.
+ *
+ * Backend split (`--project integration` vs `integration-live`), in file order:
+ *   1. The LIVE block is the block-limit canary: it asserts the composed deploy
+ *      is rejected, which is what makes "no green functional live integration
+ *      coverage" a verified claim rather than an assumption.
```
andrew-fleming: Briefly discussed offline, but to reiterate: this is a current limitation with ledger v8 but, as of now, the canary will likely fail on the forthcoming stack (with ledger v9)

contracts/vitest.config.ts
```diff
@@ -125,6 +141,31 @@ export default defineConfig({
           include: ['test/integration/specs/**/*.spec.ts'],
         },
       },
+      {
+        // Same files as `integration`; `MIDNIGHT_BACKEND=live` (set by
+        // `test:live integration`) flips the `isLiveBackend()`-gated blocks —
+        // today the composed-deploy block-limit canary — on, and skips the dry
```
andrew-fleming: **nit**: This is both a grammatically correct and totally unreadable sentence haha. AI prose loves to just abuse the spaced em-dash. I'd rephrase this sentence

scripts/live/paths.ts
```diff
@@ -0,0 +1,53 @@
...
+);
+export const CONTRACTS = path.join(REPO_ROOT, 'contracts');
+export const SRC = path.join(CONTRACTS, 'src');
+export const ARTIFACTS = path.join(CONTRACTS, 'artifacts');
+/** Integration mocks compile into the same `artifacts/` tree as `src`, so the
+ * truncated-key scan has to be told about them explicitly. */
+export const INTEGRATION_MOCKS = path.join(
+  CONTRACTS,
+  'test/integration/_mocks',
+);
```
andrew-fleming, line 21: **followup** Wouldn't a better long-term approach be to compile integration artifacts into a separate `integration/` directory? This would be way simpler and would avoid the same-name issue with the cft mock

andrew-fleming (commented):
Looking good, @0xisk! I left a few comments. Also, I think we need to update CONTRIBUTING docs accordingly

contracts/test/integration/specs/confidentialFungibleToken.spec.ts
0xisk: Scoped to ledger v8 in `44009e0da6b17f58e06f83841fd8e151775ff21d`

contracts/vitest.config.ts
0xisk: Fair. Rephrased in `d14268df57c4b13e30b9fd4a42f855e87ec9069a`

scripts/live/paths.ts
0xisk: Agreed, and opened this issue for it https://github.com/OpenZeppelin/compact-contracts/issues/758

CONTRIBUTING.md
```diff
@@ -212,12 +229,13 @@ Environment knobs:
...
 | `MIDNIGHT_LIVE_MAX_SCAN_BLOCKS` | 3600 | Above this indexer head, the guard asks you to `env:up` rather than scan. |
+| `MIDNIGHT_LIVE_KEEP_ENV` | unset | `1` leaves the stack running after the run instead of tearing it down. |
 
-`unit-live` runs up to 3 workers in parallel, so their output interleaves. It is tagged per worker: a `▶ live worker N/3 ready` banner when a worker's wallets are funded, a `[wN] ❯ <file>` line as each spec file starts, and a `[wN] ✓ <test> (<ms>) [done/total]` line per test — showing the worker, the result, and overall progress through the run. Each worker also writes a detailed log to `logs/live-harness-wN.log`.
+`integration-live` runs one worker (only the deployer wallet is in play). `unit-live` runs up to 3 workers in parallel, so their output interleaves. It is tagged per worker: a `▶ live worker N/3 ready` banner when a worker's wallets are funded, a `[wN] ❯ <file>` line as each spec file starts, and a `[wN] ✓ <test> (<ms>) [done/total]` line per test — showing the worker, the result, and overall progress through the run. Each worker also writes a detailed log to `logs/live-harness-wN.log`.
```
andrew-fleming, line 238: **nit/followup** Even though only the deployer wallet is in play, `live.setup.ts` still builds and funds 4 seeds. I'd expect this to be faster with how this is phrased. Perhaps this is something we can improve bc this takes 7 min when AFAICT it really doesn't need to

```text
live wallet 'SIGNER3' built — NIGHT 50000000000000, dust 74402993350983051718
Duration 426.64s — setup 424.44s, tests 1.69s
```

andrew-fleming (approved):
LGTM! Left a small comment and fwiw here's the issue that prolonged this review https://github.com/OpenZeppelin/compact-contracts/issues/759

Issue #722, opened 2026-07-28
Add Confidential Note Fungible Token family


0xisk:
**🧐 Motivation**

No token in the library hides the full transaction graph. `ConfidentialFungibleToken` hides amounts but keeps the account graph public. The native shielded token (#544) gets transfer privacy from Zswap but publishes every mint and burn amount as a public supply delta. A regulated confidential asset needs all three hidden at once — amounts (including issuance and burns), sender, and recipient — plus auditor visibility and seizure.

Only a note (UTXO) model can deliver that. Sender privacy on an account model has no cheap trick: a debit writes to *some* public per-account slot, and hiding which slot needs either touch-all-N cover traffic (loses on cost) or an unindexed commitment set with in-circuit membership proofs and nullifiers. The latter *is* the note model.

**📝 Details**

Umbrella for the `ConfidentialNoteFungibleToken` family. Value lives as **notes** — `(value, nonce)` owned by `pk = Hf(sk)` — represented on the public ledger only by a hiding commitment `cm = H(domain, value, nonce, pk)` in a Merkle tree. Spending publishes `nf = H(domain, nonce)` and proves membership without revealing which leaf.

Three design moves carry the family:

- **Audit-derived nonces** — every output nonce comes out of the audit ECDH, so an output the auditor cannot open cannot exist. Auditor completeness is structural, not policy.
- **Shared nullifiers** — the nullifier preimage is the nonce alone, no owner secret. Owner-spend and seizure race on the same nullifier: escrow-free clawback, and the authority never holds spend keys.
- **Supply as a policy layer** — the core writes no public supply. A deployment picks none, confidential + attested, or public.

Shipped as composable pieces (core, role gates, compliance extensions, supply variants, presets) rather than a monolith. Each piece is a sub-issue of this one.

Draft implementation: #679 — **not audited, not production**. Design doc: `confidential-note-token.md`.

PR #679, opened 2026-07-15
feat(token): confidential note fungible token draft

0xisk:
## Types of changes

What types of changes does your code introduce to OpenZeppelin Midnight Contracts?
_Put an `x` in the boxes that apply_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Update (if none of the other choices apply)

Fixes #??? 

DRAFT tier-4 exploration: a note-based confidential token with full graph privacy (amounts — including issuance and burns — sender, and recipient hidden), built as composable pieces:

- `ConfidentialNoteToken` — the core: commitment tree, nullifier single-spend, value conservation, issuer gate, plus ungated `_`-prefixed building blocks for composition.
- `extensions/ConfidentialNoteTokenAudit` — mandatory auditor viewing; every output nonce derives from the audit ECDH, so all notes are auditor-recoverable by construction.
- `extensions/ConfidentialNoteTokenDelivery` — on-chain note delivery so recipients discover funds by scanning chain data.
- `extensions/ConfidentialNoteTokenSupply` — supply as a homomorphic ElGamal ciphertext with `attestSupply` publishing a proof-backed public total.
- `presets/RegulatedConfidentialNoteToken` — the wired deployable token: issuer mint, burns, and escrow-free shared-nullifier seizure.
- `crypto/NoteDelivery` — the ECDH note-delivery primitive (unit-tested, 7 passing).
- Mocks, simulators, and witness harnesses; design doc at `contracts/src/token/docs/hybrid-confidential-token.md` and a privacy exploration overview at `contracts/privacy_readme.md`.

NOT audited, NOT production — opened as a draft for design discussion.

Known gaps (deliberate for the draft):
- No vitest suites for the token modules yet (only the NoteDelivery primitive is unit-tested); simulators/witness harnesses are in place for them.
- `privacy_readme.md` references `privacy-spectrum-*.svg`/`.html` visuals that are not committed yet.
- Preset circuits are large (transfer k=18, ~136k rows); cost-reduction ideas tracked separately (see the `feat/crypto-hashers` branch and its `compact-bug-report/`).

## PR Checklist

- [x] I have read the [Contributing Guide](../CONTRIBUTING.md#your-first-code-contribution)
[...]

PR #743, opened 2026-07-28
feat(token): add confidential note token core

### LINKED ISSUE #723: Add Confidential Note Fungible Token core
**🧐 Motivation**

The mechanism every other piece of the family builds on: a self-contained note-based token with full graph privacy (amounts, sender, recipient hidden), complete and deployable on its own with no extension.

**📝 Details**

`contracts/src/token/ConfidentialNoteFungibleToken.compact`

- State is the Zcash split: `_commitments: HistoricMerkleTree<32, Bytes<32>>` ("what value exists", append-only) and `_nullifiers: Set<Bytes<32>>` ("what was spent"). Neither reveals amounts or owners.
- `cm = H(domain, value, nonce, pk)` with a 256-bit nonce for hiding; `nf = H(domain, nonce)`, deliberately omitting any owner secret so seizure can be escrow-free.
- `_consumeNote` is the heart: proves Merkle membership against a historical root, checks the owner's spend secret, inserts the nullifier (k=14, ~12.2k rows).
- Deliberately role-free and supply-free: ungated `_`-prefixed blocks (`_mint`, `_mintNote`, `_transfer`, `_burn`, `_consumeNote`) plus value conservation; the composing contract supplies all policy.
- Standalone mode returns created notes to the caller for out-of-band handoff; the Delivery extension replaces that with on-chain discovery.

Landed in #679 (draft).

0xisk:
## Types of changes

*Put an *``* in the boxes that apply*

Fixes [#723](https://github.com/OpenZeppelin/compact-contracts/issues/723)

The `ConfidentialNoteFungibleToken` core, extracted from the exploration draft [#679](https://github.com/OpenZeppelin/compact-contracts/issues/679) onto a clean branch off `main`. **Core only** — no extensions, no presets, no crypto primitives, no concurrency utils. Those are the sibling sub-issues of [#722](https://github.com/OpenZeppelin/compact-contracts/issues/722).

Value lives as **notes**: `(value, nonce)` owned by `pk = Hf(sk)`, published only as a hiding commitment `cm = H(domain, value, nonce, pk)` in a `HistoricMerkleTree`. Spending publishes `nf = H(domain, nonce)` and proves membership in-circuit without revealing which leaf, so amounts, sender, and recipient all stay off the public ledger. The ledger carries the tree and the nullifier set and nothing else: no balances, no accounts, no supply.

**Surface**

- Self-gated `transfer` / `burn`: spending requires the owner's secret via `wit_SecretKey`, and output nonces come from the caller's own randomness witness. Created notes return to the caller as a local private result and move out of band.
- Ungated `_mint` / `_mintNote` / `_transfer` / `_burn` / `_consumeNote` building blocks. No roles, no initialization — the composing contract supplies policy.
- Exported pure circuits `derivePk` / `commitOf` / `nullifierOf` so wallets and auditors recompute off-circuit exactly what the circuits do.

**Costs** (compiler's own numbers, matching the `@circuitInfo` tags)

| circuit        | k  | rows   |
| -------------- | -- | ------ |
| `transfer`     | 16 | 36 394 |
| `_transfer`    | 15 | 25 732 |
| `burn`         | 15 | 25 584 |
| `_burn`        | 15 | 19 119 |
| `_consumeNote` | 14 | 12 248 |
| `_mint`        | 14 | 10 964 |
| `_mintNote`    | 13 | 6 766  |

**Changes from the draft in **[**#679**](https://github.com/OpenZeppelin/compact-contracts/issues/679)

- Domain tags read `OZ:note:*` instead of `OZ:cnt:*`. These are permanent on-chain constants, so the rename had to land before any deployment; costs are unchanged.
- The module doc no longer names extension files that do not exist on `main`.
- A full unit suite, which the draft did not have for this module.

## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a confidential note-based fungible token supporting minting, private transfers, change handling, and burning.
  * Protects transaction privacy by keeping amounts, ownership details, nonces, and spending secrets private while publishing only necessary cryptographic records.
  * Prevents double spending and validates note ownership, membership, and value conservation.
  * Supports secure note commitments and nullifiers for reliable transaction tracking.


0xisk:
## Recent additions

`703d4cc9..d13be552`. All six note-token test files are new in this PR, split by **what is claimed**; the technique is chosen per test.

| File | Claim |
|---|---|
| `.test.ts` | what a circuit did |
| `.property.test.ts` | the same claims over generated inputs |
| `.invariant.test.ts` | what holds across a sequence of calls |
| `.privacy.test.ts` | what the chain sees |
| `.concurrency.test.ts` | whether a call lands after someone else's |
| `.compatibility.test.ts` | the wire format and published shape |

This batch added `.compatibility`, `.property`, `.invariant`, and reworked the other two: privacy inputs are now generated, concurrency is now exhaustive at 34 cases.

**Why `.compatibility`.** Every other suite compares the module against itself, so all stay green when the wire format moves. Renaming `OZ:note:commit` keeps them passing while making every note in a deployed instance unspendable. It pins the digests, the ledger layout (the only static assertion of `HistoricMerkleTree` and `depth: 32`), and the circuit surface, keyed on the generated `Circuits` type so a renamed circuit is a compile error, cross-checked against `ProvableCircuits` and `Ledger`. Toolchain versions are recorded, not asserted: 0.31.0 and 0.31.1 give byte-identical output.

**New shared module.** `#test-utils/compiler/contractInfo.ts` reads `contract-info.json`. No Midnight package types it, so shapes come from the 472 artifacts here, which caught three a guess would miss: `Counter` has no element type, `Map` carries `key`/`value` not `type`, and `ledger` is absent rather than empty when there is no state.

**Fixed a flake** (`c749f521`): the privacy helpers assumed a digest is always 64 hex chars, but the runtime zero-trims, so ~1 in 256 is narrower.

**Filed:** #752 (compiler version declared in three places; CI pins 0.31.0, local gives 0.31.1) · #749 · #750 · #751

**Verification.** Note-token unit 127 passed / 5 skipped · harness 146 / 4 skipped · biome clean on 153 files. Each addition checked non-vacuous by mutation. Two caveats: the other 37 unit files fail to load in this worktree (uncompiled mocks, zero failed tests), and `yarn types` excludes test files, so the compile-time exhaustiveness is editor-only.

contracts/test-utils/harness/publishedTx.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+  const body = (await res.json()) as { data?: T; errors?: unknown };
+  if (body.errors) {
+    throw new Error(`indexer gql errors: ${JSON.stringify(body.errors)}`);
+  }
+  if (!body.data) {
+    throw new Error('indexer gql: empty data');
+  }
+  return body.data;
+}
```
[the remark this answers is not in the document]
0xisk: Valid, fixed in c3a88e20.

Bounded each request by the remaining deadline rather than a fixed 10s ceiling: `publishedTxsSince` issues one request per block, so a flat ceiling still allows 1200s against a 120s deadline. Timeouts now retry, protocol errors surface at once.

Added the module first test. Removing the abort signal makes both timeout cases hang, so it reproduces your case.

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,460 @@
...
+
+  /**
+   * @witness wit_SecretKey
+   * @description Returns the caller's spend secret, from which the circuits
+   * derive the identity `pk = Hf(sk)` that owns notes.
+   *
+   * @returns {Bytes<32>} secretKey - A 32-byte cryptographically secure random
+   * value.
+   */
+  witness wit_SecretKey(): Bytes<32>;
```
andrew-fleming: **followup**: I think we should scope the witness with the name of where it's used e.g. wit_ConfidentialNoteSK (or whatever name/acronym we'd like) to avoid the forceful reuse of `wit_SecretKey`, for example, if two modules have the same witness sig. I didn't include it in the CFT so this is really my bad

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,460 @@
...
+   * @param {Field} recipientPk - The identity receiving the note.
+   * @param {Uint<128>} value - The amount to create.
+   * @return {Note} - The new note, returned as a local private result.
+   */
+  export circuit _mint(recipientPk: Field, value: Uint<128>): Note {
+    const note = Note { value: value, nonce: freshNonce(pad(32, "OZ:note:out")) };
+    _mintNote(note, recipientPk);
+    // The note returns only to the local caller; revealing it on-chain would
+    // expose the nonce (spend-critical) and the amount.
+    return disclose(note);
```
andrew-fleming: I'm thinking that it might be better to forward the disclose to the contract so it's more explicit that a value or derivation thereof is disclosed. Thoughts? If so, this would need to be done across all modules so it's consistent

contracts/test-utils/harness/publishedTx.ts
```diff
@@ -0,0 +1,240 @@
...
+/** How long to wait between polls. */
+const POLL_INTERVAL_MS = 1_000;
+
+/**
+ * Ran out of time, either on one request or on the caller's whole deadline.
+ *
+ * Kept distinct from a protocol failure so {@link awaitPublishedTxs} can poll
+ * through transient slowness while a real indexer error still surfaces at once.
+ */
+class IndexerTimeout extends Error {}
```
andrew-fleming:
```suggestion
export class IndexerTimeout extends Error {}
```

Is it reasonable that a caller (us) might care about seeing this?

contracts/test-utils/harness/publishedTx.ts
```diff
@@ -0,0 +1,240 @@
...
+    const pause = Math.min(POLL_INTERVAL_MS, deadline - Date.now());
+    if (pause <= 0) {
+      break;
+    }
+    await new Promise((resolve) => setTimeout(resolve, pause));
+  }
+
+  throw new Error(
+    `indexer: expected ${min} transaction(s) after block ${height}, saw ${seen.length}`,
+  );
```
andrew-fleming, line 259: Ran this and the indexer indeed got stuck but reported it as a missing tx. When the indexer accepts the connection and never answers, AFAICT `gql` raises `IndexerTimeout` and this catch eats it and the terminal shows no cause

contracts/test-utils/harness/publishedTx.ts
```diff
@@ -0,0 +1,240 @@
...
+  height: number,
+  min = 1,
+  timeoutMs = 120_000,
+): Promise<PublishedTx[]> {
+  const deadline = Date.now() + timeoutMs;
+  let seen: PublishedTx[] = [];
+
+  while (Date.now() < deadline) {
+    try {
+      seen = await publishedTxsSince(height, undefined, deadline);
```
andrew-fleming: `awaitPublishedTxs` doesn't forward `contractAddress` so it resolves on the first tx in the window from any source including another worker (`unit-live` defaults to 3). Since a vacuous pass doesn't fail, round 2 never fires thus the mitigation can't reach the case it was written for. If we just run a single file, we wouldn't see this issue bc there's nothing competing

Here's the bad scenario flow as I understand it:

```
target + other tx fire within the window ->
other tx wins ->
neg privacy assertions scan the other tx ->
the spec passes w/o examining the target
```

Do you agree? If so, I think we just need to forward the contract address

contracts/test-utils/harness/test/publishedTx.test.ts
```diff
@@ -0,0 +1,209 @@
...
+afterEach(() => {
+  globalThis.fetch = realFetch;
+  vi.useRealTimers();
+});
+
+// ---------------------------------------------------------------------------
+// The timeout contract
+// ---------------------------------------------------------------------------
+
+describe('awaitPublishedTxs timeout', () => {
```
andrew-fleming, line 113: I'd include this test if we agree on improving the timeout msg

```
it('should keep the timeout as the cause of the failure', async () => {}
```

contracts/test-utils/harness/test/publishedTx.test.ts
andrew-fleming, line 1: A few test suggestions (including the happy path):

```
it('should return the txs once the indexer catches up', async () => {}

it('should read every block in the window, oldest first', async () => {}

it('should ignore txs from another contract', async () => {}
```

contracts/test-utils/concurrency/types.ts
```diff
@@ -0,0 +1,106 @@
...
+ *   2. the declared effects must equal the ones the replay recomputes.
+ *
+ * So a conflict is not a race in wall-clock time. It is a divergence between
+ * the state a transcript was built on and the state it is applied to, which is
+ * why {@link ConcurrencyHarness} separates `build` from `land`: a spec pins two
+ * builds to one snapshot, then applies them in order. No same-block trickery on
+ * either backend.
+ *
+ * Which of the two couplings a backend can see differs, so a claim that depends
+ * on (2) belongs in a live spec. See the SCOPE note on `DryReplayHarness`.
```
andrew-fleming, line 21: Very helpful docs 👍

contracts/test-utils/concurrency/race.ts
```diff
@@ -0,0 +1,35 @@
...
+): Promise<Outcome> {
+  const snapshot = await harness.snapshot();
+  const pendingFirst = await harness.build(first, snapshot);
+  const pendingSecond = await harness.build(second, snapshot);
+
+  await harness.land(pendingFirst);
+  const attempt = await harness.attempt(pendingSecond);
+
+  return attempt.outcome === 'landed' ? 'both-landed' : 'second-rejected';
+}
```
andrew-fleming: Would it be better to carry out the reason upon rejection and return `attempt`?

contracts/test-utils/concurrency/race.ts
andrew-fleming, line 1: What do you think about adding direct unit tests for this file?

contracts/test-utils/concurrency/race.ts
```diff
@@ -0,0 +1,35 @@
...
+export async function race<S>(
+  harness: ConcurrencyHarness<S>,
+  first: Call,
+  second: Call,
+): Promise<Outcome> {
+  const snapshot = await harness.snapshot();
+  const pendingFirst = await harness.build(first, snapshot);
+  const pendingSecond = await harness.build(second, snapshot);
+
+  await harness.land(pendingFirst);
```
andrew-fleming: Might be worth wrapping this in a try/catch. It'd be helpful for differentiating a harness bug from something like an invalid scenario

contracts/test-utils/concurrency/parties.ts
```diff
@@ -0,0 +1,75 @@
...
+ * @param factory - Contract-specific wallet and contract construction.
+ */
+export function createParties<W, C>(
+  names: readonly string[],
+  factory: PartyFactory<W, C>,
+): PartySet<W, C> {
+  const parties: Record<string, Party<W, C>> = {};
+  const contracts: Record<string, C> = {};
+
+  for (const name of names) {
```
andrew-fleming: **nit:** I think it's worth adding a duplicate name check bc they'll silently collapse

andrew-fleming: And asserting that name is not empty

contracts/test-utils/concurrency/parties.ts
```diff
@@ -0,0 +1,75 @@
...
+ * @param label - Party name, or any tag the spec wants to be reproducible.
+ * @param size - Byte length; 32 suits a Compact `Bytes<32>` secret.
+ */
+export function labelledSecret(label: string, size = 32): Uint8Array {
+  const bytes = new Uint8Array(size);
+  const encoded = new TextEncoder().encode(label);
+  if (encoded.length > size) {
+    throw new Error(
+      `labelledSecret: '${label}' needs ${encoded.length} bytes, limit is ${size}`,
+    );
```
andrew-fleming, line 26: **followup**: This is essentially redefined in a few of the `/token/test/...test.ts` files. We can remove those and just use this one. It'd prob be better to move this to `test-utils/fixtures/` somewhere. Not necessary now, a small repo improvement for the future

contracts/test-utils/concurrency/parties.ts
andrew-fleming: Worth adding tests for these cases as well

contracts/test-utils/concurrency/test/DryReplayHarness.test.ts
andrew-fleming, line 1: I think we need more tests...we have no direct calls for `apply`, `land`, `attempt`, `classify`, and `replay`. Most of which the `DryReplayHarness` exists for (replay, conflict scoring, discrimination) has no coverage

contracts/test-utils/concurrency/DryReplayHarness.ts
```diff
@@ -0,0 +1,215 @@
...
+        effects: pending.effects,
+        program: pending.transcript,
+      },
+      this.costModel,
+    ).state;
+  }
+
+  /** Every actor shares one deployed ledger; it lives here, not on them. */
+  private deploy(constructorArgs: readonly unknown[]): ChargedState {
+    const [deployer] = Object.values(this.contracts);
```
andrew-fleming: **followup**: Correct me if I'm wrong, but the shared ledger gets seeded with the first party's secrets and each subsequent race is asymmetric in a way no spec can see. If true, might be worth being explicit about this with a doc line or adding a deployer field to `HarnessOptions`

contracts/test-utils/concurrency/backend.ts
```diff
@@ -0,0 +1,48 @@
...
+ * and `error?: string`, so the ledger's own message is readable at runtime.
+ * Read it; do not hardcode it. The Display strings behind it are static data in
+ * the wasm binary, exported nowhere.
+ *
+ * Known open question, settled only by running it: this module's conflicts are
+ * GUARANTEED-segment failures (no `Kernel.checkpoint` anywhere in the note
+ * core), and a guaranteed-phase failure keeps the transaction out of any block.
+ * So there may be no `FinalizedTxData` and no `TransactionResult` to read, and
+ * the signal is a submit-time rejection or a timeout instead. `FailEntirely` and
+ * `TransactionResult.error` are what a FALLIBLE-segment failure produces. The
```
andrew-fleming, line 29: **followup**: The docs are really informative here. This feels like it should be in a design doc though. Just throwing it out there

contracts/test-utils/compiler/contractInfo.ts
```diff
@@ -0,0 +1,368 @@
...
+}
+
+/** The whole of `contract-info.json`. */
+export interface ContractInfo {
+  readonly 'compiler-version': string;
+  readonly 'language-version': string;
+  readonly 'runtime-version': string;
+  readonly circuits: readonly CircuitInfo[];
+  readonly witnesses: readonly WitnessInfo[];
+  /** Child contracts. `unknown` because it is empty in all 472 artifacts here. */
```
andrew-fleming: I wouldn't post exact numbers like this. This is immediately wrong once something is added or removed it becomes both unreasonable and valueless to try and remember to update this

andrew-fleming: This should be a rule for the repo: No exact numbers that quickly rot and provide no value

contracts/test-utils/compiler/contractInfo.ts
```diff
@@ -0,0 +1,368 @@
...
+  readonly contracts: readonly unknown[];
+  /**
+   * The public ledger. ABSENT, not empty, when a contract declares no ledger
+   * state (64 of 472 artifacts). Prefer {@link ledgerSlots}.
+   */
+  readonly ledger?: readonly LedgerSlot[];
```
andrew-fleming: Soooo this seems backwards. I measured through all the artifacts and 0 were `ABSENT` and 35 were `[]`

contracts/test-utils/compiler/contractInfo.ts
```diff
@@ -0,0 +1,368 @@
...
+ */
+export function readContractInfo(artifactName: string): ContractInfo {
+  const path = new URL(
+    `../../artifacts/${artifactName}/compiler/contract-info.json`,
+    import.meta.url,
+  );
+
+  try {
+    return JSON.parse(readFileSync(path, 'utf8')) as ContractInfo;
+  } catch (cause) {
```
andrew-fleming, line 474: AFAIU everything above this line exists so TS can stop you misreading a slot, but that only holds if the json actually matches the declarations and `as` skips the check, no?

Line 28 covers the maintenance side: "On an unrecognized name, add the variant. Widening to string defeats the point." I agree with the approach but it only helps someone who already knows a new name turned up and nothing here surfaces that

contracts/test-utils/assertions/rejection.ts
```diff
@@ -0,0 +1,101 @@
...
+      `expected a rejection including "${reason}", but the call resolved`,
+    );
+  }
+  if (!rejectionIncludes(thrown, reason)) {
+    const chain = causeChain(thrown)
+      .map((text, depth) => `  [${depth}] ${text}`)
+      .join('\n');
+    throw new Error(
+      `expected a rejection including\n  ${reason}\nbut the cause chain was:\n${chain}`,
+    );
```
andrew-fleming, line 103:
```suggestion
      `expected a rejection including\n  ${reason}\nbut the cause chain was:\n${chain}`,
      { cause: thrown },
    );
```
Preserves the object

contracts/test-utils/assertions/rejection.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  if (!rejected) {
+    throw new Error(
+      `expected a rejection including "${reason}", but the call resolved`,
+    );
+  }
+  if (!rejectionIncludes(thrown, reason)) {
+    const chain = causeChain(thrown)
+      .map((text, depth) => `  [${depth}] ${text}`)
```
andrew-fleming:
```suggestion
      .map((text, index) => `  [${index}] ${text}`)
```
**nit**: `depth` is correct for a linear `causeChain` but I believe this is false as soon as `AggregateError` branches

contracts/test-utils/assertions/rejection.ts
```diff
@@ -0,0 +1,101 @@
...
+ * On a miss it prints the entire chain, so a run on a backend that wraps errors
+ * differently reports what it actually got instead of just "no match".
+ *
+ * @param call - The pending call, e.g. `token.burn(30n)`.
+ * @param reason - Substring to find, at any depth.
+ */
+export async function expectRejection(
+  call: Promise<unknown>,
+  reason: string,
+): Promise<void> {
```
andrew-fleming:
```suggestion
): Promise<unknown> {
```
**nit**: This would be better if it was composable, no?

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,460 @@
...
+  witness wit_Path(cm: Bytes<32>): MerkleTreePath<32, Bytes<32>>;
+
+  /**
+   * @witness wit_NonceRandomness
+   * @description Returns the seed the core derives its default output nonces
+   * from.
+   *
+   * @dev MUST be fresh and secret per invocation. A reused seed re-derives an
+   * identical note, and because the nullifier depends only on the nonce, the
+   * two notes share one nullifier: spending either burns both.
```
andrew-fleming: Two things that should be documented:
1. The MUST is not enforceable
2. More importantly, the recipient of a transfer depends on the discipline of the sender to use a fresh nonce. A careless sender can create a race that multiple recipients won't ever know about until it's too late and they have no way of protecting themselves

andrew-fleming: The collision guarantees that one of two unrelated recipients loses their funds and they won't even know a race is happening (unless alice lets them know she reused the same randomness). I think this needs to be refactored

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,460 @@
...
+   * invocation, which share a single witness call: without it, a transfer's
+   * output and change would collide on one nonce, hence one nullifier.
+   *
+   * @param {Bytes<32>} slot - The per-output domain tag.
+   * @return {Field} - The output note's nonce.
+   */
+  circuit freshNonce(slot: Bytes<32>): Field {
+    return degradeToTransient(persistentHash<Vector<3, Bytes<32>>>(
+      [wit_NonceRandomness(), pad(32, "OZ:note:nonce:core"), slot]));
+  }
```
andrew-fleming: This derives output nonces from data only the sender controls; thus, a bug in the sender's rng destroys other people's notes. One way to get around this issue would be to fold in the `nullifierOf` the spent nullifier input (call it a `spendId` or something)

andrew-fleming: To visualize, it'd be something like this for `transfer` and `burn`:

```
const input = _consumeNote(pk);          // not _inputNote()
const spendId = nullifierOf(input);
const outNote = Note { value: value, nonce: freshNonce(pad(32, "OZ:note:out"), spendId) };
```

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,460 @@
...
+ * building blocks, and the composing contract decides who may create value.
+ * The `_`-prefixed blocks (`_mintNote`, `_transfer`, `_burn`, `_consumeNote`)
+ * accept caller-built notes, so a composing contract can source nonces from its
+ * own emission policy; `_mint` instead derives the core-default nonce. Like all
+ * `_` circuits they carry no authorization — the composer gates them.
+ *
+ * @dev Nonces are spend-critical: the nullifier preimage is the nonce alone,
+ * with no owner secret, so any party that knows a nonce derives the SAME
+ * nullifier. Output nonces MUST therefore be unique and unpredictable, and
+ * `wit_NonceRandomness` MUST return a fresh, secret seed per invocation. That
```
andrew-fleming: If the architecture regarding reused randomness does not change, the following must be documented:
- who bears the loss
- that the recipient cannot detect the collision at receipt
- and that the sender's own change note collides too

andrew-fleming: And on a different note, I think it's worth stating that the sender knows the randomness, the pk, and the value so they'll know *when* their sent note has been spent. It's nothing super crazy but worth being explicit about

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,460 @@
...
+ * nullifier `nf = H(domain, nonce)` marks the note spent. Spending proves that
+ * the note's commitment is in the tree without revealing which leaf.
+ *
+ * The public ledger therefore holds only the commitment tree and the nullifier
+ * set: no balances, no accounts, no supply. A transfer publishes one nullifier
+ * and two commitments, and nothing else.
+ *
+ * The module is deliberately barebones — the note machinery and nothing else.
+ * It holds no roles and needs no initialization. `transfer` and `burn` work out
+ * of the box because they are self-gated: spending requires the owner's secret
```
andrew-fleming: Somewhere in here, what do you think about explicitly defining how we're hashing in this module? A quick AI-gen draft:

```
 * @dev NOTATION. `H(...)` is `persistentHash` (SHA-256) over the NAMED STRUCT
 * shown, `CommitPreimage` for `cm` and `NullifierPreimage` for `nf`, using
 * Compact's struct encoding. It is NOT a byte concatenation of the listed
 * fields, and no concatenation reproduces it. `Hf(...)` is that same hash
 * composed with `degradeToTransient`, which reduces the digest to a `Field`.
 *
 * Do NOT reimplement these. `commitOf`, `nullifierOf`, and `derivePk` are
 * exported as pure circuits so wallets and auditors can call the exact
 * derivations the circuits use. An implementation that gets the encoding wrong
 * produces notes you cannot locate and identities you cannot spend from, with
 * no recovery path. If you must reimplement, check against the pinned vectors
 * in `ConfidentialNoteFungibleToken.compatibility.test.ts`.
 ```

File: contracts/src/token/ConfidentialNoteFungibleToken.compact [hunk 25]
```diff
@@ -0,0 +1,460 @@
...
+   * @param {Field} recipientPk - The identity receiving `outNote`.
+   * @param {Note} outNote - The recipient's output note.
+   * @param {Note} changeNote - The spender's change note.
+   * @return {[]} - Empty tuple.
+   */
+  export circuit _transfer(spenderPk: Field, recipientPk: Field, outNote: Note, changeNote: Note): [] {
+    const input = _consumeNote(spenderPk);
+    assert(input.value == outNote.value + changeNote.value,
+           "ConfidentialNoteFungibleToken: transfer does not conserve value");
+    _mintNote(outNote, recipientPk);
```
andrew-fleming, line 623:
```suggestion
    assert(input.value == outNote.value + changeNote.value,
           "ConfidentialNoteFungibleToken: transfer does not conserve value");
    assert(outNote.nonce != changeNote.nonce,
           "ConfidentialNoteFungibleToken: output notes share a nonce");
    assert(outNote.nonce != input.nonce && changeNote.nonce != input.nonce,
           "ConfidentialNoteFungibleToken: output reuses the spent input nonce");
    _mintNote(outNote, recipientPk);
```
The two assert suggestions:
- prevent transferring a dead note
- prevent two notes sharing a nullifier

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,460 @@
...
+   *
+   * @notice Taking both output notes as parameters is what lets a composing
+   * contract own its emission policy (audit-derived nonces, delivery records,
+   * supply accounting) while the core still enforces conservation.
+   *
+   * @warning No authorization. Whoever supplies a valid input note and its path
+   * can spend it; the composer decides who may call this.
+   *
+   * Requirements:
+   *
```
andrew-fleming: I'd mention that it's the composer's responsibility to provide nonce uniqueness. Output nonces are not checked for global uniqueness and they can't be

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,460 @@
...
+   * account-based token cannot offer.
+   *
+   * @warning No authorization. The composer gates who may create value.
+   *
+   * @param {Field} recipientPk - The identity receiving the note.
+   * @param {Uint<128>} value - The amount to create.
+   * @return {Note} - The new note, returned as a local private result.
+   */
+  export circuit _mint(recipientPk: Field, value: Uint<128>): Note {
+    const note = Note { value: value, nonce: freshNonce(pad(32, "OZ:note:out")) };
```
andrew-fleming:
```suggestion
    const note = Note { value: value, nonce: freshNonce(pad(32, "OZ:note:mint")) };
```
In its current state, tokens that are minted cannot be transferred if there's a pinned seed. Any reason not to make the tag unique to minting?

contracts/src/token/test/ConfidentialNoteFungibleToken.privacy.test.ts
andrew-fleming, line 1: Two things:
1. Great work on separating the tests like this. I think it makes a lot of sense
2. Since this particular suite pins the seed (which I agree with here), we need to have a check that the notes are spendable. Alice is currently transferring dead notes to bob :(
    - this is fixed with the suggestion to change the tag in `_mint`, but it should still be tested nonetheless

andrew-fleming (commented):
Very nice work, @0xisk! It's a partial review but I left some comments :)

contracts/src/token/ConfidentialNoteFungibleToken.compact
0xisk: Good catch, agree with your finding. :thumbsup:

contracts/src/token/ConfidentialNoteFungibleToken.compact
0xisk: Good catch :thumbsup:

contracts/src/token/ConfidentialNoteFungibleToken.compact
0xisk: Done in [`6896040b`](https://github.com/OpenZeppelin/compact-contracts/commit/6896040b523ce21e700c8f6f63af4b4678687f48).

contracts/src/token/ConfidentialNoteFungibleToken.compact
0xisk: Done in [`6145aad0`](https://github.com/OpenZeppelin/compact-contracts/commit/6145aad0e417637733c4c35e87a84c4c2dc88c2c) — binder is a `Field`: the spent `nf` for transfer/burn, the recipient for `_mint`. Read from the peek, since `_transfer`/`_burn` already consume.

0xisk: Covered by [`6145aad0`](https://github.com/OpenZeppelin/compact-contracts/commit/6145aad0e417637733c4c35e87a84c4c2dc88c2c) above.

contracts/src/token/ConfidentialNoteFungibleToken.compact
0xisk: Done, differently — [`5f23a4ac`](https://github.com/OpenZeppelin/compact-contracts/commit/5f23a4aca259c9c6e4d534be2df125f324ccd607) reserves every output nonce in `_issuedNonces` inside `_mintNote`, so both cases (and any other reuse) fail as "nonce already issued". Per-key pin, distinct mints still commute.

contracts/src/token/test/ConfidentialNoteFungibleToken.privacy.test.ts
0xisk: Done in [`4cf665d6`](https://github.com/OpenZeppelin/compact-contracts/commit/4cf665d69c6c6c216169d01e7fd67c98d52160fb) — recipient spend plus chained transfer and chained burn. All three failed before [`6896040b`](https://github.com/OpenZeppelin/compact-contracts/commit/6896040b523ce21e700c8f6f63af4b4678687f48)/[`6145aad0`](https://github.com/OpenZeppelin/compact-contracts/commit/6145aad0e417637733c4c35e87a84c4c2dc88c2c).

0xisk:
Beyond the threads above, this push also lands:

- [`5f23a4ac`](https://github.com/OpenZeppelin/compact-contracts/commit/5f23a4aca259c9c6e4d534be2df125f324ccd607) `_issuedNonces`: a duplicate output nonce now fails the tx instead of producing a dead note. Per-key `member` pin, so distinct mints in one block still commute; the concurrency suite asserts both.
- [`fac487ff`](https://github.com/OpenZeppelin/compact-contracts/commit/fac487ffcfb80ffdbc2c79d7d3fc90193321cc55) tree depth is a module generic (`import "…"<32> prefix Core_;`). Mock stays at 32, so vectors and rows are unchanged.
- [`cc42c77d`](https://github.com/OpenZeppelin/compact-contracts/commit/cc42c77d66fa7f36b418fdbd72571d70b370906a), [`e005c661`](https://github.com/OpenZeppelin/compact-contracts/commit/e005c6611faf769e52e49c9fb983d1d4e52c0020), [`a7a907ba`](https://github.com/OpenZeppelin/compact-contracts/commit/a7a907ba7cd3f8a5725f47cda90ebcf482ebb33e) module header rewritten into NFR sections (derivations, compatibility, privacy, security, concurrency, client-side duties, scalability, performance); per-circuit docs trimmed to what is local.
- [`6145aad0`](https://github.com/OpenZeppelin/compact-contracts/commit/6145aad0e417637733c4c35e87a84c4c2dc88c2c) also renames the change slot tag to `OZ:note:chng`.
- [`f5689c3a`](https://github.com/OpenZeppelin/compact-contracts/commit/f5689c3ac22ae76fc80899a8d74248ef1762afcb) fixes a latent flake in the transcript-shape test (zero-trimmed digests).

Row impact: `transfer` 36394 → 53501 (still k=16), `burn` and `_transfer` cross into k=16, `_mint` into k=15.

contracts/src/token/ConfidentialNoteFungibleToken.compact
0xisk: Resolved by design in [`5f23a4ac`](https://github.com/OpenZeppelin/compact-contracts/commit/5f23a4aca259c9c6e4d534be2df125f324ccd607): `_mintNote` reserves every output nonce in `_issuedNonces`, so a repeated nonce fails the tx instead of committing a dead note. Nobody bears a loss, the recipient never receives a colliding note, and the change note goes through the same check. The residual case, two mints to the same recipient under a repeated seed, is rejected too. Header Security "Freshness" documents it.

0xisk: Documented in [`cc42c77d`](https://github.com/OpenZeppelin/compact-contracts/commit/cc42c77d66fa7f36b418fdbd72571d70b370906a), header Privacy "Counterparty": whoever knows a note's nonce derives its `tag` and `nf`, so sees when it is minted and when it is spent, not by whom or for how much.

contracts/src/token/ConfidentialNoteFungibleToken.compact
0xisk: The MUST is now enforced on-chain by [`5f23a4ac`](https://github.com/OpenZeppelin/compact-contracts/commit/5f23a4aca259c9c6e4d534be2df125f324ccd607): a reused nonce is rejected at `_mintNote`, whichever circuit produced it, so a recipient no longer depends on the sender's discipline. What stays client-side is only the quality of the randomness behind `_mint`'s own outputs, stated under Security "Freshness" in [`cc42c77d`](https://github.com/OpenZeppelin/compact-contracts/commit/cc42c77d66fa7f36b418fdbd72571d70b370906a).

contracts/src/token/ConfidentialNoteFungibleToken.compact
0xisk: Mostly moot after [`5f23a4ac`](https://github.com/OpenZeppelin/compact-contracts/commit/5f23a4aca259c9c6e4d534be2df125f324ccd607): the core checks every output nonce against `_issuedNonces` at commit time, so a composer-built note that reuses one fails. The composer still owns unpredictability of the nonces it builds; the `_mintNote` `@notice` in [`e005c661`](https://github.com/OpenZeppelin/compact-contracts/commit/e005c6611faf769e52e49c9fb983d1d4e52c0020) says so.

contracts/src/token/ConfidentialNoteFungibleToken.compact
0xisk: Done in [`cc42c77d`](https://github.com/OpenZeppelin/compact-contracts/commit/cc42c77d66fa7f36b418fdbd72571d70b370906a) as header Notation + Derivations (Notation renamed in [`c0650f2d`](https://github.com/OpenZeppelin/compact-contracts/commit/c0650f2dbae3cc566d0fbd083aa8a9619d97a697)): `H`/`Hf` defined, struct-encoded preimages, do not reimplement but call the exported pure circuits, vectors pinned in the compatibility suite. Shorter than your draft, same content.

contracts/src/token/ConfidentialNoteFungibleToken.compact
0xisk: Done in [`3a951b02`](https://github.com/OpenZeppelin/compact-contracts/commit/3a951b02a882243b7dbe7a3c3b8ce800010a5dc1): `wit_ConfidentialNoteSK`, `wit_ConfidentialNoteInputNote`, `wit_ConfidentialNotePath`, `wit_ConfidentialNoteNonceRandomness`. Rows unchanged; the witness impl, simulator and the one shared test-util fixture that quoted the `wit_Path` error string move with it.

contracts/test-utils/harness/publishedTx.ts
0xisk: Agreed, that flow was possible. Done in [`b92ca26b`](https://github.com/OpenZeppelin/compact-contracts/commit/b92ca26b32ec6bb766d9be0704477c7a7af30f8b): `awaitPublishedTxs` takes the address as a required parameter and forwards it, so only the target's txs count toward `min`. The four live call sites read it off the simulator backend. Required rather than defaulted so a caller cannot fall back to the old vacuous wait.

contracts/test-utils/harness/publishedTx.ts
0xisk: Done in [`eadd4800`](https://github.com/OpenZeppelin/compact-contracts/commit/eadd48005942fbf5a5ef94dda2bdf9bbd10ab9ab): the last `IndexerTimeout` is kept as `cause` and named in the message ("timed out waiting for … last request: no response in Nms"). It is cleared on any successful poll, so a stall the indexer recovers from still reports a short window rather than a timeout.

contracts/test-utils/harness/publishedTx.ts
0xisk: Yes. Exported in [`eadd4800`](https://github.com/OpenZeppelin/compact-contracts/commit/eadd48005942fbf5a5ef94dda2bdf9bbd10ab9ab) so a spec can assert on the cause.

contracts/test-utils/harness/test/publishedTx.test.ts
0xisk: Added in [`eadd4800`](https://github.com/OpenZeppelin/compact-contracts/commit/eadd48005942fbf5a5ef94dda2bdf9bbd10ab9ab) under that name, plus one for the recovery case: a timed-out request followed by answered polls reports the short window with no cause.

contracts/test-utils/harness/test/publishedTx.test.ts
0xisk: All three added in [`b92ca26b`](https://github.com/OpenZeppelin/compact-contracts/commit/b92ca26b32ec6bb766d9be0704477c7a7af30f8b) under your names.

contracts/test-utils/compiler/contractInfo.ts
0xisk: Confirmed on main: 76 artifacts, 0 absent, 39 `[]`. Fixed in [`fead3c9b`](https://github.com/OpenZeppelin/compact-contracts/commit/fead3c9b255dc706ccdeb1faa2a07973ba8a6fee): `ledger` is now required on the type and in the validator, the doc says "always present, `[]` when empty", and `ledgerSlots` is a plain accessor.

contracts/test-utils/compiler/contractInfo.ts
0xisk: Correct, the cast checked nothing. Done in [`fead3c9b`](https://github.com/OpenZeppelin/compact-contracts/commit/fead3c9b255dc706ccdeb1faa2a07973ba8a6fee): the unions are derived from `as const` arrays and `readContractInfo` walks the parsed JSON, throwing "unrecognized type-name X; add the variant to CompactTypeName" (same for storage kinds and missing top-level keys). Validation sits outside the "compile first" catch so a mismatch is never reported as a missing build. Tests run on inline objects via the new `parseContractInfo`, so `test:harness` still needs no compiled artifact.

contracts/test-utils/compiler/contractInfo.ts
0xisk: Agreed. Removed in [`b83fa1ee`](https://github.com/OpenZeppelin/compact-contracts/commit/b83fa1ee1f773b3923026a73b22097a4b2495cb4); the one claim that mattered (`ledger` always present) is now enforced by the validator instead of asserted by a count. On the repo rule: happy to add a line to CONTRIBUTING in a follow-up PR rather than here.

contracts/test-utils/assertions/rejection.ts
0xisk: Done in [`98f0ff8c`](https://github.com/OpenZeppelin/compact-contracts/commit/98f0ff8c28782ac476318254266d3aa8b830ef5b).

contracts/test-utils/assertions/rejection.ts
0xisk: Done in [`c0472cd9`](https://github.com/OpenZeppelin/compact-contracts/commit/c0472cd9a27c65a043be3a61b592d624e9785b38): returns the rejection so a spec can assert further on it (e.g. `cause instanceof IndexerTimeout`) without re-catching. The seven call sites ignore the value.

contracts/test-utils/assertions/rejection.ts
0xisk: Correct, it is breadth-first position, not depth. Done in [`7034d89a`](https://github.com/OpenZeppelin/compact-contracts/commit/7034d89a7abb0b26f8cc3ed5cdd3487fcac5f07a): renamed to `index` and the `causeChain` doc no longer claims the order matches nesting.

contracts/test-utils/concurrency/race.ts
0xisk: Done in [`fc54aeab`](https://github.com/OpenZeppelin/compact-contracts/commit/fc54aeab6b36398d0961cbdc0bd35811773ef15f): each of the five steps is wrapped so a throw names the phase (`race: build of second call 'alice.transfer' failed`) and keeps the original as `cause`. A rejected `attempt` is a result, not an error, so it is left alone.

contracts/test-utils/concurrency/race.ts
0xisk: Done in [`fc54aeab`](https://github.com/OpenZeppelin/compact-contracts/commit/fc54aeab6b36398d0961cbdc0bd35811773ef15f): `race` returns `{ outcome: 'second-rejected', reason }` and the matrix now asserts a non-empty reason on every rejecting case.

contracts/test-utils/concurrency/race.ts
0xisk: Done in [`c096a7d1`](https://github.com/OpenZeppelin/compact-contracts/commit/c096a7d169de0ee7f8285bda406cf751ab543f85): eight tests against a recording in-memory harness (same snapshot for both builds, land-before-attempt ordering, verdict and reason pass-through, phase-tagged throws).

contracts/test-utils/concurrency/parties.ts
0xisk: Done in [`96ae161d`](https://github.com/OpenZeppelin/compact-contracts/commit/96ae161d41f4043c3edd93a60e2d24a3f12ea47f): duplicate and blank names throw at construction, naming the offender.

0xisk: Included in [`96ae161d`](https://github.com/OpenZeppelin/compact-contracts/commit/96ae161d41f4043c3edd93a60e2d24a3f12ea47f).

0xisk: Both cases tested in [`96ae161d`](https://github.com/OpenZeppelin/compact-contracts/commit/96ae161d41f4043c3edd93a60e2d24a3f12ea47f).

contracts/test-utils/concurrency/test/DryReplayHarness.test.ts
0xisk: Agreed. Done in [`1cb449fd`](https://github.com/OpenZeppelin/compact-contracts/commit/1cb449fd11d9d258151f3cc82f23e9fb61894728): `apply`, `land`, `attempt`, `classify` and `replay` are pinned against the compiled note mock, including the two that would silently break the matrix (replay against the current state, not the build snapshot; a transcript invalid against its own snapshot is rethrown, not scored). It lives under `src/` because only the `unit` project depends on a compile; the harness suite's header pointed at a `harness invariants` describe that never existed, now repointed here.

contracts/test-utils/concurrency/backend.ts
0xisk: Agreed, it reads as a design doc. Deferred to #861 on the release-train backlog; `backend.ts` keeps a summary and a link once that lands.

contracts/test-utils/concurrency/parties.ts
0xisk: Deferred to #862 on the release-train backlog: move `labelledSecret` to `test-utils/fixtures/` and drop the six local copies. Kept out of this PR to avoid churning the reviewed test files.

contracts/test-utils/concurrency/DryReplayHarness.ts
0xisk: You are right: `deploy()` took whichever instance `Object.values` yielded first, so that party's witnesses answered the constructor. Moot for the note core (no initializer) but real for any module with one. Done in [`0009a312`](https://github.com/OpenZeppelin/compact-contracts/commit/0009a3120e4dc5da78f1d46da2b4315ce7fdb778): `HarnessOptions.deployer` names the party, defaults to the first, throws on an unknown name before anything deploys. Dry only; the live backend deploys from the wallet pool's own deployer alias, which is not a party, so the option says so rather than guessing a mapping.

0xisk:
Merged `main` and moved the branch onto the 0.34 toolchain (compiler 0.34.0, compact-runtime 0.19.0, compact-simulator 0.4, midnight-js 5.0.0-beta.7, pragma `>= 0.26.0`). Merge commit rather than rebase so the commit links in the threads above stay valid.

- [`768ee438`](https://github.com/OpenZeppelin/compact-contracts/commit/768ee43863722e9e0f27d72c8c14d6e40644654a) merge, no conflicts.
- [`6a46cd0a`](https://github.com/OpenZeppelin/compact-contracts/commit/6a46cd0a7103e6e1a8ce15f5ce96ca9f3e88c75f) pragma bump on the module and mock. The contract itself needed nothing else.
- [`9c0db6fa`](https://github.com/OpenZeppelin/compact-contracts/commit/9c0db6fab34527a88cb5af2033d0d8e49f06dfce) dry replay harness on runtime 0.19: `createDryHarness` is now async (contract constructor became async), proof data comes from `callProofDataTrace`, deploy time pinned to 0 because the runtime now defaults it to `Date.now()`, which would make differential runs diverge. `runTranscript` / `CostModel` / `ChargedState` unchanged.
- [`b8613037`](https://github.com/OpenZeppelin/compact-contracts/commit/b86130373b7cea40984af23f3d94d2ed36e89b4a) privacy suite's direct-drive probe ported (async factory, async circuits). The other note suites go through the simulator and are untouched.
- [`f4cdd2a7`](https://github.com/OpenZeppelin/compact-contracts/commit/f4cdd2a7cb57fd2f975398fdb23e053a742a0b02) rows re-measured: every circuit 0.5–1% cheaper, no k change (`transfer` 53501 → 53110, still k=16). Compatibility vectors unchanged, so no wire-format impact. No `--feature-zkir-v3` needed for this module.

Note for anyone checking out the branch: the artifact records the runtime version, so run `compile:token` before the note suites go green.

pepebndc:
### Predictable output nonces can be reserved before the intended spend

`_mintNote` reserves `issuedTagOf(note)` in a global set. The tag depends on the nonce, independently of the owner and value.

If a composer permits attacker-chosen output nonces, an attacker can reserve a predictable victim nonce by creating a note with that nonce first. The victim's transaction then fails with `nonce already issued`. If retries derive the same nonce, the denial persists.

This requires both attacker control over a note-creation path and predictability of the victim's output nonce. The core's default derivation uses secret randomness and does not expose this condition.

**Proposed solution:** Document unpredictability as a requirement for composer-defined nonce policies, alongside uniqueness. Derive output nonces from fresh secret randomness or a secret unavailable to untrusted callers. A public counter or public transaction context alone is insufficient. If a reservation collision occurs, the wallet must derive a fresh nonce before rebuilding the transaction.

Keep the global uniqueness check: removing it would permit notes to share a nullifier. Add a regression test for an attacker reserving a predictable nonce before the intended spend.

contracts/test-utils/harness/test/publishedTx.test.ts
```diff
@@ -0,0 +1,343 @@
...
+/** A GraphQL 200 carrying `data`. */
+const ok = (data: unknown): Response =>
+  ({
+    ok: true,
+    status: 200,
+    json: async () => ({ data }),
+  }) as unknown as Response;
+
+/** The contract under observation. Stub blocks carry calls to this address. */
+const ADDR = '0xc0ffee';
```
andrew-fleming: <3

contracts/src/token/ConfidentialNoteFungibleToken.compact
andrew-fleming: Very nice 👍

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,626 @@
...
+   * @dev Historic, not plain:
+   * `checkRoot` accepts any past root, so a proof survives concurrent inserts.
+   * @type {HistoricMerkleTree<depth, Bytes<32>>} _commitments - The commitment
+   * tree, holding up to `2^depth` leaves.
+   */
+  export ledger _commitments: HistoricMerkleTree<depth, Bytes<32>>;
+
+  /**
+   * @description Published nullifiers: what has been spent. Unlinkable to any
+   * commitment, owner, or value, so the set reveals only a spend count.
```
andrew-fleming:
```suggestion
   * commitment, owner, or value, so the set reveals only a spend count.
   * To an observer who does not hold the note, the set reveals only a spend
   * count; see the header's Privacy notice for what a nonce-holder sees.
```
to be thorough

contracts/test-utils/harness/publishedTx.ts
```diff
@@ -0,0 +1,260 @@
...
+      const calls = tx.contractActions
+        .filter((action) => action.entryPoint !== undefined)
+        .map((action) => ({
+          address: action.address ?? '',
+          entryPoint: action.entryPoint ?? '',
+          state: action.state ?? '',
+        }));
+      if (contractAddress !== undefined) {
+        const wanted = contractAddress.replace(/^0x/, '');
+        if (!calls.some((call) => call.address.replace(/^0x/, '') === wanted)) {
```
andrew-fleming: Since `contractAddress` is required in `awaitPublishedTxs`, I think we should probably normalize casings too, no?

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,626 @@
...
+ *
+ * @notice Scope:
+ * Note machinery only: no roles, no initialization. `transfer` and `burn` are
+ * self-gated by `wit_ConfidentialNoteSK` and the caller's randomness witness.
+ * Created notes return to the caller as a private result and reach recipients
+ * out of band.
+ *
+ * @notice Client-side duties:
+ * - Identity: `pk = Hf(sk)` is this token's address, unrelated to the Zswap
+ *   coin key. Derive `sk` from the wallet seed under a module-specific domain
```
andrew-fleming: Two things here:
1. `'pk = Hf(sk)' is this token's address...` reads like this to me: "`pk = ...` is this token's contract address..."
2. It's fine to specify that it's unrelated to `ZswapCoinPublicKey` but what about global ids that also use `H(sk)`?

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,626 @@
...
+ *   - One `nf` per note spent.
+ *   - One `tag` and one opaque leaf digest per note created.
+ *   - The Merkle root the spend proves against.
+ *   - The transcript shape, which does not vary with amount or recipient.
+ * - Public, as state:
+ *   - The commitment tree, the nullifier set, the issued-nonce set.
+ *   - No balances, no accounts, no supply.
+ * - Private:
+ *   - Amounts.
+ *   - Sender and recipient `pk`s.
```
andrew-fleming: Sooo this appears true in the module, but wouldn't implementations make this not true if a user re-used their sk from a different contract (using the untagged global ids e.g. FungibleToken)?

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,626 @@
...
+  /**
+   * @description `pk = Hf(sk)`. Exported so wallets and auditors derive
+   * identities the way the circuits do.
+   *
+   * @param {Bytes<32>} sk - The owner's spend secret.
+   * @return {Field} - The spend identity that owns notes.
+   */
+  export pure circuit derivePk(sk: Bytes<32>): Field {
+    return degradeToTransient(persistentHash<Bytes<32>>(sk));
+  }
```
andrew-fleming: What do you think about including a tag here? My thinking is that the module claims unlinkability in the header doc...but it's using the same scheme as the global account id in the public modules

andrew-fleming: Further, it might be worth integrating instance salt to harden unlinkability. If there are two instances of the CNFT, even with the tag, and a user re-uses their SK, we have the same id across both. We can't enforce a strong and unique salt value, but I think it's better than not having it at all. lmk what you think

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,626 @@
...
+ *   recipient for
+ *   `_mint`), and `_mintNote` rejects an already-issued nonce, so a repeated
+ *   seed fails loudly instead of producing a dead note.
+ * - Ownership: a spend proves `cm` binds `Hf(wit_ConfidentialNoteSK())`; no
+ *   secret, no spend.
+ * - No authorization on `_` circuits: `_mint`, `_mintNote`, `_transfer`,
+ *   `_burn`, `_consumeNote` are ungated. Re-exporting one from a deployed
+ *   contract is a permissionless mint or spend. The composer gates them.
+ * - Atomicity: a failed assert rejects the whole tx before any state moves,
+ *   and the sender pays no fee.
```
andrew-fleming: I'm not sure I like `Atomicity` as a section. This is describing protocol behavior in general. This would fit better in GUIDELINES. I'd suggest either cutting it or describing where it's relevant to the module

andrew-fleming: If you want to keep it, maybe something like this?
```suggestion
 * - Atomicity: conservation assumes a call applies whole. `_transfer`
 *   publishes the input's nullifier before minting its outputs, so a partially
 *   applied call would destroy value. Compact gives this by default. A
 *   composer that introduces `kernel.checkpoint()` gives it up
```

contracts/src/token/test/ConfidentialNoteFungibleToken.test.ts
```diff
@@ -0,0 +1,875 @@
...
+  });
+
+  it('should not let anyone other than the owner burn', async () => {
+    spendAs(BOB_SK, input);
+
+    await expectRejection(
+      token.burn(30n),
+      'wit_ConfidentialNotePath: commitment not found in tree',
+    );
+  });
```
andrew-fleming: Correct me if I'm wrong but the error is thrown by the test witness's `findPathForLeaf` lookup before `_consumeNote` runs `checkRoot` or `cm == path.leaf`, no? So they show that the ref wallet can't build a path for a note it doesn't own (and not that the contract rejects a forged one)

Hypothetically, if we delete the `cm == path.leaf` assert, the test would still pass and there'd be no attempted proof if ran live

contracts/src/token/test/ConfidentialNoteFungibleToken.test.ts
```diff
@@ -0,0 +1,875 @@
...
+
+  it('should not consume a note under an owner pk it was not committed to', async () => {
+    await expectRejection(
+      token._consumeNote(BOB),
+      'wit_ConfidentialNotePath: commitment not found in tree',
+    );
+  });
```
andrew-fleming: Similar thing here

contracts/src/token/test/ConfidentialNoteFungibleToken.test.ts
```diff
@@ -0,0 +1,875 @@
...
+  // that is not in the tree, so no membership proof exists.
+  it('should not let anyone other than the owner spend', async () => {
+    spendAs(BOB_SK, input);
+
+    await expectRejection(
+      token.transfer(CAROL, 30n),
+      'wit_ConfidentialNotePath: commitment not found in tree',
+    );
+    expect(await isSpent(input)).toBe(false);
+  });
```
andrew-fleming: And here. I think it's fine to keep these tests but we should also make sure that there are tests triggering the `cm == path.leaf` assert

contracts/src/token/test/ConfidentialNoteFungibleToken.test.ts
andrew-fleming: Using the real path (`pathOverride = await pathFor(input, ALICE)`) and expecting `path does not match input commitment` would make the circuit the thing that rejects

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,626 @@
...
+ *   - The Merkle root the spend proves against.
+ *   - The transcript shape, which does not vary with amount or recipient.
+ * - Public, as state:
+ *   - The commitment tree, the nullifier set, the issued-nonce set.
+ *   - No balances, no accounts, no supply.
+ * - Private:
+ *   - Amounts.
+ *   - Sender and recipient `pk`s.
+ *   - Nonces and secret keys.
+ *   - Which leaf a spend consumes.
```
andrew-fleming: If I understand correctly, wouldn't there be a leak if proving against a stale root? If so, it might be worth documenting some sort of caveat here

contracts/src/token/test/ConfidentialNoteFungibleToken.test.ts
```diff
@@ -0,0 +1,875 @@
...
+    expect(await nullifierCount()).toBe(0n);
+  });
+});
+
+// ---------------------------------------------------------------------------
+// _transfer: the conserving building block
+// ---------------------------------------------------------------------------
+
+describe('ConfidentialNoteFungibleToken: _transfer', () => {
+  let input: Note;
```
andrew-fleming: What do you think about testing this scenario?

- mint max
- transfer all
- burn all
- add a `_transfer` case whose outputs overflow 128 bits back to the input value (pinning that conservation compares at full width)

contracts/src/token/ConfidentialNoteFungibleToken.compact
andrew-fleming: > What do you think about including a tag here? My thinking is that the module claims unlinkability in the header doc...but it's using the same scheme as the global account id in the public modules

And if you agree, I'd include a regression test against `computeAccountId`

contracts/src/token/ConfidentialNoteFungibleToken.compact
```diff
@@ -0,0 +1,626 @@
...
+   * @returns {Bytes<32>} seed - A 32-byte cryptographically secure random
+   * value.
+   */
+  witness wit_ConfidentialNoteNonceRandomness(): Bytes<32>;
+
+  /**
+   * @description Mints a note of `value` to `recipientPk` with a core-derived
+   * nonce and returns it.
+   *
+   * @circuitInfo k=15, rows=17317
```
andrew-fleming:
```suggestion
   * @constraints k=15, rows=17317
```
We have to update all of these

PR #780, opened 2026-08-13
Improve proposal module

### LINKED ISSUE #828: Improve Proposal multisig module
Should include improvements for:
- Expired proposals can no longer be treated as active or executed
- Terminal proposal states remain preserved after cancellation or execution

andrew-fleming:
Fixes #828 

ShieldedMultiSig will be refactored separately. This PR proposes to just include minor adjustments to the preset in order for tests to pass, so it's technically correct just not complete atm. To be explicit, the following items are deferred to a separate PR specifically targeting shielded multisig:

- refactor caller auth mechanism
- add cap to expiry
- re-export ledger vals

## Summary by CodeRabbit

- **New Features**
  - Proposals now support configurable expiry deadlines.
  - Proposal status includes a new **Expired** state.
  - Status automatically reflects active, expired, executed, cancelled, or unknown proposals.
  - Proposal creation now requires a future expiry time.

- **Bug Fixes**
  - Expired proposals can no longer be treated as active or executed.
  - Terminal proposal states remain preserved after cancellation or execution.

- **Tests**
  - Added comprehensive coverage for expiry boundaries, lifecycle transitions, invalid deadlines, and backend time behavior.


contracts/test-utils/fixtures/blockTime.ts
```diff
@@ -0,0 +1,51 @@
...
+ * Sets the block time the dry backend reports to circuits. Persists across
+ * subsequent circuit calls on the same simulator.
+ *
+ * @param sim - A dry-backend simulator instance.
+ * @param secondsSinceEpoch - The chain time to report.
+ */
+export function setBlockTime(sim: unknown, secondsSinceEpoch: bigint): void {
+  const qc = queryContext(sim);
+  qc.block = { ...qc.block, secondsSinceEpoch };
+}
```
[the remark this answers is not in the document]
andrew-fleming: Fixed [3bfcf38](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/3bfcf38e80f5a4cf8d7ca5fab3105fe391340b0a)

pepebndc (approved):
LGTM

contracts/src/multisig/ProposalManager.compact
```diff
@@ -57,18 +62,69 @@ module ProposalManager {
...
   export ledger _proposals: Map<Uint<64>, Proposal>;
 
+  // ─── State Sentinels ────────────────────────────────────────────
+
+  /**
+   * @description The `Proposal.state` value marking a proposal as executed.
+   *
+   * @returns {Uint<64>} 1.
+   */
+  export pure circuit executedState(): Uint<64> {
```
0xisk, line 115: 🔴 **blocking:** the two new sentinel circuits (and the three whose proving cost changed: `_createProposal`, `assertProposalActive`, `getProposalStatus`) carry no `@circuitInfo k=…, rows=…`. The library applies that tag throughout — `Signer`, and every token / access / security module. Unblocked by annotating the 2 new + 3 modified circuits; a module-wide backfill (0/17 today) can ride with the deferred multisig PR.

contracts/src/multisig/ProposalManager.compact
```diff
@@ -57,18 +62,69 @@ module ProposalManager {
...
     to: Recipient,
     color: Bytes<32>,
     amount: Uint<128>,
-    status: ProposalStatus
+    state: Uint<64>
   }
 
   // ─── State ──────────────────────────────────────────────────────
 
   export ledger _nextProposalId: Counter;
```
0xisk, line 105: 🔴 **blocking:** both exported ledger fields are undocumented. `NonFungibleToken.compact:99-115` is the convention — `@description` plus `@type` lines for key and value. Worth stating here: `_nextProposalId` is a monotonic id source that is read back (so ids are never reused, and that read-back is what serializes concurrent creates), and `_proposals` is keyed by it and stores the overlaid `state` encoding. Both are readable off-chain without a proof, so integrators consume them directly.

contracts/src/multisig/ProposalManager.compact
```diff
@@ -34,16 +34,21 @@ module ProposalManager {
...
   export enum ProposalStatus {
     Inactive,
     Active,
     Executed,
-    Cancelled
+    Cancelled,
+    Expired
   }
 
   export enum RecipientKind {
```
0xisk, line 54: 🔴 **blocking:** `RecipientKind` and `Recipient` (line 60) are undocumented, while `ProposalStatus` and `Proposal` got thorough blocks in this PR. `RecipientKind` should say what each variant means and which rail accepts it — `toShieldedRecipient` rejects `UnshieldedUser`, `toUnshieldedRecipient` rejects `ShieldedUser`, and that mismatch is exactly the hazard `_createProposal` warns about. `Recipient` should state that `address` is raw 32 bytes whose interpretation depends on `kind`.

contracts/src/multisig/presets/ShieldedMultiSig.compact
```diff
@@ -56,7 +56,8 @@ export circuit deposit(coin: ShieldedCoinInfo): [] {
 export circuit createShieldedProposal(
```
0xisk, line 56: 🔴 **blocking:** `createShieldedProposal` has no doc comment, so this PR adds a **required** parameter to an integrator-facing circuit without stating its units (seconds since the Unix epoch), its constraints (must be in the future, must exceed the reserved sentinel range), or that the preset imposes no upper cap. The module documents all of it, but the preset is what dApps call. Unblocked by documenting `expiry`; the preset's other missing docs are pre-existing and out of scope here.

contracts/src/multisig/presets/ShieldedMultiSig.compact
0xisk, line 1: ❔ **question (non-blocking):** `_approveProposal` (line 130) pins `_proposalApprovals.member(id) == false` on the first approval, so two *first* approvers of one proposal collide on that read as well, independently of the `_approvalCount` cell. Is that worth folding into the same fix? Flagging it so the approval-count refactor removes both collision points rather than one.

0xisk, line 1: 🔵 **followup (pre-existing, for the deferred multisig refactor):** `_approveProposal` line 136 does `getApprovalCount(id) + 1` — a read-modify-write on `_approvalCount[id]`. The read pins the old count into the transaction, so two signers approving the same proposal in one block conflict: the loser is rejected at inclusion and has to rebuild, re-prove and resubmit. Concurrent approval is the normal multisig flow, and the same pinned read makes `executeShieldedProposal` conflict with a late approval landing in the same block.

The per-signer approval inserts are blind per-key writes and never conflict, so the count cell is the only collision point. One direction: drop `_approvalCount` and prove the threshold at execute time from the per-signer entries (K membership checks over caller-supplied approvers).

0xisk, line 1: 🔵 **followup:** the approval-concurrency limitation above isn't documented anywhere in the preset. There is a convention for this — `ConfidentialFungibleToken.compact:163` carries `@warning Known limitations.` / `Concurrency. Two transfers to the same recipient in the same block conflict. Documented limitation; acceptable for the v1 target use case`. The header here documents the `getCaller` / contract-signer limitation at length but is silent on approvals. Worth writing down even while the fix is deferred.

contracts/src/multisig/ProposalManager.compact
```diff
@@ -57,18 +62,69 @@ module ProposalManager {
...
     color: Bytes<32>,
     amount: Uint<128>,
-    status: ProposalStatus
+    state: Uint<64>
   }
 
   // ─── State ──────────────────────────────────────────────────────
 
   export ledger _nextProposalId: Counter;
   export ledger _proposals: Map<Uint<64>, Proposal>;
```
0xisk, line 106: 🔵 **followup:** lapsed proposals have no cleanup path. `_cancelProposal` routes through `assertProposalActive`, which rejects once the deadline passes, and there is no `_proposals.remove`, so every expired proposal keeps its entry and its original deadline forever and this map grows monotonically. The struct docs explain that the retained deadline is what tells a reader when it lapsed — worth stating the consequence too: no pruning circuit exists, growth is unbounded, and a consumer that needs cleanup must compose it. Adding a prune circuit is a separate design decision, not this PR.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- Refactor ProposalManager (#780)
```
0xisk, line 12: ⚪ **nitpick:** the breaking changes aren't marked, and this file uses a `**Breaking:**` prefix elsewhere (see the `_isInitialized` entry). Three integrator-facing breaks here: `Proposal.status` (enum) → `state: Uint<64>`, the new required `expiry` parameter on `_createProposal` / `createShieldedProposal`, and `getProposalStatus` returning `Inactive` instead of reverting for a missing id.

contracts/src/multisig/ProposalManager.compact
```diff
@@ -57,18 +62,69 @@ module ProposalManager {
...
+   *   proposal is active while the block time is below it and expired
+   *   afterwards.
+   *
+   * `_createProposal` rejects an `expiry` in the reserved range, so a proposal
+   * that has not reached a terminal state always has `state` greater than
+   * `cancelledState()`. Note that `Expired` is NOT terminal: a lapsed proposal
+   * still carries its original deadline, which is what tells a reader when it
+   * lapsed.
+   *
+   * @warning `state` is a DEADLINE, not a ready-at time. `state > blockTime`
```
0xisk, line 81: ⚪ **nitpick:** this reads awkwardly, and it states the comparison backwards from the code — `blockTimeLt(state)` is `blockTime < state`, so the subject is the block time, not `state`. It also contradicts the list just above it, where `state` is sometimes a sentinel rather than a deadline. Suggested rewrite:

```
@warning In the deadline range, `state` marks when a proposal STOPS being
executable, not when it starts: it is executable while `blockTime < state`
and inert once `blockTime >= state`.
```

contracts/src/multisig/ProposalManager.compact
```diff
@@ -173,23 +229,33 @@ module ProposalManager {
...
    * - Proposal must exist.
-   * - Proposal status must be Active.
+   * - Proposal must not be executed or cancelled.
+   * - Proposal must not have expired.
    *
    * @param {Uint<64>} id - The proposal ID.
    *
    * @returns {[]} Empty tuple.
    */
   export circuit assertProposalActive(id: Uint<64>): [] {
```
0xisk, line 247: ⚪ **nitpick:** worth one line here on how this failure differs from a state conflict. A conflicting read is transient — rebuild against fresh state, re-prove, and the transaction lands. `proposal expired` at inclusion is terminal: an integrator retry loop that treats all rejections alike will re-prove a dead proposal forever.

contracts/src/multisig/test/ProposalManager.test.ts
```diff
@@ -343,7 +495,150 @@ describe('ProposalManager', () => {
...
+      const id = await contract._createProposal(
+        recipient,
+        COLOR,
+        AMOUNT,
+        EXPIRY,
+      );
+      expect((await contract.getProposal(id)).state).toEqual(EXPIRY);
+
+      await contract._markExecuted(id);
+      expect((await contract.getProposal(id)).state).not.toEqual(EXPIRY);
```
0xisk, line 552: ⚪ **nitpick:** assert the exact value instead of the negation — `toEqual(contract.executedState())` pins which sentinel was written, where `.not.toEqual(EXPIRY)` would also pass for any other value.

contracts/src/multisig/test/ProposalManager.test.ts
```diff
@@ -343,7 +495,150 @@ describe('ProposalManager', () => {
...
+    });
+  });
+
+  // A consuming contract that imports `_proposals` by name can write it
+  // directly, bypassing every guard here. Compact modules encapsulate circuits,
+  // not state. These specs pin what that buys an attacker, so the hazard is
+  // recorded as behavior rather than only as a doc warning. They are not
+  // defended against; the module cannot defend against its own consumer.
+  describe('direct state writes by a consumer', () => {
```
0xisk, line 561: ⚪ **nitpick:** `state = 0` is the one reserved value these specs never decode — the group covers the sentinel forge (`1`) and backdated deadlines (`> 2`). `forceProposalState(id, 0n)` should read `Expired` and fail `assertProposalActive` with `proposal not active` (the sentinel check trips before the expiry one). One test completes the blast-radius set and pins which message that corner produces.

contracts/src/multisig/test/ProposalManager.test.ts
```diff
@@ -400,4 +715,302 @@ describe('ProposalManager', () => {
       );
     });
   });
+
+  // Needs control over the reported block time, which only the dry backend
+  // offers. See the live group below for what can be checked against a node.
+  describe.skipIf(isLiveBackend())('expiry (dry only)', () => {
```
0xisk, line 721: ⚪ **nitpick:** the rejections at `0` / `1` / `2` are pinned, but the smallest *legal* expiry is never accepted anywhere — the lowest accepted value in the suite is `NOW + 1` (1,000,001). `setBlockTime(2)` plus `create(3)` would pin the reserved range as exactly {0, 1, 2} from the acceptance side too, so a later widening to `>=` couldn't slip through on that side.

contracts/src/multisig/presets/ShieldedMultiSig.compact
```diff
@@ -198,6 +199,14 @@ export circuit getProposalStatus(id: Uint<64>): Proposal_ProposalStatus {
   return Proposal_getProposalStatus(id);
 }
 
+export pure circuit executedState(): Uint<64> {
```
0xisk, line 202: ⚪ **nitpick (if-minor):** one line each pointing at `Proposal.state`'s encoding would let an off-chain reader decode the field without opening the module. Low priority given the preset carries no docs today.

0xisk (commented):
_(removed)_

0xisk (commented):
Looking good @andrew-fleming! Left some blockings, questions and nitpicks comments. I'm still running the live testing locally.

0xisk (changes requested):

0xisk:
Confirmed all tests run successfully on a local live node on my end, Great work @andrew-fleming! :ok_hand: 
[live-proposal-manager.log](https://github.com/user-attachments/files/31226093/live-proposal-manager.log)
[live-shielded-multisig.log](https://github.com/user-attachments/files/31226094/live-shielded-multisig.log)

0xisk:
@andrew-fleming can you please create a tracking issue for this PR?

contracts/src/multisig/ProposalManager.compact
andrew-fleming: >  the two new sentinel circuits...carry no @circuitInfo k=…, rows=…

I disagree with adding `@circuitInfo` to the sentinels because…there’s no number to provide. The sentinels return literals so they fold at compile time. Wherever the sentinels are used, the cost for the op is the same as if the returned value was inlined. Documenting `k=5, rows=26` (the output with the hacky mock workaround with impuritizing circuits) would only document the mock scaffolding. lmk if you disagree

contracts/src/multisig/presets/ShieldedMultiSig.compact
andrew-fleming: Yes, I think we have to fold it into the same fix

andrew-fleming: I think I agree. Will confirm and/or discuss when tackling the refactor

andrew-fleming: Same response, will address in refactor

contracts/src/multisig/test/ProposalManager.test.ts
andrew-fleming: The suggestion creates a duplicate of the `should write the executed sentinel into state` test. Dropping it

contracts/src/multisig/presets/ShieldedMultiSig.compact
andrew-fleming: Agreed. Deferred for now

contracts/src/multisig/ProposalManager.compact
andrew-fleming: `@circuitInfo` is added to noted circuits (not the sentinels): [ff0a0f2](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/ff0a0f25f3fac66c7547019cfb600c050101c535)

contracts/src/multisig/ProposalManager.compact
andrew-fleming: fixed: [b66f939](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/b66f939ba71ca960622c96ee8787ed750d4ef941)

contracts/src/multisig/ProposalManager.compact
andrew-fleming: added: [245533a](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/245533a9958d3efa6112b537cd336e2ccac651e4)

contracts/src/multisig/presets/ShieldedMultiSig.compact
andrew-fleming: added: [a5ff168](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/a5ff1687c93e283a9e3ccc106731f5c18525dcf7)

contracts/src/multisig/ProposalManager.compact
andrew-fleming: added extra doc line on this: [2cda51c](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/2cda51c511e0348b45446d07b351bff66fa2fa6e)

CHANGELOG.md
andrew-fleming: added/fixed: [f6197af](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/f6197af0d322c70ee67c980a63d2e7c5e0917723)

contracts/src/multisig/ProposalManager.compact
andrew-fleming: agreed and fixed: [27f8db2](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/27f8db2e8bd721c8b5a4c91841132668092799f9)

contracts/src/multisig/ProposalManager.compact
andrew-fleming: added: [e0a0bea](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/e0a0beaeb04313eea4221b71da1b64fc2536df1a)

contracts/src/multisig/test/ProposalManager.test.ts
andrew-fleming: removed: [93e7d53](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/93e7d53c289df32747a3400f26f1c8da90e5e69b)

contracts/src/multisig/test/ProposalManager.test.ts
andrew-fleming: test added: [9c1ae09](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/9c1ae09569a294656c9637d4f4980bd22265d806)

contracts/src/multisig/test/ProposalManager.test.ts
andrew-fleming: test added: [5a2d8d9](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/5a2d8d92ad65f9b83726d1007fd0bac1c71a2822)

contracts/src/multisig/ProposalManager.compact
andrew-fleming: > NonFungibleToken.compact:99-115 is the convention — @description plus @type lines for key and value.

In the very same module referenced as the convention [nft:88-90](https://github.com/OpenZeppelin/compact-contracts/blob/main/contracts/src/token/NonFungibleToken.compact#L88-L90):

```
  /// Public state
  export sealed ledger _name: Opaque<"string">;
  export sealed ledger _symbol: Opaque<"string">;
```

I'm not opposed to documenting the ledger fields. My issue is if we are blocking PRs to enforce a convention, the convention should actually be the convention and applied consistently in the repo. https://github.com/OpenZeppelin/compact-contracts/pull/874 fixes this

contracts/src/multisig/test/ProposalManager.test.ts
```diff
@@ -400,4 +716,316 @@ describe('ProposalManager', () => {
...
+      contract = await fresh();
+      const anHourAgo = nowSeconds() - 3600n;
+      await expect(
+        contract._createProposal(
+          contract.shieldedUserRecipient(Z_RECIPIENT),
+          COLOR,
+          AMOUNT,
+          anHourAgo,
+        ),
+      ).rejects.toThrow();
```
0xisk, line 934: ⚪ **nitpick:** bare `rejects.toThrow()` accepts any error. Pin `'ProposalManager: expiry not in the future'`, or comment why the live failure surfaces with a different message.

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Changed (breaking)
```
0xisk, line 10: ⚪ **nitpick:** the 0.4.0-alpha.1 section marks breaks inline as `- **Breaking:** …` under `### Changed`, not with a separate heading. Also missing an `### Added` entry for `Expired`, the `executedState` / `cancelledState` circuits, and the `blockTime` fixture.

0xisk (approved):
LGTM! Thank you @andrew-fleming!

contracts/src/multisig/ProposalManager.compact
0xisk: > > the two new sentinel circuits...carry no @circuitInfo k=…, rows=…
> 
> I disagree with adding `@circuitInfo` to the sentinels because…there’s no number to provide. The sentinels return literals so they fold at compile time. Wherever the sentinels are used, the cost for the op is the same as if the returned value was inlined. Documenting `k=5, rows=26` (the output with the hacky mock workaround with impuritizing circuits) would only document the mock scaffolding. lmk if you disagree

I agree with you. :+1: TY!

contracts/src/multisig/ProposalManager.compact
0xisk: > My issue is if we are blocking PRs to enforce a convention, 
That actually got me hesitate a bit about marking this as blocking :) 

> the convention should actually be the convention and applied consistently in the repo. https://github.com/OpenZeppelin/compact-contracts/pull/874 fixes this

I agree thanks for opening this PR.

CHANGELOG.md
andrew-fleming: Haha I mentioned a few times in other PRs, we need to define what goes into the changelog because my changelog and your changelog are operating on different planets

contracts/src/multisig/test/ProposalManager.test.ts
andrew-fleming: My bad, fixed: [01e5cb8](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/01e5cb861c307a3fe002b737e3f444b3b5411747)

CHANGELOG.md
andrew-fleming: Added "Breaking" as the other entries: [89d743a](https://github.com/OpenZeppelin/compact-contracts/pull/780/commits/89d743a9380d1b096305cefbbc9d30ba34fade40)

0xisk (approved):
LGTM! Thank you @andrew-fleming!

0xisk:
@andrew-fleming lets get the conflicts fixed and merge this PR. TY!

0xisk (approved):

PR #818, opened 2026-08-21
Add domain to secretToScalar

### LINKED ISSUE #792: H-01 Account Identifiers And Encryption Secrets Share Undomain-Separated Hash
## H-01 Account Identifiers And Encryption Secrets Share Undomain-Separated Hash

**Severity:** High

---

[The account identifier derivation in `Utils`](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/utils/Utils.compact#L157-L159) hashes a secret key using `persistentHash` over a single-element vector of `Bytes<32>`. [The derivation of the ElGamal secret scalar](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/crypto/ElGamal.compact#L108-L110) applies the identical construction to its input and then truncates the result into the Jubjub scalar field. Neither includes a domain separation tag, so both compute the same function of the same input.

The first output is public. `ConfidentialFungibleToken` stores it as the [map key for balances, encryption keys, memo lists and escrow entries, all of which](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L267-L271) are readable on chain. The second output is the private key protecting every ciphertext held under that identifier. For any 32-byte value used in both roles, the private scalar is a truncation of a value published on chain.

The two secrets arrive through separate witnesses, [`wit_ConfidentialTokenSK` and `wit_ConfidentialTokenEK`](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L309-L317), and no circuit constrains them to differ. The condition for exposure is precise: the 32-byte value used as the encryption secret must never also be a value hashed into an account identifier. A wallet returning the same value from both witnesses satisfies that condition directly, and because the account identifier helper is shared across the access control and token modules, so does a wallet that reuses one secret as an identity in any of those modules and as the encryption secret here. In either case an observer reads the identifier from the ledger, applies the same truncation the circuit applies, and holds the account's encryption secret, which opens the spendable balance, the pending balance, both ciphertexts of every allowance the account granted or received, and every entry in its memo list, requiring no privileged access and no computation beyond one hash truncation.

Consider adding distinct domain separation tags to both derivations, so that a shared input produces unrelated outputs and reuse of a single secret across the two roles becomes harmless. Consider also documenting whether integrating wallets are permitted to derive the account secret and the encryption secret from common key material, and if so, stating that domain separation is what makes the practice safe.

andrew-fleming:
Fixes #792

## Summary by CodeRabbit

- **Security Enhancements**
  - Added domain separation to ElGamal secret-to-scalar and encryption-key derivation.
  - Prevented account identifiers from being directly reused to derive encryption scalars.
  - Centralized account-ID derivation for consistent behavior across modules.

- **Documentation**
  - Clarified hashing boundaries and safe reuse of secret material.

- **Tests**
  - Added coverage for domain-separation tags, argument ordering, and collision prevention.

### DIFF
--- a/contracts/src/crypto/ElGamal.compact
+++ b/contracts/src/crypto/ElGamal.compact
@@ -34,6 +34,13 @@ pragma language_version >= 0.23.0;
  * scalar. Feeding a raw `persistentHash` output into `ecMulGenerator` would
  * occasionally exceed the Jubjub scalar field order and fault at runtime.
  *
+ * @dev Domain separation. `persistentHash<Vector<N, Bytes<32>>>` is NOT
+ * domain-separating on its own. Two circuits hashing the same arity and
+ * element type share one hash domain, and the hash cannot tell which slot was
+ * "meant" as a tag. `secretToScalar` therefore prefixes a fixed constant, and
+ * that constant is placed FIRST so no call to `expandRandomness` (whose two
+ * slots are both caller-supplied) can reproduce the tuple by ordinary use.
+ *
  * @dev TRUST ASSUMPTION — subgroup membership. This is the module's most
  * load-bearing assumption. Every `JubjubPoint` reaching a curve operation is
  * assumed to be in the Jubjub prime-order subgroup, including points supplied by
@@ -102,11 +109,34 @@ module ElGamal {
    * @description Maps a 32-byte secret to a valid Jubjub scalar. See the
    * module-level hash-to-scalar note for why `degradeToTransient` is required.
    *
+   * @dev Domain separation here is load-bearing, not cosmetic. This output is
+   * the private key protecting every ciphertext held under an account. Public
+   * account identifiers elsewhere in the library are derived as
+   * `persistentHash([secretKey])` (`Utils.computeAccountId`), and that
+   * identifier is stored in the clear as a ledger map key. Untagged, this
+   * circuit would compute that same hash, so a secret used in both roles would
+   * have its encryption key recoverable from a published identifier by applying
+   * `degradeToTransient`. The tag makes the two derivations unrelated, so a
+   * wallet MAY safely derive its account secret and its encryption secret from
+   * common key material.
+   *
+   * @dev The tag is FIRST, deliberately. `expandRandomness` hashes
+   * `[seed, tag]` with both slots caller-supplied, so tagging this circuit as
+   * `[secret, tag]` would make `secretToScalar(x)` identical to
+   * `expandRandomness(x, <this tag>)` — reachable by passing this domain string
+   * to the parameter that exists to receive domain strings. Leading with the
+   * tag means reproducing this output through `expandRandomness` would require
+   * passing the domain constant as the *seed*. Do not normalise the orderings.
+   *
    * @param secret - The 32-byte secret to map.
    * @return A Field guaranteed to be a valid Jubjub scalar.
    */
   export pure circuit secretToScalar(secret: Bytes<32>): Field {
-    return degradeToTransient(persistentHash<Vector<1, Bytes<32>>>([secret]));
+    return degradeToTransient(
+             persistentHash<Vector<2, Bytes<32>>>(
+               [pad(32, "ElGamal:secretToScalar"), secret]
+             )
+           );
   }
 
   /**
--- a/contracts/src/crypto/ElGamal.compact
+++ b/contracts/src/crypto/ElGamal.compact
@@ -123,7 +123,7 @@ module ElGamal {
    * @dev The tag is FIRST, deliberately. `expandRandomness` hashes
    * `[seed, tag]` with both slots caller-supplied, so tagging this circuit as
    * `[secret, tag]` would make `secretToScalar(x)` identical to
-   * `expandRandomness(x, <this tag>)` — reachable by passing this domain string
+   * `expandRandomness(x, <this tag>)` reachable by passing this domain string
    * to the parameter that exists to receive domain strings. Leading with the
    * tag means reproducing this output through `expandRandomness` would require
    * passing the domain constant as the *seed*. Do not normalise the orderings.
--- a/contracts/src/crypto/ElGamal.compact
+++ b/contracts/src/crypto/ElGamal.compact
@@ -111,10 +111,10 @@ module ElGamal {
    *
    * @dev Domain separation here is load-bearing, not cosmetic. This output is
    * the private key protecting every ciphertext held under an account. Public
-   * account identifiers elsewhere in the library are derived as
-   * `persistentHash([secretKey])` (`Utils.computeAccountId`), and that
-   * identifier is stored in the clear as a ledger map key. Untagged, this
-   * circuit would compute that same hash, so a secret used in both roles would
+   * account identifiers derived from `Utils.computeAccountId`, for example,
+   * are computed as `persistentHash([secretKey])`, and that identifier
+   * is stored in the clear as a ledger map key. Untagged, this circuit
+   * would compute that same hash, so a secret used in both roles would
    * have its encryption key recoverable from a published identifier by applying
    * `degradeToTransient`. The tag makes the two derivations unrelated, so a
    * wallet MAY safely derive its account secret and its encryption secret from
--- a/contracts/src/crypto/test/ElGamal.test.ts
+++ b/contracts/src/crypto/test/ElGamal.test.ts
@@ -90,9 +90,9 @@ describe('ElGamal', () => {
     // both slots caller-supplied, so the collision cannot be removed.
     // Both flip if the order becomes `[secret, tag]`
     it('is not reproducible by passing the domain tag as expandRandomness tag', async () => {
-      expect(await contract.expandRandomness(EK_A, SECRET_TO_SCALAR_TAG)).not.toBe(
-        await contract.secretToScalar(EK_A),
-      );
+      expect(
+        await contract.expandRandomness(EK_A, SECRET_TO_SCALAR_TAG),
+      ).not.toBe(await contract.secretToScalar(EK_A));
     });
 
     it('is reproducible only with the domain tag in the seed slot', async () => {
--- a/contracts/src/crypto/test/ElGamal.test.ts
+++ b/contracts/src/crypto/test/ElGamal.test.ts
@@ -29,6 +29,9 @@ const b32 = (label: string): Uint8Array => {
 const EK_A = b32('elgamal-ek-A');
 const EK_B = b32('elgamal-ek-B');
 
+// Mirrors the `pad(32, ...)` tag `secretToScalar` prefixes its input with.
+const SECRET_TO_SCALAR_TAG = b32('ElGamal:secretToScalar');
+
 // Explicit encryption randomness. Any value below the Jubjub scalar field
 // order (~2^252) is a valid scalar; these small constants keep the tests
 // deterministic and let us assert that distinct randomness yields distinct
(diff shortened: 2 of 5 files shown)


contracts/src/crypto/ElGamal.compact
```diff
@@ -102,11 +109,34 @@ module ElGamal {
...
+   * passing the domain constant as the *seed*. Do not normalise the orderings.
+   *
    * @param secret - The 32-byte secret to map.
    * @return A Field guaranteed to be a valid Jubjub scalar.
    */
   export pure circuit secretToScalar(secret: Bytes<32>): Field {
-    return degradeToTransient(persistentHash<Vector<1, Bytes<32>>>([secret]));
+    return degradeToTransient(
+             persistentHash<Vector<2, Bytes<32>>>(
+               [pad(32, "ElGamal:secretToScalar"), secret]
```
0xisk, line 137: ### 1. Module-wide `@circuitInfo` rows are stale

🔴 **blocking:** This hash grew from 1 to 2 elements and is inlined in `register` (via `derivePk`) and every `assertDecryptsTo` caller, so their row counts shifted, yet the diff updates zero `@circuitInfo` annotations (#819 updated 4 of them for a comparable change). Wrong row/k numbers on a release RC mislead deploy planning against the block byte budget.

Unblock: re-measure the module and update the annotations.

contracts/src/utils/Utils.compact
```diff
@@ -150,6 +150,16 @@ module Utils {
    * ## ID Derivation
    * `accountId = persistentHash(secretKey)`
    *
+   * @dev The absence of a domain-separation tag is DELIBERATE. This identifier
+   * is global by design: the same key material yields the same identity in
+   * every module deriving one, so a user who wishes to carry one identity
+   * across modules can. Modules wanting a per-deployment, unlinkable identity
+   * should not use this circuit.
+   *
+   * @dev NOT FOR PRIVATE DERIVATION. This returns an identifier, not a secret,
```
0xisk, line 159: ### 2. Untagged `computeAccountId` is a deviation from the audit recommendation

❔ **question:** H-01 asked for tags on both derivations; this keeps the id untagged and guards the shared domain with a doc rule a future module can miss. Global identity needs one shared derivation, untagged is not required: tagging it (e.g. `OZ:accountId:v1`) keeps cross-module identity, satisfies the recommendation fully, and it's still alpha. Deliberate, or worth tagging now?

contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -27,20 +27,37 @@ const padTag = (s: string): Uint8Array => {
...
+ * that one is deliberately untagged and must stay so.
+ *
  * @note The field-element derivation from EK uses 31 bytes of the hash output
  * (empirically determined); the effective collision resistance is therefore 248 bits.
  */
 const derivePk = (ek: Uint8Array) => {
-  const rt_type = new CompactTypeVector(1, new CompactTypeBytes(32));
-  const ekHash = persistentHash(rt_type, [ek]);
+  const rt_type = new CompactTypeVector(2, new CompactTypeBytes(32));
+  const ekHash = persistentHash(rt_type, [SECRET_TO_SCALAR_TAG, ek]);
```
0xisk, line 54: ### 3. The newly permitted sk == ek sharing has no token-level test

🔴 **blocking:** The header now documents that a wallet may return one value from both witnesses, and integrators rely on that at the token level, where only the ElGamal unit test pins the math.

Unblock: register with EK == SK, read `CFT__encryptionKeys[accountId]` from public state, and assert it differs from `ecMulGenerator(convertBytesToField(31, accountId))`. Deterministic, and byte-identical (i.e. failing) on the pre-fix code.

contracts/src/utils/Utils.compact
andrew-fleming: Deliberate. A global id tag is a convention and not a security issue when we include a tag in `secretToScalar`. Changing the convention deserves its own discussion

contracts/src/crypto/ElGamal.compact
andrew-fleming: fixed [a9eedb4](https://github.com/OpenZeppelin/compact-contracts/pull/818/commits/a9eedb4bcd05aa82dd88e46eb28f5883e0cefcee)

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: added [ebbed40](https://github.com/OpenZeppelin/compact-contracts/pull/818/commits/ebbed4094b4bfa97399bde7120209ea8214acafd)

contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -154,6 +171,22 @@ describe.skipIf(isLiveBackend())(
         expect(storedPk).toEqual(expectedPk);
       });
 
+      it('keeps the encryption scalar unrelated to the public accountId when SK and EK are the same secret', async () => {
```
0xisk, line 168: ### 1. Full-flow sk == ek coverage is missing

🔴 **blocking:** The new header permits one secret for both witnesses, yet only `register` runs that way. `_debit`'s `assertDecryptsTo` (transfer, approve, transferFrom, burn) has no sk == ek test.

Unblock: `switchIdentity(sk, sk)`, register both parties, `_mint`, `transfer`, `sweep`, assert `balanceOf` decrypts to the expected values.

contracts/src/crypto/test/ElGamal.test.ts
```diff
@@ -61,6 +70,37 @@ describe('ElGamal', () => {
...
+      const accountId = persistentHash(
+        new CompactTypeVector(1, new CompactTypeBytes(32)),
+        [EK_A],
+      );
+
+      expect(await contract.secretToScalar(EK_A)).not.toBe(
+        convertBytesToField(31, accountId, 'attacker'),
+      );
+    });
+
```
0xisk, line 88: ### 2. Test comment overstates the residual collision

🔴 **blocking:** "the collision cannot be removed" is wrong. Tagging `expandRandomness` itself (arity 3, tag first) removes it outright, so the claim only holds for `secretToScalar` acting alone. As written the `toBe` test below pins the weakness as expected behavior, and a real fix later reads as a regression.

Unblock: reword to "accepted residual for now; removable by tagging `expandRandomness` itself", and rename the `toBe` test so it reads as a limitation rather than a guarantee.

contracts/src/crypto/ElGamal.compact
```diff
@@ -102,11 +109,34 @@ module ElGamal {
    * @description Maps a 32-byte secret to a valid Jubjub scalar. See the
    * module-level hash-to-scalar note for why `degradeToTransient` is required.
    *
+   * @dev Domain separation here is load-bearing, not cosmetic. This output is
+   * the private key protecting every ciphertext held under an account. Public
```
0xisk, line 113: ### 3. Tag position now differs across the library

🔵 **followup:** this mandates tag-first, while `ShieldedAccessControl.compact:744` hashes `[value, pad(32, "ShieldedAccessControl:nullifier")]` tag-last.

Write the rule down once (Utils or a crypto-conventions note) so a module copying SAC's shape for a private derivation does not recreate H-01.

contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -27,20 +27,37 @@ const padTag = (s: string): Uint8Array => {
...
+ * recoverable from the published identifier. See `buildAccountIdHash` below —
+ * that one is deliberately untagged and must stay so.
+ *
  * @note The field-element derivation from EK uses 31 bytes of the hash output
  * (empirically determined); the effective collision resistance is therefore 248 bits.
  */
 const derivePk = (ek: Uint8Array) => {
-  const rt_type = new CompactTypeVector(1, new CompactTypeBytes(32));
-  const ekHash = persistentHash(rt_type, [ek]);
+  const rt_type = new CompactTypeVector(2, new CompactTypeBytes(32));
```
0xisk, line 53: ### 4. Tests mirror the derivations in TS instead of calling the circuits

🔵 **followup:** the file already imports the `elgamal` pure circuits, and the simulator exposes `computeAccountId`. Calling `elgamal.derivePk(sk)` and `cft.computeAccountId(sk)` makes the test break when either real derivation drifts.

Non-blocking: the load-bearing assertion reads real ledger output, so coverage exists today.

contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -27,20 +27,37 @@ const padTag = (s: string): Uint8Array => {
 // Helpers
 // ---------------------------------------------------------------------------
 
+/**
+ * @description The domain-separation tag `ElGamal.secretToScalar` prefixes its
+ * input with, mirroring Compact's `pad(32, "ElGamal:secretToScalar")`. The tag
+ * is FIRST so no `expandRandomness([seed, tag])` call can reproduce the tuple.
+ */
+const SECRET_TO_SCALAR_TAG = (() => {
```
0xisk: ### 5. `padTag` already exists

⚪ **nitpick:** this reimplements `padTag` ten lines above. Use `padTag('ElGamal:secretToScalar')`.

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: fixed [c7a24a0](https://github.com/OpenZeppelin/compact-contracts/pull/818/commits/c7a24a0ed77100df82fadef00a73940cd75d0570)

contracts/src/crypto/test/ElGamal.test.ts
andrew-fleming: Good catch, fixed [79b11f1](https://github.com/OpenZeppelin/compact-contracts/pull/818/commits/79b11f128cc6594b5a6b7163bac6d8d393b61def)

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: fixed [d3fbecf](https://github.com/OpenZeppelin/compact-contracts/pull/818/commits/d3fbecfbaa5c10c927c643675466828b47afb4f9)

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: Not sure I agree with flagging this. We could only call the circuits, yes, but with that comes the risk of an uncaught typo in the tag that silently passes e.g. `secretToScalr` or even `fuck-oopsIWroteABadWord`. We want the tests to fail when there's a derivation drift. lmk if you disagree

contracts/src/crypto/ElGamal.compact
0xisk: @andrew-fleming will that be covered here?

contracts/src/token/test/ConfidentialFungibleToken.test.ts
0xisk: agree with you.

contracts/src/crypto/ElGamal.compact
andrew-fleming: Out of scope, let's keep the PR focused on addressing the fix. The rule for this implementation is already documented

contracts/src/crypto/ElGamal.compact
```diff
@@ -102,11 +109,34 @@ module ElGamal {
...
+   * account identifiers elsewhere in the library are derived as
+   * `persistentHash([secretKey])` (`Utils.computeAccountId`), and that
```
0xisk: nit: "account identifiers elsewhere in the library" reads library-wide, but only `Utils.computeAccountId` has this shape. `ShieldedAccessControl.computeAccountId` hashes `[secretKey, instanceSalt, pad(32, "ShieldedAccessControl:accountId")]`, so it is already tagged and salted and is not at risk here. Scoping the sentence to accounts derived via `Utils.computeAccountId` makes the threat precise.

contracts/src/crypto/ElGamal.compact
0xisk: Agree, out of scope. Filed as #854.

0xisk (approved):
Thank you @andrew-fleming!

contracts/src/crypto/ElGamal.compact
andrew-fleming: fixed: [94bcb19](https://github.com/OpenZeppelin/compact-contracts/pull/818/commits/94bcb19e8bf9a19d1e9c7083c8b3f6cab2e18e6d)

0xisk:
@andrew-fleming we need to solve conflict here again.

PR #819, opened 2026-08-21
Fix nonce reset with epoch

### LINKED ISSUE #799: L-01: Credit Randomness Nonce Can Be Reset By The Recipient
## L-01 Credit Randomness Nonce Can Be Reset By The Recipient

**Severity:** Low

---

The credit [path derives both its balance randomness and its memo ephemeral by hashing the recipient account together with the memo-list length](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L645-L675). The comment beside it explains that the length serves as a monotonic nonce so that each credit differs even when the wallet reuses its seed. This protection matters because the memo is an [additive one-time pad: a](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/crypto/EcdhMask.compact#L142-L159) repeated ephemeral to one recipient makes the difference of two stored memo values equal the difference of the two delivered amounts exactly, readable by anyone.

However, the length is not monotonic. [Pruning replaces the recipient's list with an empty one](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L995-L1001), after which the counter walks the same sequence a second time and each credit following the prune pairs with the credit that held the same count before it. Every input to the derivation must match for a pair to form, the seed among them, so credits from different senders never pair with one another and the party that gains is a passive observer rather than either participant. What such an observer obtains is a relation between amounts rather than an amount, though two equal amounts produce byte-identical entries and a single amount learned outside the system converts every connected difference into an exact figure.

[The module header records the reset](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L64-L71) and calls it an in-epoch backstop rather than a substitute for seed freshness. What it does not record is who ends an epoch. Pruning is available only to the account holder, so the counter guarding a sender's pad is advanced and reset by the recipient of that sender's transfers. The exposure still requires the sender to have reused a seed, so this describes the conditions under which a defence-in-depth measure stops functioning rather than a failure reachable against a correctly implemented wallet.

The addition worth noting is that the counter feeds both derived values and not only the memo ephemeral, so a revisited count repeats the balance randomness as well. Two such credits produce pending deltas that share an identical first component, which makes the reuse detectable without any secret, and whose second components differ by the amount difference expressed as a curve point.

Consider deriving the nonce from a per-account counter that only ever increases, stored independently of the memo list so that pruning cannot return it to an earlier value, which keeps the measure effective for exactly the wallets it exists to protect and removes a counterparty's influence over it. Moreover, consider describing in the seed guidance that the current counter is advanced by the recipient rather than by the sender whose randomness it protects.

andrew-fleming:
Fixes #799

### DIFF
--- a/contracts/src/token/ConfidentialFungibleToken.compact
+++ b/contracts/src/token/ConfidentialFungibleToken.compact
@@ -66,9 +66,11 @@ pragma language_version >= 0.23.0;
  * and the re-approve refund) derive randomness from a static domain tag with
  * no per-operation nonce, so two operations under one seed share randomness
  * and their public ciphertext deltas reveal the amount difference. The credit
- * path folds the recipient's memo-list length in as a nonce (defense in
- * depth), but that counter RESETS on `clearMemos`, so it is an in-epoch
- * backstop only, never a substitute for freshness.
+ * path folds the recipient's `_creditEpochs` counter in as a nonce (defense in
+ * depth): it advances on every credit, is never reset, and no counterparty can
+ * hold it back, so a repeated seed still yields distinct randomness there. It
+ * is a backstop, not a substitute for freshness. It covers only credits, and
+ * only against reuse of one seed to one recipient.
  *
  * Seed generation is a WALLET responsibility and OUT OF SCOPE for this
  * contract, which cannot enforce it. The witness shipped in this repo
@@ -269,6 +271,13 @@ module ConfidentialFungibleToken {
   export ledger _encryptionKeys: Map<Bytes<32>, JubjubPoint>;
   export ledger _memos: Map<Bytes<32>, List<EcdhMask_Ciphertext>>;
   export ledger _escrow: Map<Bytes<32>, Map<Bytes<32>, EscrowEntry>>;
+
+  // Per-account credit counter. `_credit` uses it as the nonce that keeps each
+  // credit's randomness distinct, so it MUST only ever increase. No circuit
+  // resets it. It is kept out of `_memos` for that reason: the memo list is
+  // prunable by its owner, and a nonce that could be walked back would let a
+  // reused seed repeat an earlier credit's randomness.
+  export ledger _creditEpochs: Map<Bytes<32>, Counter>;
 
   export ledger _isInitialized: Boolean;
 
@@ -642,20 +651,25 @@ module ConfidentialFungibleToken {
 
     const recipientPk = _encryptionKeys.lookup(disclose(account));
 
-    // Ensure the recipient's memo list exists; its length is a per-recipient
-    // monotonic nonce that makes this credit's randomness unique even if the
-    // wallet reuses its seed. Freshness is essential for the memo OTP: a repeated
-    // ephemeral to the same recipient would reuse the pad.
+    // Ensure the recipient's memo list exists.
     if (!_memos.member(disclose(account))) {
       _memos.insert(disclose(account), default<List<EcdhMask_Ciphertext>>);
     }
-    const count = _memos.lookup(disclose(account)).length() as Field;
+
+    // Advance the recipient's credit epoch and use it as this credit's nonce.
+    // The nonce must come from a source that only increases; anything the
+    // recipient can reset would let a reused seed repeat the randomness below.
+    if (!_creditEpochs.member(disclose(account))) {
+      _creditEpochs.insert(disclose(account), default<Counter>);
+    }
+    _creditEpochs.lookup(disclose(account)).increment(1);
+    const count = _creditEpochs.lookup(disclose(account)) as Field;
     const seed = wit_RandomnessSeed();
 
     // Randomness must be a VALID Jubjub scalar. `transientHash`'s raw Field output
     // can exceed the Jubjub scalar-field order and fault `ecMulGenerator`, so each
     // scalar is derived via `degradeToTransient(persistentHash(...))`, which
-    // truncates into the scalar field. The per-recipient count, folded into a
+    // truncates into the scalar field. The per-recipient epoch, folded into a
     // Bytes<32> nonce, keeps each credit's randomness unique even if the seed
     // repeats (essential for the memo OTP). This inlines the tag + expansion into
     // a single hash per scalar. The residual persistentHash cost is poseidon-gated
--- a/contracts/src/token/ConfidentialFungibleToken.compact
+++ b/contracts/src/token/ConfidentialFungibleToken.compact
@@ -67,10 +67,11 @@ pragma language_version >= 0.23.0;
  * no per-operation nonce, so two operations under one seed share randomness
  * and their public ciphertext deltas reveal the amount difference. The credit
  * path folds the recipient's `_creditEpochs` counter in as a nonce (defense in
- * depth): it advances on every credit, is never reset, and no counterparty can
- * hold it back, so a repeated seed still yields distinct randomness there. It
- * is a backstop, not a substitute for freshness. It covers only credits, and
- * only against reuse of one seed to one recipient.
+ * depth). A sender advances it with every credit and nothing rolls it back
+ * (neither the recipient nor any circuit) so the party whose randomness it
+ * protects is the one moving it, and a repeated seed still yields distinct
+ * randomness there. It is a backstop, not a substitute for freshness. It covers
+ * only credits, and only against reuse of one seed to one recipient.
  *
  * Seed generation is a WALLET responsibility and OUT OF SCOPE for this
  * contract, which cannot enforce it. The witness shipped in this repo
--- a/contracts/src/token/ConfidentialFungibleToken.compact
+++ b/contracts/src/token/ConfidentialFungibleToken.compact
@@ -707,7 +707,7 @@ module ConfidentialFungibleToken {
    * `extensions/ConfidentialFungibleTokenPublicSupply`) if the deployment tracks
    * `totalSupply`.
    *
-   * @circuitInfo k=15, rows=27142
+   * @circuitInfo k=15, rows=27179
    *
    * Requirements:
    *
@@ -754,7 +754,7 @@ module ConfidentialFungibleToken {
    * credits `to`, and pushes an encrypted memo to `to`. The amount is hidden;
    * the (sender, recipient) pair is public.
    *
-   * @circuitInfo k=16, rows=41617
+   * @circuitInfo k=16, rows=41654
    *
    * @notice Returns the caller's (sender's) `accountId` for caller-side gating.
    * The sender is derived from the same authentication `_debit` performs, so the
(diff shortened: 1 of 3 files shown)


contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -838,6 +838,48 @@ describe.skipIf(isLiveBackend())('ConfidentialFungibleToken: memos', () => {
...
     ).toBe(0n);
   });
+
+  // The credit nonce comes from `_creditEpochs`, which only ever increases. If
+  // it could return to an earlier value, a reused seed would repeat a credit's
+  // randomness and a repeated memo ephemeral reuses the one-time pad, making
+  // two equal amounts produce byte-identical entries. Both tests below hold the
+  // seed, recipient, and amount constant, so the nonce is the only thing that
+  // can distinguish the two credits.
+  it('produces distinct memos for equal amounts across a clearMemos', async () => {
```
0xisk, line 848: ### 1. Balance-randomness half of L-01 has no regression test

🔴 **blocking:** The audit notes the nonce feeds both derived values, yet only the memo surface is tested here. A refactor dropping `countHash` from the `rBalance` derivation alone would pass this suite and reintroduce half of L-01, and the PR closes #799 on merge.

Unblock: add a pending-side test. `_mint(10)` → capture `CFT__pending` → `sweep()` → `clearMemos()` → `_mint(10)` → assert the two ciphertexts differ. Deterministic, since `sweep` resets pending to `encryptZero` (identity points) and the test seed is fixed; on the pre-fix code the two ciphertexts are byte-identical.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -693,7 +708,7 @@ module ConfidentialFungibleToken {
    * `extensions/ConfidentialFungibleTokenPublicSupply`) if the deployment tracks
    * `totalSupply`.
    *
-   * @circuitInfo k=15, rows=27142
+   * @circuitInfo k=15, rows=27179
```
0xisk, line 720: ### 2. Stale `@circuitInfo` rows outside the credit path

❔ **question:** Only the 4 credit-path annotations changed, but a new ledger field can shift rows in circuits that never touch it, and CI doesn't validate the annotations. Was the whole module re-measured? `clearMemos` still reads `k=13, rows=2316`.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -270,6 +273,13 @@ module ConfidentialFungibleToken {
   export ledger _memos: Map<Bytes<32>, List<EcdhMask_Ciphertext>>;
   export ledger _escrow: Map<Bytes<32>, Map<Bytes<32>, EscrowEntry>>;
 
+  // Per-account credit counter. `_credit` uses it as the nonce that keeps each
+  // credit's randomness distinct, so it MUST only ever increase. No circuit
+  // resets it. It is kept out of `_memos` for that reason: the memo list is
+  // prunable by its owner, and a nonce that could be walked back would let a
+  // reused seed repeat an earlier credit's randomness.
+  export ledger _creditEpochs: Map<Bytes<32>, Counter>;
```
0xisk, line 290: ### 3. Lifetime credit count is now permanent public state

🔵 **followup:** Pruning used to reset the visible count; now state queries reveal each account's lifetime credit count forever. Tx-history observers learned this before either way, so the trade-off is sound and the audit asked for it. Add one line here recording it.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -270,6 +273,13 @@ module ConfidentialFungibleToken {
   export ledger _memos: Map<Bytes<32>, List<EcdhMask_Ciphertext>>;
   export ledger _escrow: Map<Bytes<32>, Map<Bytes<32>, EscrowEntry>>;
 
+  // Per-account credit counter. `_credit` uses it as the nonce that keeps each
```
0xisk, line 279: ### 4. `Counter` here does not buy commutativity

⚪ **nitpick:** The circuit reads the value back, so credits to one recipient still conflict (the header's concurrency limitation). Add a line: Counter is chosen for no-reset monotonicity, and credits now survive a concurrent `clearMemos` (the old length read didn't).

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -642,20 +652,25 @@ module ConfidentialFungibleToken {
...
-    // monotonic nonce that makes this credit's randomness unique even if the
-    // wallet reuses its seed. Freshness is essential for the memo OTP: a repeated
-    // ephemeral to the same recipient would reuse the pad.
+    // Ensure the recipient's memo list exists.
     if (!_memos.member(disclose(account))) {
       _memos.insert(disclose(account), default<List<EcdhMask_Ciphertext>>);
     }
-    const count = _memos.lookup(disclose(account)).length() as Field;
+
+    // Advance the recipient's credit epoch and use it as this credit's nonce.
```
0xisk, line 669: ### 5. "Epoch" names a concept the fix removed

⚪ **nitpick:** Epoch meant a memo-prune generation; the fix severs the counter from pruning, so this is a plain credit counter. Rename to `_creditNonces` or `_creditCounters` while it's still alpha API (`CFT__creditEpochs` ships in public state).

contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -838,6 +838,48 @@ describe.skipIf(isLiveBackend())('ConfidentialFungibleToken: memos', () => {
...
+    await cft.register();
+
+    const epoch = async () =>
+      (await cft.getPublicState()).CFT__creditEpochs.lookup(
+        ALICE.accountId,
+      ).read();
+
+    await cft._mint(ALICE.accountId, 10n);
+    const first = await epoch();
+    expect(first).toBeGreaterThan(0n);
```
0xisk: ### 6. Epoch test asserts a range where the value is known

⚪ **nitpick:** Values are exact on a fresh deploy: `toBe(1n)` here and `toBe(2n)` at line 881 instead of `toBeGreaterThan`; the exact values catch an increment-by-wrong-step bug the range check waves through.

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: Yes, they were re-measured and are correct

```
Compiling 14 circuits:
  circuit "_burn" (k=14, rows=14673)  
  circuit "_burnFrom" (k=16, rows=36274)  
  circuit "_mint" (k=15, rows=27179)  
  circuit "_move" (k=16, rows=41643)  
  circuit "allowance" (k=10, rows=938)  
  circuit "approve" (k=16, rows=43067)  
  circuit "balanceOf" (k=10, rows=989)  
  circuit "clearMemos" (k=13, rows=2316)  
  circuit "isRegistered" (k=9, rows=305)  
  circuit "pendingOf" (k=10, rows=986)  
  circuit "register" (k=13, rows=6203)  
  circuit "sweep" (k=13, rows=2879)  
  circuit "transfer" (k=16, rows=41654)  
  circuit "transferFrom" (k=16, rows=63244)  
Overall progress [====================] 14/14
```

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: Agreed. I think it'd be best to address this after the H and M issues are merged to avoid conflicts across PRs. Happy to open an issue if you agree

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: added [3d8c9f0](https://github.com/OpenZeppelin/compact-contracts/pull/819/commits/3d8c9f0810e274fa1c7a01a5d81302871eb8bba6)

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: added [e855cde](https://github.com/OpenZeppelin/compact-contracts/pull/819/commits/e855cde3e562dca1e82400c48aaf409b9203d702)

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: fixed [b9e6946](https://github.com/OpenZeppelin/compact-contracts/pull/819/commits/b9e694663917bbf379fc1dd455813912f9dbb40b)

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: added [e9364d8](https://github.com/OpenZeppelin/compact-contracts/pull/819/commits/e9364d88752dedc660a92ba255b63d7a5f761f23)

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -86,6 +89,11 @@ pragma language_version >= 0.23.0;
  *  - approve does not disclose the cap (private-cap escrow design).
  *  - Counterparty graph (sender_id, recipient_id) is public on every
  *    transfer.
+ *  - Per-account lifetime credit count is public and permanent
+ *    (`_creditEpochs`); unlike the memo list, it does not reset on
+ *    `clearMemos`.
+ *  - Per-account lifetime credit count is public and permanent
+ *    (`_creditEpochs`). It does not reset on `clearMemos`.
```
0xisk: ### 7. Disclosure bullet is duplicated in the module header

⚪ **nitpick:** the lifetime-credit-count bullet appears twice with near-identical wording (64e9d9c added one, e855cde the other). Keep the first, it carries the memo-list contrast, and delete this one.

contracts/src/token/ConfidentialFungibleToken.compact
0xisk: Agreed, filed #834 for after the H and M PRs land.

0xisk (approved):
LGTM! Thank you @andrew-fleming !

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: fixed [f13574c](https://github.com/OpenZeppelin/compact-contracts/pull/819/commits/f13574c1b40c5b9078924cdd091be4f98949ed63)

0xisk (approved):

0xisk (approved):

PR #821, opened 2026-08-23
Fix memo clearing

### LINKED ISSUE #793: H-02: Memo Clearing Is Unconditional and Cannot Exclude Unread Credits
## H-02 Memo Clearing Is Unconditional and Cannot Exclude Unread Credits

**Severity:** High

---

Incoming credit is delivered through a per-account [memo list, and the the channel a recipient's wallet scans](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L97-L112) and fold into its local plaintext cache. The need is structural, because the stored balance is exponential ElGamal and yields only a curve point on decryption, so the memo carries the one representation of the amount that the recipient can actually state. Each credit performs both halves of that delivery, [raising the pending ciphertext and appending the memo](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L677-L686). Pruning is offered through [`clearMemos`, which replaces the caller's entire list with an empty one and takes no](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L995-L1001) argument identifying which entries the caller has read.

The two halves of a credit can therefore be separated by an intervening prune:

1. A wallet reads its memo list and folds every entry into its durable cache locally.

2. The wallet submits `clearMemos`.

3. A credit to the same account is ordered first, raising the pending ciphertext and appending a memo the wallet has never seen.

4. `clearMemos` executes and replaces the list, discarding that memo.

5. The pending increase remains on chain with no surviving record of the amount it represents.

The circuit already [warns that clearing is destructive and tells callers to clear only memos they](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L981-L987) have already folded into a durable backup, but that instruction cannot be followed. Nothing in the call expresses which memos the caller intends to remove, so there is no expected count, no epoch and no list digest against which the circuit could reject a list that grew after the wallet last looked. A caller who does everything the warning asks still loses any memo appended between reading the list and the transaction executing, and the appending party is a third party sending value rather than the caller. This is an interaction between two independently valid transactions rather than a failed assertion inside one, so the guarantee that a reverting circuit leaves no state behind does not apply. That both remain valid follows from what each one reads. A credit reads the list contents, taking its length as a randomness nonce, so two credits to one recipient genuinely conflict, whereas `clearMemos` reads only whether the account has a list at all and then overwrites it with a constant, and appending an entry leaves that membership unchanged. Its recorded reads therefore still hold at inclusion even though the write discards what arrived in the meantime. Confirmation that a transaction recording only a membership check remains valid after another transaction changes that key's value would settle the behavior described here.

The credited amount is then stranded, because the recipient cannot name it and every spending path requires them to. Sweeping does not help and makes matters worse, since [`sweep` folds pending into spendable with no plaintext claim](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L835-L846), so it succeeds and merges an unknown quantity into the balance that debits draw from. From [that point the claim a debit demands](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L599-L607) can no longer be produced for any value, and the account's whole balance becomes unspendable rather than only the lost credit. A recipient can still search for a small missing amount by discrete logarithm against the point they can decrypt, so the outcome is recoverable for dust, but the [per-operation ceiling is now the full unsigned 128 bit range](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L285-L287), so a sender choosing a large amount places the value beyond any search, whether by accident or deliberately against a prune the sender can observe.

Consider making pruning state aware by having `clearMemos` accept the caller's view of the list, such as an expected entry count or a per-account epoch that increments on every append, and rejecting the call when it does not match current state, so that a list which grew since the caller last read it is never silently discarded. Consider also making pruning non destructive by removing a caller-specified prefix rather than replacing the list, so that entries arriving during the call survive it.

andrew-fleming:
Waiting on #819

Resolves #793

### DIFF
--- a/contracts/src/token/ConfidentialFungibleToken.compact
+++ b/contracts/src/token/ConfidentialFungibleToken.compact
@@ -66,9 +66,11 @@ pragma language_version >= 0.23.0;
  * and the re-approve refund) derive randomness from a static domain tag with
  * no per-operation nonce, so two operations under one seed share randomness
  * and their public ciphertext deltas reveal the amount difference. The credit
- * path folds the recipient's memo-list length in as a nonce (defense in
- * depth), but that counter RESETS on `clearMemos`, so it is an in-epoch
- * backstop only, never a substitute for freshness.
+ * path folds the recipient's `_creditEpochs` counter in as a nonce (defense in
+ * depth): it advances on every credit, is never reset, and no counterparty can
+ * hold it back, so a repeated seed still yields distinct randomness there. It
+ * is a backstop, not a substitute for freshness. It covers only credits, and
+ * only against reuse of one seed to one recipient.
  *
  * Seed generation is a WALLET responsibility and OUT OF SCOPE for this
  * contract, which cannot enforce it. The witness shipped in this repo
@@ -269,6 +271,13 @@ module ConfidentialFungibleToken {
   export ledger _encryptionKeys: Map<Bytes<32>, JubjubPoint>;
   export ledger _memos: Map<Bytes<32>, List<EcdhMask_Ciphertext>>;
   export ledger _escrow: Map<Bytes<32>, Map<Bytes<32>, EscrowEntry>>;
+
+  // Per-account credit counter. `_credit` uses it as the nonce that keeps each
+  // credit's randomness distinct, so it MUST only ever increase. No circuit
+  // resets it. It is kept out of `_memos` for that reason: the memo list is
+  // prunable by its owner, and a nonce that could be walked back would let a
+  // reused seed repeat an earlier credit's randomness.
+  export ledger _creditEpochs: Map<Bytes<32>, Counter>;
 
   export ledger _isInitialized: Boolean;
 
@@ -642,20 +651,25 @@ module ConfidentialFungibleToken {
 
     const recipientPk = _encryptionKeys.lookup(disclose(account));
 
-    // Ensure the recipient's memo list exists; its length is a per-recipient
-    // monotonic nonce that makes this credit's randomness unique even if the
-    // wallet reuses its seed. Freshness is essential for the memo OTP: a repeated
-    // ephemeral to the same recipient would reuse the pad.
+    // Ensure the recipient's memo list exists.
     if (!_memos.member(disclose(account))) {
       _memos.insert(disclose(account), default<List<EcdhMask_Ciphertext>>);
     }
-    const count = _memos.lookup(disclose(account)).length() as Field;
+
+    // Advance the recipient's credit epoch and use it as this credit's nonce.
+    // The nonce must come from a source that only increases; anything the
+    // recipient can reset would let a reused seed repeat the randomness below.
+    if (!_creditEpochs.member(disclose(account))) {
+      _creditEpochs.insert(disclose(account), default<Counter>);
+    }
+    _creditEpochs.lookup(disclose(account)).increment(1);
+    const count = _creditEpochs.lookup(disclose(account)) as Field;
     const seed = wit_RandomnessSeed();
 
     // Randomness must be a VALID Jubjub scalar. `transientHash`'s raw Field output
     // can exceed the Jubjub scalar-field order and fault `ecMulGenerator`, so each
     // scalar is derived via `degradeToTransient(persistentHash(...))`, which
-    // truncates into the scalar field. The per-recipient count, folded into a
+    // truncates into the scalar field. The per-recipient epoch, folded into a
     // Bytes<32> nonce, keeps each credit's randomness unique even if the seed
     // repeats (essential for the memo OTP). This inlines the tag + expansion into
     // a single hash per scalar. The residual persistentHash cost is poseidon-gated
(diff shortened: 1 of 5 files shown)


contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -976,28 +1002,47 @@ module ConfidentialFungibleToken {
...
     const accountId = _computeAccountId();
-    if (_memos.member(disclose(accountId))) {
-      _memos.insert(disclose(accountId), default<List<EcdhMask_Ciphertext>>);
-    }
+
+    // Both reads are unconditional, so both are pinned in the transcript and a
+    // concurrent credit invalidates this tx
+    assert(_memos.member(disclose(accountId)),
+           "ConfidentialFungibleToken: no memo list");
+    assert(_creditEpochs.lookup(disclose(accountId)) as Field == expectedEpoch as Field,
```
0xisk, line 1046: 🔴 **blocking:** The claim "a credit landing between the read and this transaction is never silently discarded" is only half-pinned: the stale-epoch test builds the tx *after* the credit (in-circuit assert), while the #793 sequence builds it *before* and relies on the pinned-read replay rejection, which no test exercises — and the audit text explicitly asked for that behavior to be confirmed. Unblock: capture the `clearMemos` transcript against pre-credit state, apply a credit, replay via `QueryContext.runTranscript` against the new state, and assert rejection (on pre-fix code the same replay succeeds and silently discards the memo); a live-backend two-tx test is the alternative recipe.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -86,6 +89,11 @@ pragma language_version >= 0.23.0;
  *  - approve does not disclose the cap (private-cap escrow design).
  *  - Counterparty graph (sender_id, recipient_id) is public on every
  *    transfer.
+ *  - Per-account lifetime credit count is public and permanent
+ *    (`_creditEpochs`); unlike the memo list, it does not reset on
+ *    `clearMemos`.
+ *  - Per-account lifetime credit count is public and permanent
+ *    (`_creditEpochs`). It does not reset on `clearMemos`.
```
0xisk: ⚪ **nitpick:** Duplicated bullet — the lifetime-credit-count disclosure appears twice with slightly different wording. It originates on #819's head, so fixing it there avoids re-inheriting it here on rebase.

contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -832,12 +837,99 @@ describe.skipIf(isLiveBackend())('ConfidentialFungibleToken: memos', () => {
...
 
-    await cft.clearMemos();
+    await cft.clearMemos(await currentEpoch(ALICE.accountId));
 
     expect(
       (await cft.getPublicState()).CFT__memos.lookup(ALICE.accountId).length(),
     ).toBe(0n);
   });
+
+  it('rejects a clearMemos whose epoch is stale', async () => {
```
0xisk, line 847: ⚪ **nitpick:** Extending this test with a successful retry at the new epoch would pin the documented recovery flow ("the wallet retries against the memos it has now seen") for two extra lines.

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: Agreed, fixed (with a tmp sim solution that can be improved with the sim)[03ca123](https://github.com/OpenZeppelin/compact-contracts/pull/821/commits/03ca1233779a4e98cc1211c37760f9ec7d595bfa)

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: added [a004ae1](https://github.com/OpenZeppelin/compact-contracts/pull/821/commits/a004ae109de2faab8ee59438988ec62392d91a19)

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: Yes, fixed in here [b8ecf41](https://github.com/OpenZeppelin/compact-contracts/pull/821/commits/b8ecf411b64fc282bb1d5b3eba264a44888b3daf)

PR #824, opened 2026-08-24
fix(token): require token existence in NonFungibleToken _approve

### LINKED ISSUE #798: M-04: _approve Permits Recording Approvals for Non-existent Tokens, Breaking Downstream Circuits
## M-04 `_approve` Permits Recording Approvals for Non-existent Tokens, Breaking Downstream Circuits

**Severity:** Medium

---

The exported [`_approve` circuit](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NonFungibleToken.compact#L838-L854) of `NonFungibleToken` performs its approver validation only when the `auth` argument is non-zero: the [token-existence check sits inside the](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NonFungibleToken.compact#L845) guarded [branch, while the approval write](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NonFungibleToken.compact#L853) executes unconditionally. The zero-`auth` path is documented as a deliberate escape hatch, and its requirements list no condition that `tokenId` exist.

The mock contract `MockNonFungibleToken`exports [the `_approve` circuit, bypassing the](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/test/mocks/MockNonFungibleToken.compact#L93-L99) generation of `auth` through `computeAccountId` [as implemented in approve circuit.](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NonFungibleToken.compact#L355) Consequently, following the mock contract implementation, an approval can be recorded for a token that does not exist, and the invariant that a populated approval implies an existing token no longer holds. That invariant is exactly what the module relies upon elsewhere: [a comment in `_unsafeTransferFrom` states that supplying an](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NonFungibleToken.compact#L502-L503) `auth` argument verifies token existence and that the returned previous owner therefore need not be checked against zero. Two exploits follow. In the first, an approval is planted on an unminted identifier and the approvee calls the public `transferFrom` or `_unsafeTransferFrom` with the zero account as the source: the update circuit reads the owner as zero, passes authorization on the stale approval, skips the balance decrement, credits the recipient and writes the owner entry, **minting the token without mint authorization** and permanently blocking the composer's own gated mint of that identifier. In the second, an approval is planted on an identifier about to be minted legitimately; because the update circuit clears approvals only when the source is non-zero, a mint never clears the approval, so the planted approval survives and the planter transfers the freshly minted token away from its owner.

Consider enforcing token existence independently of approval state. For example, make [_checkAuthorized](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NonFungibleToken.compact#L603-L613) revert for owner == Utils_zeroAccount() unconditionally (before evaluating approvals), and explicitly reject fromAddress == Utils_zeroAccount() [in transferFrom](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NonFungibleToken.compact#L458-L468) / [_unsafeTransferFrom. In addition, consider updating the mock contracts by](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NonFungibleToken.compact#L495-L509) removing all internal circuits that are not supposed to be exported to avoid integration issues for composers extending the contract.

0xisk:
## Types of changes

Fixes audit finding M-04: `NonFungibleToken._approve` recorded approvals for non-existent tokens on the zero-`auth` path.

## PR Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NFT transfer and approval validation for nonexistent tokens.
  * Prevented stale approvals and unauthorized minting through transfer flows.
  * Approval clearing now works safely when tokens no longer exist.

* **Testing**
  * Added coverage for approval authorization, token existence requirements, uninitialized circuits, and related regression scenarios.

### DIFF
--- a/contracts/src/token/NonFungibleToken.compact
+++ b/contracts/src/token/NonFungibleToken.compact
@@ -58,7 +58,7 @@ pragma language_version >= 0.23.0;
  * or in ledger writes. Canonicalization zeroes out the inactive branch of the Either,
  * ensuring that two values with the same active branch always resolve to the same map key
  * regardless of what data the inactive branch carries. Write paths are canonicalized in
- * `_update` (for `_owners` and `_balances`), `_approve` (for `_tokenApprovals`), and
+ * `_update` (for `_owners` and `_balances`), `_unsafeApprove` (for `_tokenApprovals`), and
  * `_setApprovalForAll` (for `_operatorApprovals`). Read paths are canonicalized in
  * `balanceOf`, `isApprovedForAll`, and `_isAuthorized`.
  *
@@ -338,7 +338,7 @@ module NonFungibleToken {
    * In the case of an external (non-contract) caller, the caller's identity is derived from the `wit_NonFungibleTokenSK`
    * witness as `persistentHash(secretKey)`.
    *
-   * @circuitInfo k=13, rows=3468
+   * @circuitInfo k=13, rows=3415
    *
    * Requirements:
    *
@@ -383,7 +383,7 @@ module NonFungibleToken {
    * In the case of an external (non-contract) caller, the caller's identity is derived from the `wit_NonFungibleTokenSK`
    * witness as `persistentHash(secretKey)`.
    *
-   * @circuitInfo k=13, rows=2936
+   * @circuitInfo k=13, rows=2947
    *
    * Requirements:
    *
@@ -439,7 +439,7 @@ module NonFungibleToken {
    * In the case of an external (non-contract) caller, the caller's identity is derived from the `wit_NonFungibleTokenSK`
    * witness as `persistentHash(secretKey)`.
    *
-   * @circuitInfo k=13, rows=4800
+   * @circuitInfo k=13, rows=4739
    *
    * Requirements:
    *
@@ -477,7 +477,7 @@ module NonFungibleToken {
    * are not currently supported. Tokens sent to a contract address may become irretrievable.
    * Once contract-to-contract calls are supported, this circuit may be deprecated.
    *
-   * @circuitInfo k=13, rows=4797
+   * @circuitInfo k=13, rows=4736
    *
    * Requirements:
    *
@@ -499,10 +499,13 @@ module NonFungibleToken {
                    ): [] {
     assertInitialized();
     assert(!Utils_isTargetZero(to), "NonFungibleToken: invalid receiver");
-    // Setting an "auth" argument enables the `_isAuthorized` check which verifies that the token exists
-    // (fromAddress != 0). Therefore, it is not needed to verify that the return value is not 0 here.
     const auth = left<Bytes<32>, ContractAddress>(_computeAccountId());
     const previousOwner = _update(to, tokenId, auth);
+
+    // The `_isAuthorized` check inside `_update` already implies the token exists, but that
+    // holds only while approvals are never recorded for unminted tokens. Assert it directly
+    // so a misused `_unsafeApprove` cannot turn this circuit into an unauthorized mint.
+    assert(!Utils_isTargetZero(previousOwner), "NonFungibleToken: nonexistent token");
 
     const canonFrom = Utils_canonicalize<Bytes<32>, ContractAddress>(fromAddress);
     assert(previousOwner == canonFrom, "NonFungibleToken: incorrect owner");
@@ -643,8 +646,8 @@ module NonFungibleToken {
 
     // Execute the update
     if (!Utils_isTargetZero(disclose(fromAddress))) {
-      // Clear approval. No need to re-authorize
-      _approve(Utils_zeroAccount(), tokenId, Utils_zeroAccount());
+      // Clear approval. No need to re-authorize or re-read the owner
+      _unsafeApprove(Utils_zeroAccount(), tokenId, Utils_zeroAccount(), false);
       const canonFrom = Utils_canonicalize<Bytes<32>, ContractAddress>(fromAddress);
       const newBalance = _balances.lookup(disclose(canonFrom)) - 1 as Uint<128>;
       _balances.insert(disclose(canonFrom), disclose(newBalance));
@@ -666,7 +669,7 @@ module NonFungibleToken {
   /**
    * @description Mints `tokenId` and transfers it to `to`.
    *
-   * @circuitInfo k=11, rows=1473
+   * @circuitInfo k=11, rows=1411
    *
    * Requirements:
    *
@@ -690,7 +693,7 @@ module NonFungibleToken {
   /**
    * @description Mints `tokenId` and transfers it to `to`. It does NOT check if the recipient is a ContractAddress.
    *
-   * @circuitInfo k=11, rows=1470
+   * @circuitInfo k=11, rows=1408
    *
    * Requirements:
    *
@@ -723,7 +726,7 @@ module NonFungibleToken {
    * The approval is cleared when the token is burned.
    * This circuit does not check if the sender is authorized to operate on the token.
    *
-   * @circuitInfo k=10, rows=509
+   * @circuitInfo k=10, rows=512
    *
    * Requirements:
    *
@@ -748,7 +751,7 @@ module NonFungibleToken {
    * interactions are supported in Compact. This restriction prevents assets from being inadvertently
    * locked in contracts that cannot currently handle token receipt.
    *
-   * @circuitInfo k=12, rows=2067
+   * @circuitInfo k=11, rows=2005
    *
    * Requirements:
    *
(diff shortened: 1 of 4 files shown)


contracts/src/token/test/nonFungibleToken.test.ts
```diff
@@ -818,6 +818,140 @@ describe('NonFungibleToken', () => {
...
+  // non-zero. That guard held the `_requireOwned` existence check, and the
+  // approval write sat outside it, so the zero-auth path recorded approvals
+  // for unminted tokens. Other circuits read the broken invariant back: a
+  // recorded approval implies the token exists.
+  //
+  // Fixed by splitting the circuit the way Solidity overloads it: `_approve`
+  // always requires existence and delegates to `_unsafeApprove`, whose
+  // `isExistenceRequired` flag mirrors Solidity's `emitEvent`. These tests
+  // failed on the unfixed code and pin the intended behaviour.
+  describe('audit M-04: approvals for nonexistent tokens', () => {
```
andrew-fleming: I don't think it's necessary to have an `audit` describe block or reference it with so much context with comments. I suggest favoring self-describing `describe` and `it` blocks and leave comments that help the reader understand what's happening (if it's complex and/or not obvious). In two months, these comments will be superfluous noise. lmk if you disagree

contracts/src/token/test/nonFungibleToken.test.ts
```diff
@@ -818,6 +818,140 @@ describe('NonFungibleToken', () => {
...
+      // without a stale approval.
+      await token
+        ._approve(SPENDER.either, NON_EXISTENT_TOKEN, ZERO_ACCOUNT)
+        .catch(() => undefined);
+
+      // On the unfixed code the stale approval satisfied `_isAuthorized`, so
+      // `_checkAuthorized` never reached its nonexistent-token assert:
+      // `_update` read the owner as zero, skipped the balance decrement,
+      // credited SPENDER and wrote the owner entry, leaving SPENDER holding a
+      // token nobody minted and blocking the composer's own gated mint of it.
```
andrew-fleming: Is the context on what the previous code was necessary to rehash and explain? I ask bc it's not a complex test and the `it` block + the variable names tell the story. If we want to include comments here, maybe something like this:

```
// Plant approval on nonexistent id
...

// Attempt to mint nonexistent token
...
```

contracts/src/token/test/nonFungibleToken.test.ts
```diff
@@ -818,6 +818,140 @@ describe('NonFungibleToken', () => {
...
+        ._approve(SPENDER.either, TOKENID_1, ZERO_ACCOUNT)
+        .catch(() => undefined);
+
+      // `_update` clears approvals only when the source is non-zero, so a
+      // mint leaves any planted approval standing.
+      await token._mint(OWNER.either, TOKENID_1);
+      expect(await token.getApproved(TOKENID_1)).toEqual(ZERO_ACCOUNT);
+
+      // On the unfixed code SPENDER then took OWNER's token on the strength
+      // of the surviving approval.
```
andrew-fleming: Forgive me: again, not a fan of leaving "unfixed code" relics

contracts/src/token/NonFungibleToken.compact
```diff
@@ -833,6 +837,43 @@ module NonFungibleToken {
...
+   * @param {Either<Bytes<32>, ContractAddress>} auth - An account authorized to operate on all tokens held by the owner of the token
+   * @param {Boolean} isExistenceRequired - Whether `tokenId` must exist
+   * @return {[]} - Empty tuple.
+   */
+  export circuit _unsafeApprove(
+                   to: Either<Bytes<32>, ContractAddress>,
+                   tokenId: Uint<128>,
+                   auth: Either<Bytes<32>, ContractAddress>,
+                   isExistenceRequired: Boolean
+                   ): [] {
```
andrew-fleming, line 876: I suspect that this circuit is anticipatory for extensions? When should users use `_unsafeApprove` vs `_approve`?

andrew-fleming (commented):
Looking good, Isk! I left a few nit-ish comments on the tests

The one concern though is that both the base branch and the target branch should be the release branch. The target branch can easily be changed, not the base branch though. Not sure if it's worth moving this back a few commits or opening a new PR

contracts/src/token/NonFungibleToken.compact
andrew-fleming: Discussed offline

contracts/src/token/test/nonFungibleToken.test.ts
0xisk: Fixed in 940c6f6a

contracts/src/token/test/nonFungibleToken.test.ts
0xisk: Fixed in 940c6f6a

contracts/src/token/test/nonFungibleToken.test.ts
0xisk: Fixed in 940c6f6a

contracts/src/token/test/nonFungibleToken.test.ts
```diff
@@ -816,6 +816,116 @@ describe('NonFungibleToken', () => {
...
+      expect(await token._getApproved(NON_EXISTENT_TOKEN)).toEqual(
+        ZERO_ACCOUNT,
+      );
+    });
+
+    it('should not approve a nonexistent token into existence', async () => {
+      // Plant an approval on a nonexistent id
+      await token
+        ._approve(SPENDER.either, NON_EXISTENT_TOKEN, ZERO_ACCOUNT)
+        .catch(() => undefined);
```
andrew-fleming: Two things here:
1. This is catching "undefined", but it's really swallowing the "nonexistent token" error. It's odd and a little misleading to leave it as such
2. This test never actually plants the approval bc it fails (why the catch is misleading). The test at line 909 already does this correctly

andrew-fleming: If you agree with those points, namely the second, we can probably remove this test entirely

contracts/src/token/test/nonFungibleToken.test.ts
```diff
@@ -816,6 +816,116 @@ describe('NonFungibleToken', () => {
...
+
+      expect(await token._ownerOf(NON_EXISTENT_TOKEN)).toEqual(ZERO_ACCOUNT);
+      expect(await token.balanceOf(SPENDER.either)).toEqual(0n);
+    });
+
+    it('should not approve a token before it is minted', async () => {
+      // Plant an approval on the id OWNER is about to mint
+      await token
+        ._approve(SPENDER.either, TOKENID_1, ZERO_ACCOUNT)
+        .catch(() => undefined);
```
andrew-fleming: Same thing here, nothing's planted

contracts/src/token/test/nonFungibleToken.test.ts
```diff
@@ -816,6 +816,116 @@ describe('NonFungibleToken', () => {
...
+
+      // `_update` clears approvals only when the source is non-zero, so a
+      // planted approval would survive the mint
+      await token._mint(OWNER.either, TOKENID_1);
+      expect(await token.getApproved(TOKENID_1)).toEqual(ZERO_ACCOUNT);
```
andrew-fleming: Though this test is probably worth including with `_unsafeApprove` e.g.

```
_unsafeApprove(...)
_mint(...)
getApproved(...)
```

andrew-fleming (commented):
Changes look good! I left a final few comments regarding tests. Afterwards, we should be good to go

contracts/src/token/test/nonFungibleToken.test.ts
0xisk: Agree — removed in be3786c4.

contracts/src/token/test/nonFungibleToken.test.ts
0xisk: Agree — removed in 6b0f28ee.

contracts/src/token/test/nonFungibleToken.test.ts
0xisk: Agree — added in 6b0f28ee.

andrew-fleming (approved):
LGTM!

PR #825, opened 2026-08-24
Fix escrow owner memo nonce

### LINKED ISSUE #796: M-02: Escrow Owner Memo Encrypted Under Randomness Controlled by Spender
## M-01 Escrow Owner Memo Encrypted Under Randomness Controlled by Spender

**Severity:** Medium

---

An allowance is recorded as an [escrow entry holding two ciphertexts and an owner-readable memo. Both ciphertexts encode the amount in the exponent, so neither party can cheaply](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L241-L255) recover the number, and the memo is therefore the owner's only channel for learning the live remaining allowance. That memo is an ECDH one-time pad addressed to the owner's public key. Because the recipient key is public ledger state, the ephemeral is the only secret standing [between an observer and the plaintext, and the `EcdhMask` module states this requirement and assigns it](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/crypto/EcdhMask.compact#L22-L38): the ephemeral must be secret and unpredictable rather than merely distinct, and the caller supplies it and owns both properties.

In the escrow spend path that assignment breaks. [The ephemeral is expanded from a seed supplied by the caller, but the caller at that moment is the](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L1088-L1091) **spender** rather than the **owner**. The expansion is a deterministic public hash, so possession of the seed is possession of the [ephemeral, and the value protected belongs to a party absent](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L1101-L1103) from the call who contributes no entropy to it.

Two routes reach the plaintext: - A spender who publishes the seed allows any observer to recompute the mask and read the owner's remaining allowance directly from public state. - A spender who reuses one seed across two spends against the same owner repeats the pad, so the difference of the two stored memos equals the difference of the two plaintexts and requires no secret at all.

The second route requires no intent, leaves the spender unaware, and is the behavior of the only witness present in the repository. The same construction in `approve` is sound precisely because the caller there is the owner.

The design already models the spender as hostile for the integrity of this field, since the remaining amount is recomputed in-circuit so that the spender cannot forge it. Its confidentiality is nevertheless left to that same spender's discretion, and the owner has no remedy. An owner who rotates a fresh high-entropy seed on every call they make still leaks, because the leaking call is not theirs, and nothing in escrow state records that a pad repeated. The authority conferred by `approve` extends to spending up to the cap and not to disclosing it, and the disclosure survives revocation. Consider folding a per-escrow monotonic counter into the memo ephemeral. The addition worth noting is that the [`_credit` path already applies exactly this mechanism](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L645-L652), folding the recipient's memo-list length into its randomness so that each credit remains unique even when the seed repeats, and recording in its comment that the freshness is essential for the memo one-time pad. The remedy is therefore not a new construction but the one used a single function away. Applying it removes the accidental repetition that carries most of the practical exposure, though no construction prevents a spender who deliberately publishes the ephemeral, since any construction must let the spender compute the pad. It is further advisable to state which party each confidentiality property protects, so that a wallet author supplying the seed understands that the hygiene guards a counterparty rather than themselves.

andrew-fleming:
Add `_escrowSpendEpochs`, a per-(owner, spender) counter folded into all three derivations (`rSpender`, `rOwner` and `eOwnerMemo`), so each spend is uniquely randomized regardless of the seed. It's kept out of `EscrowEntry` because approve replaces that entry wholesale and keyed to the pair `_spendEscrow` already reads so no new state is pinned

Resolves #796 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved escrow spending so repeated transactions generate fresh encrypted outputs and owner memo randomness.
  - Ensured randomness remains unique across repeated spends, including transactions with unchanged or changing remaining balances.
  - Preserved randomness independently of replaceable escrow entries.

- **Tests**
  - Added coverage for repeated transfers and burns from escrow, including equal and changing remainder scenarios.

### DIFF
--- a/contracts/src/token/ConfidentialFungibleToken.compact
+++ b/contracts/src/token/ConfidentialFungibleToken.compact
@@ -65,27 +65,30 @@ pragma language_version >= 0.23.0;
  * it. The debit and re-approve-refund ciphertexts (`_debit`, `approve`) derive
  * randomness from a static domain tag with no per-op nonce, so two
  * operations under one seed share randomness and their public ciphertext deltas
- * reveal the amount difference. `_spendEscrow` is the exception. Every value it
- * derives folds in `_escrowSpendEpochs`, so its randomness is unique per spend
- * regardless of the seed. The credit path folds the recipient's memo-list length
- * in as a nonce (defense in depth), but that counter RESETS on `clearMemos`, so it
- * is an in-epoch backstop only, never a substitute for freshness.
+ * reveal the amount difference. `_spendEscrow` is the exception: every value it
+ * derives folds in the spend epoch AND both counterparties, so its randomness
+ * is unique per spend regardless of the seed. The credit path folds the
+ * memo-list length in as a nonce (defense in depth), but that counter RESETS on
+ * `clearMemos`, so it is an in-epoch backstop only, never a substitute for
+ * freshness.
  *
  * @notice WHOSE confidentiality a seed protects is not always the caller's. On
  * most paths a wallet's seed hygiene guards its own amounts, so the incentive
- * and the exposure sit with the same party. The escrow owner memo is the
+ * and the exposure sit with the same party. The escrow spend path is the
  * exception: `_spendEscrow` expands every random value it uses from the
  * SPENDER's seed, but all three protect the OWNER. Two re-encrypt their escrow
  * and the third masks their remaining allowance. A spender reusing a seed leaks
  * a counterparty's balance, not their own, and the owner cannot prevent it by
- * rotating their own seed as the leaking call is not theirs. `_escrowSpendEpochs`
- * removes the accidental case by making each ephemeral unique regardless of the
- * seed; nothing can stop a spender who deliberately publishes it, since any
- * construction must let the spender compute the pad. Only the seed and the epoch
- * are needed to make the pad unique: the mask is `kdf(e * ownerPk)`, so it
- * already varies with the owner and the preimage doesn't need to name them. Wallet
- * authors supplying a seed for `transferFrom` / `burnFrom` should understand they
- * are protecting the owner they are spending on behalf of.
+ * rotating their own seed as the leaking call is not theirs.
+ *
+ * The accidental case is removed by binding both counterparties into the
+ * derivation alongside the epoch, which counts per (owner, spender) and so
+ * restarts for each pair. That covers the two ways one seed repeats randomness
+ * across calls: one spender drawing on two owners, and two wallets sharing a
+ * seed drawing on one owner. Nothing stops a spender who deliberately publishes
+ * their seed, since any construction must let them compute the pad. Wallet
+ * authors supplying a seed for `transferFrom` / `burnFrom` should understand
+ * they are protecting the owner they are spending on behalf of.
  *
  * Seed generation is a WALLET responsibility and OUT OF SCOPE for this
  * contract, which cannot enforce it. The witness shipped in this repo
@@ -191,15 +194,15 @@ pragma language_version >= 0.23.0;
  * into the derivations is a possible defense-in-depth hardening; under review, deferred.
  *
  * Block weight. A deploy bundles every exported entry point's on-chain IR into
- * one transaction, and the value circuits are SHA256-heavy (k=16), so a
- * full-surface deployment is byte-heavy. Whether it fits is a property of the
+ * one transaction, and the value circuits are SHA256-heavy (k=16, k=17 for
+ * `transferFrom`), so a full-surface deployment is byte-heavy. Whether it fits is a property of the
  * target network, not the contract: the per-transaction block byte budget is a
  * governance-configurable, node-side ledger parameter. On the local ledger-8 dev
  * stack the full surface exceeded it and was rejected (`1010: Transaction would
  * exhaust the block limits`), while a reduced surface (e.g. receive + transfer,
  * dropping allowances and supply accounting) fit. So confirm deployability
  * against the target network; if a deployment is over budget there, the lever is
- * exposing only the k=16 value operations it needs, not just trimming getters.
+ * exposing only the value operations it needs, not just trimming getters.
  * A deployed contract can also be upgraded by its Contract Maintenance Authority
  * (CMA), so an over-budget deployment can start with a lean surface and add
  * operations by upgrade instead of fitting the whole surface into one deploy.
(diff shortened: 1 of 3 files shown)


contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -1086,9 +1122,23 @@ module ConfidentialFungibleToken {
...
+    if (!_escrowSpendEpochs.lookup(disclose(fromAddress)).member(disclose(spenderId))) {
+      _escrowSpendEpochs.lookup(disclose(fromAddress)).insert(disclose(spenderId), default<Counter>);
+    }
+    _escrowSpendEpochs.lookup(disclose(fromAddress)).lookup(disclose(spenderId)).increment(1);
+    const spendEpoch =
+      _escrowSpendEpochs.lookup(disclose(fromAddress)).lookup(disclose(spenderId)) as Field;
+
+    const rSpender   = _expandSpendRandomness(seed, pad(32, "spend_escrow_spender"), spendEpoch);
+    const rOwner     = _expandSpendRandomness(seed, pad(32, "spend_escrow_owner"),   spendEpoch);
+    const eOwnerMemo = _expandSpendRandomness(seed, pad(32, "spend_owner_memo"),     spendEpoch);
```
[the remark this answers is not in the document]
andrew-fleming: The initial approach I pushed was an attempt to keep k from exceeding 16. I fixed the changes and went with binding the owner/spender pair to keep things simple even though k is now at 17 in `transferFrom`. Something to consider improving in the future 

[170b7e7](https://github.com/OpenZeppelin/compact-contracts/pull/825/commits/170b7e74ffa5bb24beb5ce5037d0ab6b39a02d0d)

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -936,7 +980,7 @@ module ConfidentialFungibleToken {
    * caller via `approve`) and credits it to `to`, reducing both escrow copies by
    * `value`. The amount is hidden.
    *
-   * @circuitInfo k=16, rows=63207
+   * @circuitInfo k=17, rows=70164
```
0xisk: ### `transferFrom` optimization

**followup:** the five-field `persistentHash(seed, tag, owner, spender, epoch)` costs one more SHA-256 block per scalar than the two-field `expandRandomness` it replaces, so `transferFrom` goes 63207 → 70164 rows and over the k=16 boundary. k=17 doubles its proving table, and the header two hundred lines up already warns the full CFT surface exceeded the local ledger-8 block byte budget.

This PR introduces the derivation, so it is not pre-existing debt. Hashing `(owner, spender, epoch)` once into a 32-byte nonce and expanding that three times should hold k=16; worth measuring before this lands on the release branch rather than tracking it after.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -62,13 +62,33 @@ pragma language_version >= 0.23.0;
...
- * path folds the recipient's memo-list length in as a nonce (defense in
- * depth), but that counter RESETS on `clearMemos`, so it is an in-epoch
- * backstop only, never a substitute for freshness.
+ * it. The debit and re-approve-refund ciphertexts (`_debit`, `approve`) derive
+ * randomness from a static domain tag with no per-op nonce, so two
+ * operations under one seed share randomness and their public ciphertext deltas
+ * reveal the amount difference. `_spendEscrow` is the exception: every value it
+ * derives folds in the spend epoch AND both counterparties, so its randomness
+ * is unique per spend regardless of the seed. The credit path folds the
+ * memo-list length in as a nonce (defense in depth), but that counter RESETS on
```
0xisk: ### Credit-nonce reset wording collides with #819 / #821

🔵 **followup:** this paragraph still says the credit nonce RESETS on `clearMemos`, but #819 and #821 replace that with `_creditEpochs`. Not fixable here, since the correct wording depends on which of the three merges first. Whichever lands last reconciles the sentence, in the header and in the `wit_RandomnessSeed` doc.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -270,6 +304,15 @@ module ConfidentialFungibleToken {
...
   export ledger _escrow: Map<Bytes<32>, Map<Bytes<32>, EscrowEntry>>;
 
+  // Per-(owner, spender) escrow-spend counter, folded into the randomness
+  // `_spendEscrow` derives so repeat spends by one spender against one owner
+  // never share a one-time pad. It restarts for each pair, which is why both
+  // counterparties are bound in too; see `SpendRandomnessPreimage`.
+  //
+  // Kept out of `EscrowEntry`: `approve` replaces that entry wholesale, so a
+  // counter living there would restart on every re-approve.
+  export ledger _escrowSpendEpochs: Map<Bytes<32>, Map<Bytes<32>, Counter>>;
```
0xisk, line 340: ### New public state deserves a one-line justification

⚪ **nitpick:** `_escrowSpendEpochs` is new persisted public state, a per-pair spend count. It leaks nothing beyond what tx history already shows, but saying so in the ledger comment saves the next reader asking whether it is a disclosure.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -174,15 +194,15 @@ pragma language_version >= 0.23.0;
  * into the derivations is a possible defense-in-depth hardening; under review, deferred.
  *
  * Block weight. A deploy bundles every exported entry point's on-chain IR into
- * one transaction, and the value circuits are SHA256-heavy (k=16), so a
- * full-surface deployment is byte-heavy. Whether it fits is a property of the
+ * one transaction, and the value circuits are SHA256-heavy (k=16, k=17 for
+ * `transferFrom`), so a full-surface deployment is byte-heavy. Whether it fits is a property of the
```
0xisk: ### Header line over 100 chars

⚪ **nitpick:** rewrap. Biome does not format `.compact`, so nothing catches this in CI.

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: added [faa90af](https://github.com/OpenZeppelin/compact-contracts/pull/825/commits/faa90afc628d0fe0824ce5402fed6b53433193c7)

0xisk:
@andrew-fleming we also need to solve conflicts here :)

Issue #804, opened 2026-08-21
L-06: clearMemos Does Not Return the Authenticated Identifier


andrew-fleming:
## L-03 `clearMemos` Does Not Return the Authenticated Identifier

**Severity:** Low

**Source:** Midnight Foundation #02 — OpenZeppelin Compact Contracts Audit, Release 0.3.0-alpha.1 (2026-08-18)

---

The module makes caller-side gating sound by having each caller-authenticating circuit return the identifier it authenticated. The [header enumerates them as](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L149-L152) `register`, `_burn`, `transfer`, `_move`, `approve`, `transferFrom`, `sweep` and `_burnFrom`. [`clearMemos`](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L995-L1001) authenticates a caller in exactly the same way, deriving the account from `wit_ConfidentialTokenSK`, but returns `[]` and is absent from that list.

[...]

PR #830, opened 2026-08-27
Require ek in sweep and clearMemos

### LINKED ISSUE #797: M-03: Pruning and Sweeping Authenticate the Account Secret Without The Encryption Secret
## M-03 Pruning and Sweeping Authenticate the Account Secret Without The Encryption Secret

**Severity:** Medium

---

An attacker who obtains only an account's identity secret, and not its encryption secret, can leave that account permanently unable to spend. The module keeps the two secrets [separate by design](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L309-L317), the first fixing which account a caller is and the second protecting what that account can read, and every circuit that moves value proves both, as [proving the balance against the registered encryption key](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L599-L607). Two exported circuits prove [only the first. Pruning replaces the caller's memo list after deriving the identifier and nothing else, and](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L995-L1001) `sweep` folds [pending value into the spendable balance using homomorphic addition, which needs no](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L835-L846) plaintext and therefore proves no knowledge of the key, its registration check confirming only that a key exists. Holding the identity secret alone is enough to call both, and calling both in order freezes the account:

1. A third party credits the account. The amount lands in the pending pool and a memo [carrying it is appended to the account's list, which is the only channel through which the holder learns that figure, the](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L97-L112) stored balance itself decrypting only to a curve point.

2. The attacker prunes before the wallet has read that memo. The credited amount is now unknown to the holder.

3. The attacker sweeps. The unknown amount is added into the spendable balance.

4. The holder attempts any debit or approval. Each requires stating the plaintext of the spendable balance, which is now the previous balance plus an amount they cannot determine, so no claim satisfies the assertion and nothing can be moved.

Either step alone is survivable. Pruning by itself strands the credit in the pending pool while leaving the spendable balance intact and usable, and sweeping by itself is the ordinary operation the holder performs anyway. It is the pair that converts one unreadable credit into an account that cannot transact, and both halves are reachable with the same single secret. A small missing amount can still be recovered by searching for a discrete [logarithm, but the per- operation ceiling is the full unsigned 128 bit range](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L285-L287), so a large credit places it beyond any search.

The exposure assumes the two secrets hold different values. A wallet is permitted to derive both from the same material, and where it does an attacker holding one holds the other and these checks would change nothing. That is an argument for enforcing the separation wherever it is relied upon rather than against the separation itself.

Consider requiring an encryption secret witness in both circuits, asserted to re-derive the public key registered for the account, before either the memo list is replaced or pending value is merged, so that authority over what an account can read matches authority over what it can spend. Adding the check to only one of the two leaves the sequence above intact. Consider also making pruning remove an explicitly identified prefix rather than the whole list, which bounds what a caller holding one secret can destroy.

andrew-fleming:
Resolves #797

### DIFF
--- a/contracts/src/token/ConfidentialFungibleToken.compact
+++ b/contracts/src/token/ConfidentialFungibleToken.compact
@@ -42,6 +42,8 @@ pragma language_version >= 0.23.0;
  *
  *  - `wit_ConfidentialTokenEK`: verified by `ElGamal_assertDecryptsTo`,
  *    which re-derives `pk` and asserts it equals the on-chain stored pk.
+ *    `sweep` and `clearMemos` have no plaintext to decrypt, so they assert the
+ *    same binding directly via `_assertOwnsEncryptionKey`.
  *
  *  - `wit_PlaintextBalance(ct)`: verified by `ElGamal_assertDecryptsTo`,
  *    which asserts `Dec(ct, EK) == claimedValue`. The ciphertext is passed
@@ -819,7 +821,7 @@ module ConfidentialFungibleToken {
    * change, which is what would otherwise invalidate the victim's in-flight spend
    * proofs (a liveness grief).
    *
-   * @circuitInfo k=13, rows=2879
+   * @circuitInfo k=13, rows=5688
    *
    * @notice Purely homomorphic: adds the two ciphertexts (same key) and resets
    * pending to Enc(0). No plaintext is needed; the wallet already learned the
@@ -835,8 +837,7 @@ module ConfidentialFungibleToken {
   export circuit sweep(): Bytes<32> {
     assertInitialized();
     const accountId = _computeAccountId();
-    assert(_encryptionKeys.member(disclose(accountId)),
-           "ConfidentialFungibleToken: not registered");
+    _assertOwnsEncryptionKey(accountId);
     const newSpendable = ElGamal_add(
       _balances.lookup(disclose(accountId)),
       _pending.lookup(disclose(accountId)));
@@ -976,7 +977,7 @@ module ConfidentialFungibleToken {
    * @description Clears the caller's memo list, letting a wallet that has folded
    * its memos into its local balance cache prune on-chain memo storage.
    *
-   * @circuitInfo k=13, rows=2316
+   * @circuitInfo k=13, rows=5127
    *
    * @warning Clearing memos is destructive and can permanently lock funds. The
    * on-chain balance ciphertext is not directly decryptable (discrete-log
@@ -995,6 +996,7 @@ module ConfidentialFungibleToken {
   export circuit clearMemos(): [] {
     assertInitialized();
     const accountId = _computeAccountId();
+    _assertOwnsEncryptionKey(accountId);
     if (_memos.member(disclose(accountId))) {
       _memos.insert(disclose(accountId), default<List<EcdhMask_Ciphertext>>);
     }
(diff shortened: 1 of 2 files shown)


contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -1026,6 +1032,26 @@ module ConfidentialFungibleToken {
...
+   * comparing it to ledger state.
+   *
+   * @notice Circuits that move or destroy value prove this implicitly, via
+   * `ElGamal_assertDecryptsTo` on a balance. `sweep` and `clearMemos` have no
+   * plaintext to decrypt, so they assert it directly: holding the account secret
+   * alone must not be enough to clear a memo list or merge pending value, since
+   * doing both leaves the holder unable to state a balance they can no longer
+   * learn.
+   */
+  circuit _assertOwnsEncryptionKey(accountId: Bytes<32>): [] {
```
0xisk, line 1047: 🔵 **followup:** This check presumes sk and ek hold independent values; a wallet deriving ek from sk voids it (holding one means holding both). #797 accepts that premise, and the root cause is tracked in #792 (H-01). Cross-reference only, nothing to change here.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -989,12 +992,15 @@ module ConfidentialFungibleToken {
...
    * - Contract is initialized.
+   * - The caller is registered.
+   * - The caller holds the encryption secret registered for their account.
    *
    * @return {[]} - Empty tuple.
    */
   export circuit clearMemos(): [] {
     assertInitialized();
     const accountId = _computeAccountId();
+    _assertOwnsEncryptionKey(accountId);
```
0xisk, line 1003: ❔ **question:** `clearMemos` is now caller-authenticating (proves both secrets) but still returns `[]` and is absent from the module header's caller-identity-gating list, so a wrapper cannot gate it. That is #804 (L-06) — is it planned for this release branch, since it changes an exported signature?

contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -840,6 +840,79 @@ describe.skipIf(isLiveBackend())('ConfidentialFungibleToken: memos', () => {
...
+        CHARLIE.encryptionKey,
+      );
+      await expect(cft.clearMemos()).rejects.toThrow('not registered');
+      await expect(cft.sweep()).rejects.toThrow('not registered');
+    });
+
+    it('leaves the account spendable after a blocked prune-and-sweep', async () => {
+      await asAttacker();
+      await expect(cft.clearMemos()).rejects.toThrow();
+      await expect(cft.sweep()).rejects.toThrow();
```
0xisk: ⚪ **nitpick:** `rejects.toThrow()` without a message. The specific messages are pinned two tests up, but pinning them here too keeps the assertion deterministic if the `beforeEach` setup drifts.

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: > ...is it planned for this release branch...

If the plan is to resolve all issues on the release branch, then yes. L-06 should be fixed with its own PR though

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: whoops, fixed [b861c32](https://github.com/OpenZeppelin/compact-contracts/pull/830/commits/b861c32c75ec22ee41c7b92412379143de524466)

contracts/src/token/ConfidentialFungibleToken.compact
0xisk: Confirmed against #818, which is now ready to merge. Its header note is scoped to confidentiality, so it does not conflict with this guard, but an integrator reading it would not know sharing forfeits the split. Filed as #855.

contracts/src/token/ConfidentialFungibleToken.compact
0xisk: Agree, own PR. Worth landing after #821, which already changes this signature to take `expectedEpoch`.

0xisk (approved):
Thank you @andrew-fleming!

PR #831, opened 2026-08-27
Fix balance width

### LINKED ISSUE #795: M-01: Missing Running Total Balance Check Can Lock Balances Permanently
## M-02 Missing Running Total Balance Check Can Lock Balances Permanently

**Severity:** Medium

---

Balances are held as ElGamal ciphertexts and grow [by homomorphic addition, incoming value being folded into the pending pool by the credit path and later](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L677-L681) [merged into the spendable balance by `sweep`. The only ceiling applied](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L835-L846) anywhere [is a per-operation check on each individual](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L1015-L1018) value, [which is the full unsigned 128-bit maximum. Both](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L285-L287) growth operations reduce to [component-wise curve addition](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/crypto/ElGamal.compact#L273-L277), the credit path reaching it through [a helper that adds a fresh encryption of the incoming amount and](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/crypto/ElGamal.compact#L328-L335) `sweep` calling it directly on the two stored ciphertexts. Because the amount sits in the exponent, that addition accumulates in the scalar field of the subgroup rather than in the argument type, so a stored ciphertext can come to represent a quantity above the unsigned 128-bit range while remaining perfectly well formed.

No running total is held anywhere for a check to apply to. The accumulated quantity exists only as the exponent of a curve point, which no participant in a credit can read, since a sender holds neither the recipient's encryption secret nor their current plaintext. The quantity that would need bounding is therefore not available to be bounded at the point where it grows.

1. Credits accumulate into one account, each within the per-operation bound but summing past the unsigned 128-bit range.

2. The stored ciphertext remains valid and continues to decrypt, since the sum is taken in the scalar field of the subgroup, which is far larger.

3. The holder attempts a debit, an approval, or an escrow spend, each of which requires them to state the current plaintext and have it checked against the stored ciphertext.

4. No value of the witness type decrypts to the stored quantity, so the assertion cannot be satisfied by any claim and every one of those paths fails.

The balance stays intact on chain with no operation able to move it, affecting [the debit path,](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L599-L607) [`approve`](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L897-L899) [and the escrow spend alike. The](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/ConfidentialFungibleToken.compact#L1083-L1084) restriction is enforced in the constraint system and not merely by the compiler, the emitted representation carrying an explicit 128 bit range constraint on that witness, so a prover willing to submit an out of range claim is refused as well. No privileged role is needed to arrive there, since ordinary transfers between users accumulate into one recipient's pending pool, and the amounts required are large rather than unusual in kind.

Consider treating a bounded total supply as the invariant to enforce, either by requiring a [supply layer whose existing overflow check on the tracked total is made mandatory rather than](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/extensions/ConfidentialFungibleTokenPublicSupply.compact#L53-L58) optional, with the trade-off of mint and burn amounts being publicly determined through total supply delta, or by lowering the per-operation ceiling far enough that no plausible sequence of credits reaches the scalar field, and by documenting that a deployment without a bounded total can reach an unrecoverable balance. Consider carefully before instead range checking the plaintext as pending is merged into spendable, because the pending pool is written by third parties: a sender able to push a recipient past the ceiling would then make that recipient's `sweep` unsatisfiable and block their incoming value indefinitely, which relocates the failure onto the party being paid and defeats the purpose of separating pending from spendable. The check cannot be moved earlier into the credit path either, since a sender has no way to learn the recipient's current plaintext.

andrew-fleming:
Fixes [#795](https://github.com/OpenZeppelin/compact-contracts/issues/795)

### DIFF
--- a/contracts/src/crypto/ElGamal.compact
+++ b/contracts/src/crypto/ElGamal.compact
@@ -460,7 +460,7 @@ module ElGamal {
    * @warning The claim domain must stay below the Jubjub subgroup order `l`:
    * verification compares `g^claimedValue`, so two claims congruent mod `l` are
    * indistinguishable. `Uint<248>` clears `l` (~2^251.86) by under 4 bits, and
-   * that width is set by Compact's cap rather than by this bound — 251 is the
+   * that width is set by Compact's cap rather than by this bound. 251 is the
    * widest sound `Uint` if the cap ever rises.
    *
    * Requirements:
--- a/contracts/src/crypto/ElGamal.compact
+++ b/contracts/src/crypto/ElGamal.compact
@@ -457,6 +457,12 @@ module ElGamal {
    * range constraint belongs to where the value is declared, not to this
    * signature.
    *
+   * @warning The claim domain must stay below the Jubjub subgroup order `l`:
+   * verification compares `g^claimedValue`, so two claims congruent mod `l` are
+   * indistinguishable. `Uint<248>` clears `l` (~2^251.86) by under 4 bits, and
+   * that width is set by Compact's cap rather than by this bound — 251 is the
+   * widest sound `Uint` if the cap ever rises.
+   *
    * Requirements:
    *
    * - `ek` is the secret for `pk`.
--- a/contracts/src/crypto/ElGamal.compact
+++ b/contracts/src/crypto/ElGamal.compact
@@ -452,6 +452,11 @@ module ElGamal {
    * wrong plaintext fails the decryption check. Lifted special case of
    * `assertDecryptsToPoint` with the message point `g^claimedValue`.
    *
+   * @notice `claimedValue` is the widest `Uint` Compact offers, so this parameter
+   * never bounds a caller. A narrower value costs proportionally fewer rows: the
+   * range constraint belongs to where the value is declared, not to this
+   * signature.
+   *
    * Requirements:
    *
    * - `ek` is the secret for `pk`.
@@ -466,7 +471,7 @@ module ElGamal {
                    ct: Ciphertext,
                    pk: JubjubPoint,
                    ek: Bytes<32>,
-                   claimedValue: Uint<128>
+                   claimedValue: Uint<248>
                    ): [] {
     assertDecryptsToPoint(ct, pk, ek, ecMulGenerator(claimedValue as Field));
   }
--- a/contracts/src/token/ConfidentialFungibleToken.compact
+++ b/contracts/src/token/ConfidentialFungibleToken.compact
@@ -92,7 +92,10 @@ pragma language_version >= 0.23.0;
  * memo delivers the amount directly (no discrete-log recovery), so the former
  * `2^48` cap that kept wallet-side BSGS tractable is gone. Amounts stay well under
  * the Jubjub scalar-field order, so the homomorphic balance sums (and the
- * `g^value` encoding) do not wrap.
+ * `g^value` encoding) do not wrap at any plausible scale. Accumulation makes
+ * that a resource bound rather than an invariant: ~2^124 maximal credits into
+ * one account reach `l`, past which the stored point encodes the sum MOD `l` and
+ * the balance verifies against a smaller number than was deposited.
  *
  * @dev Balance claims are `Uint<248>`, wider than the per-transfer bound. A
  * balance grows by homomorphic addition with no running total to bound, so a
--- a/contracts/src/token/ConfidentialFungibleToken.compact
+++ b/contracts/src/token/ConfidentialFungibleToken.compact
@@ -92,7 +92,16 @@ pragma language_version >= 0.23.0;
  * memo delivers the amount directly (no discrete-log recovery), so the former
  * `2^48` cap that kept wallet-side BSGS tractable is gone. Amounts stay well under
  * the Jubjub scalar-field order, so the homomorphic balance sums (and the
- * `g^value` encoding) do not wrap. Balances remain Uint<128>.
+ * `g^value` encoding) do not wrap.
+ *
+ * @dev Balance claims are `Uint<248>`, wider than the per-transfer bound. A
+ * balance grows by homomorphic addition with no running total to bound, so a
+ * long enough run of in-range credits could otherwise push the stored quantity
+ * past what any `Uint<128>` claim can state, leaving the balance unspendable.
+ * The 120-bit gap between the claim width and the transfer bound puts that at
+ * ~2^120 credits. The two widths are therefore coupled: raising
+ * `MAX_TRANSFER_VALUE` narrows the gap, and 248 is the widest `Uint` Compact
+ * offers, so the claim cannot absorb it.
  *
  * @dev Value delivery via memo channel. Every credit made through `_credit`
  * (mint / transfer / transferFrom) pushes an ECDH one-time-pad memo (see
@@ -290,8 +299,9 @@ module ConfidentialFungibleToken {
    * well under the Jubjub scalar-field order, so homomorphic sums do not wrap.
    *
    * @notice This equals the argument type's own maximum, so the per-op check is
-   * effectively a no-op, retained as a single explicit ceiling / documentation
-   * point. Removing the now-vestigial check entirely is a follow-up cleanup.
+   * redundant today. KEEP IT: it compares against a hard constant rather than
+   * the argument type, so it still catches a later widening of the transfer
+   * type — which is what the balance-claim width is sized against.
    */
   export pure circuit MAX_TRANSFER_VALUE(): Uint<128> {
     return 340282366920938463463374607431768211455; // 2^128 - 1
@@ -334,7 +344,7 @@ module ConfidentialFungibleToken {
    * @param ct - The ciphertext whose plaintext is being requested.
    * @returns The wallet's cached plaintext for `ct`.
    */
-  witness wit_PlaintextBalance(ct: ElGamal_Ciphertext): Uint<128>;
+  witness wit_PlaintextBalance(ct: ElGamal_Ciphertext): Uint<248>;
 
   /**
    * @witness wit_RandomnessSeed
@@ -716,7 +726,7 @@ module ConfidentialFungibleToken {
    * `extensions/ConfidentialFungibleTokenPublicSupply`) if the deployment tracks
    * `totalSupply`.
    *
-   * @circuitInfo k=14, rows=14673
+   * @circuitInfo k=14, rows=14823
    *
    * Requirements:
    *
@@ -740,7 +750,7 @@ module ConfidentialFungibleToken {
    * credits `to`, and pushes an encrypted memo to `to`. The amount is hidden;
    * the (sender, recipient) pair is public.
    *
-   * @circuitInfo k=16, rows=41617
+   * @circuitInfo k=16, rows=41767
    *
    * @notice Returns the caller's (sender's) `accountId` for caller-side gating.
    * The sender is derived from the same authentication `_debit` performs, so the
@@ -778,7 +788,7 @@ module ConfidentialFungibleToken {
    * operations (`mint`/`burn`) live in a separate layer outside this conserving
    * surface, and only they may break conservation.
    *
-   * @circuitInfo k=16, rows=41606
+   * @circuitInfo k=16, rows=41756
    *
    * @notice Like `transfer`, the credit lands in the recipient's pending pool
    * (see the dual-balance note); the caller sweeps their own incoming value.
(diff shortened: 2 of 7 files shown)


contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -1102,6 +1103,49 @@ describe.skipIf(isLiveBackend())(
...
+        DECIMALS,
+      );
+      await cft.privateState.switchIdentity(
+        ALICE.secretKey,
+        ALICE.encryptionKey,
+      );
+      await cft.register();
+    });
+
+    it('spends a balance accumulated past the per-transfer bound', async () => {
```
0xisk, line 1259: 🔴 **blocking:** The audit names debit, approval, and escrow spend as the locked paths; this covers only the `_debit` site (via `_burn`), leaving the `approve` claim site (`ConfidentialFungibleToken.compact:908`) unpinned. Extend it: cache `2n * MAX128`, `approve(BOB, MAX128)`, then complete with `transferFrom`.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -1080,8 +1090,14 @@ module ConfidentialFungibleToken {
...
     const spenderPk = _encryptionKeys.lookup(disclose(spenderId));
-    const escrowPlaintext = wit_PlaintextBalance(entry.spenderCt);
-    ElGamal_assertDecryptsTo(entry.spenderCt, spenderPk, ek, escrowPlaintext);
+    // Balance claims are `Uint<248>`, but an escrow cannot legitimately exceed
+    // the transfer bound: `approve` caps it at `Uint<128>` and replaces rather
+    // than stacks, and `_spendEscrow` only reduces it. Narrowing states that
+    // invariant here instead of leaving it to inspection of those two circuits.
+    // The cast fails closed, so an out-of-range claim aborts rather than wraps.
+    const escrowPlaintext =
+      wit_PlaintextBalance(entry.spenderCt) as Uint<128>;
```
0xisk, line 1152: 🔴 **blocking:** "The cast fails closed" has no pinning test. Cache `(1n << 128n) + escrowValue` for the spender's escrow ct and expect `transferFrom` to throw — a truncating cast would make it succeed silently (the truncated value equals the real escrow), so that test discriminates exactly.

contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -1102,6 +1103,49 @@ describe.skipIf(isLiveBackend())(
...
+      expect(total).toBeGreaterThan(MAX128);
+
+      await cft.privateState.cachePlaintext(
+        await cft.balanceOf(ALICE.accountId),
+        total,
+      );
+      await cft._burn(MAX128);
+    });
+
+    it('keeps the transfer bound at the Uint<128> maximum', async () => {
```
0xisk, line 1333: 🔴 **blocking:** This pins the bound's value but not the widths on the compiled circuit: narrowing the witness to e.g. `Uint<136>` passes both new tests while collapsing the headroom from 2^120 to 2^8. Assert from `MockConfidentialFungibleToken`'s `contract-info.json` that `wit_PlaintextBalance`'s result `maxval` is 2^248-1 and `transfer`'s `value` arg is 2^128-1 (parse bigint-aware — both exceed double precision).

contracts/src/crypto/ElGamal.compact
```diff
@@ -466,7 +471,7 @@ module ElGamal {
                    ct: Ciphertext,
                    pk: JubjubPoint,
                    ek: Bytes<32>,
-                   claimedValue: Uint<128>
+                   claimedValue: Uint<248>
```
0xisk, line 480: 🔴 **blocking:** `MockElGamal.compact:116` still declares `claimedValue: Uint<128>` — it compiles (implicit widening) but misstates this signature and makes a >2^128 claim untestable at the ElGamal level. Widen the wrapper and add one unit test claiming `2n * MAX128` against `addCiphertexts(encrypt(pk, MAX128, r1), encrypt(pk, MAX128, r2))`.

contracts/src/crypto/ElGamal.compact
```diff
@@ -452,6 +452,11 @@ module ElGamal {
    * wrong plaintext fails the decryption check. Lifted special case of
    * `assertDecryptsToPoint` with the message point `g^claimedValue`.
    *
+   * @notice `claimedValue` is the widest `Uint` Compact offers, so this parameter
```
0xisk, line 455: 🔵 **followup:** The safety bound is ℓ, not the language cap — they coincide by only ~4 bits (2^248 < ℓ ≈ 2^251.86). One sentence stating the injectivity precondition (claim domain below the subgroup order) prevents a naive widening if the compiler cap ever rises past 251.

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -92,7 +92,16 @@ pragma language_version >= 0.23.0;
  * memo delivers the amount directly (no discrete-log recovery), so the former
  * `2^48` cap that kept wallet-side BSGS tractable is gone. Amounts stay well under
  * the Jubjub scalar-field order, so the homomorphic balance sums (and the
- * `g^value` encoding) do not wrap. Balances remain Uint<128>.
+ * `g^value` encoding) do not wrap.
```
0xisk: ⚪ **nitpick:** With accumulation, sums can wrap mod ℓ after ~2^124 maximal credits — same resource-bound class as the 2^120 paragraph below, and post-wrap the balance decrypts to sum mod ℓ (silent value loss, not lock). Worth the same qualifier here.

0xisk (changes requested):
Looking good! left some comments.

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: added [7a7274e](https://github.com/OpenZeppelin/compact-contracts/pull/831/commits/7a7274ea38793ddb316462872a4661b6acf8705f)

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: added [7488787](https://github.com/OpenZeppelin/compact-contracts/pull/831/commits/7488787e2286979097286138011b3d0be18d0fa2)

contracts/src/crypto/ElGamal.compact
andrew-fleming: added [2093f35](https://github.com/OpenZeppelin/compact-contracts/pull/831/commits/2093f3586da8b9ed2c1b00a46a2da9ebc8ea2bf4)

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: added [05c3670](https://github.com/OpenZeppelin/compact-contracts/pull/831/commits/05c3670418c3c14ed6b372cb684ed2e7c311918b)

contracts/src/crypto/ElGamal.compact
andrew-fleming: fixed [8122aa1](https://github.com/OpenZeppelin/compact-contracts/pull/831/commits/8122aa12817e383990c07cddcbaec1b7fd124891)

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: fixed [320aff0](https://github.com/OpenZeppelin/compact-contracts/pull/831/commits/320aff0c0aae842824a3b34df2e417f0364f124e)

contracts/src/token/ConfidentialFungibleToken.compact
```diff
@@ -334,7 +347,7 @@ module ConfidentialFungibleToken {
    * @param ct - The ciphertext whose plaintext is being requested.
    * @returns The wallet's cached plaintext for `ct`.
    */
-  witness wit_PlaintextBalance(ct: ElGamal_Ciphertext): Uint<128>;
+  witness wit_PlaintextBalance(ct: ElGamal_Ciphertext): Uint<248>;
```
0xisk, line 371: nit: the doc eight lines up (line 342) still says the wallet caches a `Map<ElGamal_Ciphertext, Uint128>`, but the witness now returns `Uint<248>`.

contracts/src/token/test/ConfidentialFungibleToken.test.ts
```diff
@@ -1102,6 +1104,130 @@ describe.skipIf(isLiveBackend())(
...
+
+      const claim = info.witnesses.find(
+        (w: { name: string }) => w.name === 'wit_PlaintextBalance',
+      );
+      expect(BigInt(claim['result type'].maxval)).toBe((1n << 248n) - 1n);
+
+      const value = info.circuits
+        .find((c: { name: string }) => c.name === 'transfer')
+        .arguments.find((a: { name: string }) => a.name === 'value');
+      expect(BigInt(value.type.maxval)).toBe(MAX128);
```
0xisk: nit: this half reads `transfer`'s `value` from the mock's own signature (`MockConfidentialFungibleToken.compact:63` declares `value: Uint<128>`), so widening the module's `transfer` param would still pass here. The module bound is already pinned by `keeps the transfer bound at the Uint<128> maximum` above. Either drop it or say in a comment that it pins the mock surface.

0xisk:
followup: `ElGamal.assertDecryptsTo`'s `claimedValue` widens from `Uint<128>` to `Uint<248>`, changing a signature already released in 0.3.0-alpha, but `CHANGELOG.md` has no `### Changed` entry under Unreleased. #762 on this branch carries one under `### Fixed`.

(Commented here rather than inline since `CHANGELOG.md` isn't in the diff.)

0xisk (approved):

0xisk:
🔴 **blocking:** raising this from the followup above. `ElGamal.assertDecryptsTo`'s `claimedValue` widens `Uint<128>` → `Uint<248>`, changing a signature already published in 0.3.0-alpha, and `CHANGELOG.md` has no `### Changed` entry under Unreleased. Shipping a release branch with an undocumented public-API change is the kind of gap the changelog exists to prevent, and it is a one-line fix. #762 on this branch already carries an entry under `### Fixed`.

andrew-fleming:
> raising this from the followup above. ElGamal.assertDecryptsTo's claimedValue widens Uint<128> → Uint<248>, changing a signature already published in 0.3.0-alpha, and CHANGELOG.md has no ### Changed entry under Unreleased. Shipping a release branch with an undocumented public-API change is the kind of gap the changelog exists to prevent, and it is a one-line fix. https://github.com/OpenZeppelin/compact-contracts/pull/762 on this branch already carries an entry under ### Fixed.

Agreed. Added an entry to `changed`: [614a159](https://github.com/OpenZeppelin/compact-contracts/pull/831/commits/614a159425ee3a145daba38985485ef75fe4dfc5)

contracts/src/token/ConfidentialFungibleToken.compact
andrew-fleming: fixed [98b5133](https://github.com/OpenZeppelin/compact-contracts/pull/831/commits/98b5133863198ba285a8b5dd4cdafcf8542df5c7)

contracts/src/token/test/ConfidentialFungibleToken.test.ts
andrew-fleming: fixed [15fb978](https://github.com/OpenZeppelin/compact-contracts/pull/831/commits/15fb9788f13db54452a89a18fa682da57ed746a2)

PR #833, opened 2026-08-27
fix(token)!: reject contract mint/burn recipients

### LINKED ISSUE #794: H-03: Missing Contract Commitment Claim When Recipient is a Contract
## H-03 Missing Contract Commitment Claim When Recipient is a Contract

**Severity:** High

---

[The _mint circuit and](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NativeShieldedTokenCore.compact#L234-L235) [_burn circuit](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NativeShieldedTokenCore.compact#L282-L283) in the `NativeShieldedTokenCore` contract [mint tokens](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NativeShieldedTokenCore.compact#L243) or [refund tokens](https://github.com/OpenZeppelin/compact-contracts/blob/0e9d659084fe6579dd0ff9c49c17ae710dcca480/contracts/src/token/NativeShieldedTokenCore.compact#L297) via `mintShieldedToken`/`sendImmediateShielded` to a caller-supplied recipient (mint recipient or refund address), which is typed as `Either<ZswapCoinPublicKey, ContractAddress>`. When `refundTo` is a plain public key, this is a valid, self-contained Zswap output.

However, when `refundTo` resolves to `right<...>(ContractAddress)`, the resulting coin commitment is contract-associated and, per the ledger's well-formedness rules, must be claimed as a shielded receive [by that contract in the same transaction/segment via](https://github.com/midnightntwrk/midnight-ledger/blob/409fe3cb/spec/intents-transactions.md?plain=1#L571-L577) `kernel.claimContractCall` - otherwise the transaction is rejected.

Consider accepting the target contract's entry point and call-randomness as additional witness/public inputs, computing `transientCommit`, and calling `kernel.claimContractCall` [as per this example.](https://github.com/midnightntwrk/midnight-ledger/blob/409fe3cb/ledger/tests/composable-relay.compact#L22-L31) Alternatively, consider only accepting `ZswapCoinPublicKey` (via `left<...>`) as the recipient if contract-address refunds are not a supported use case, to prevent constructing transactions that will always fail.

0xisk:
## Types of changes

Fixes #794

Based on `release/v0.3.0-alpha.1-rc.0`, not `main`. The main-based twin is `fix/audit-nst-h3`.

Only foreign contracts are unclaimable: a coin addressed to `kernel.self()` is claimed by `receiveShielded` in the same call. So the core keeps the `Either` and asserts the right arm is self, and the flavors take a bare `ZswapCoinPublicKey` and expose the arm as `_mintToSelf`. The two-contract claim the finding sketches needs a call into the recipient contract, which a library module cannot make on its caller's behalf. Nothing about the arm is persisted, so it widens again once contract-to-contract calls land, with no state migration.

Breaking: flavor `_mint` / `_burn` take a bare `ZswapCoinPublicKey` where an `Either` used to go.

Not visible in the diff:

- The specs run dry only. This branch predates the live-fixture unification (#694), so the same change was run live off-branch against a local node: 90/90 across the three NST specs. Live coverage of these specs lands with the next release line cut from main.
- The finding's derivation against the ledger and compiler sources is written up locally, not included here. Say if it should land under `docs/`.

## PR Checklist


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Restrict `NativeShieldedToken` / `NativeShieldedTokenFamily` mint recipients and burn refund targets to `ZswapCoinPublicKey`. A contract-addressed coin must be claimed as a shielded receive by the recipient contract in the same transaction, which the module cannot arrange without cross-contract calls, so such transactions were always rejected by the node (audit H-03).
```
andrew-fleming, line 17: We can simplify this, no? Details can be found in this PR -> I'd also use the PR number as the ref and not the audit issue

```suggestion
- Change NativeShieldedToken `_mint` recipient and `_burn` refund recipient to `ZswapCoinPublicKey` (#833)
```

contracts/src/token/NativeShieldedTokenCore.compact
```diff
@@ -44,6 +44,18 @@ pragma language_version >= 0.23.0;
...
+ * would be unclaimable, making the transaction malformed.
+ *
+ * @dev Recipient arm
+ * A contract-addressed commitment counts only if the recipient contract's own
+ * call in the same transaction claims it as a shielded receive. No primitive
+ * claims a receive on another contract's behalf, so nothing this module emits
+ * can satisfy the ledger's effects check for such a coin. Both `mintShieldedToken`
+ * and `sendImmediateShielded` still take the protocol's
+ * `Either<ZswapCoinPublicKey, ContractAddress>`; the circuits wrap the caller's
+ * key in `left` and never expose the address arm.
```
andrew-fleming, line 57: Two things here:
1. I'm wondering if it makes sense to keep the `Either` signatures in the core module. This is following that the core module itself is not shipped during releases to prevent accidental uses (even though this is well documented). Not sure if it's worth keeping the signature since the public API is breaking either way but I figure it's worth considering. Thoughts?
2. Correct me if I'm wrong, but it seems like this isn't taking into account cases where a contract mints to itself and this PR is quietly disallowing that. The issue is only with minting/refunding to other contracts bc they can't claim, no?

andrew-fleming: On a side note, and a bit of nitpick, the offline response included: "...replace the contract-address mint test with regression tests that reject the retired `Either` shape at the API boundary before any transaction is built..."

Regression tests should probably be added since this was stated as part of the remediation. lmk if you disagree

andrew-fleming (commented):
Looking good, @0xisk! I left some comments

contracts/src/token/NativeShieldedTokenCore.compact
0xisk: Agree, fixed in 9b9fae8a: core keeps the `Either` with a self-only right arm claimed via `receiveShielded`, flavors expose it as `_mintToSelf`. Retired-shape regression tests added on both flavors.

CHANGELOG.md
0xisk: Agree, fixed in a47d5a88.

contracts/src/token/NativeShieldedTokenCore.compact
```diff
@@ -44,6 +44,17 @@ pragma language_version >= 0.23.0;
  * compose `NativeShieldedTokenDerivedNonce` and pass its `_deriveNonce()`
  * output, at the cost of making those mints recipient-public.
  *
+ * @notice A recipient may be any coin public key, or this contract's own
+ * address. Any other contract address is rejected.
+ *
+ * @dev Recipient arm
+ * A contract-addressed commitment counts only if the recipient contract's own
+ * call claims it as a shielded receive in the same transaction, and no
+ * primitive claims a receive on another contract's behalf. So the address arm
```
andrew-fleming, line 53: Nitpick: It might be worth clarifying what "counts" means, it seems a little ambiguous

andrew-fleming (approved):
Changes look good! Left a tiny nit

contracts/src/token/NativeShieldedTokenCore.compact
0xisk: Agree — reworded in f4e599d0: the ledger is the one that accepts or rejects the commitment.

andrew-fleming (approved):
LGTM!

Issue #827, opened 2026-08-26
10b. Integrate `Keccak` primitive — replace `persistentHash` message hashing


0xisk:
**🧐 Motivation**

The signatures the stateless presets verify are produced by EVM HSMs over a Keccak-256 digest. The contracts currently hash the signed payload with `persistentHash`, so the on-chain message never matches what the signer actually signed. Verification can only be correct once the message hash is Keccak.

**📝 Details**

Split out of #475, which combined the ECDSA and Keccak work. The signature-verification half is tracked in #826.

Scope:

* Switch the signed-message hash from `persistentHash` to the `keccak256` primitive in `EcdsaSignerManager` and the stateless presets.
* Align the message encoding and domain separation with what the EVM signing infrastructure produces, so an HSM signature validates without re-encoding on the client.
* Measure real per-circuit cost (`@circuitInfo` k / rows) of `keccak256` on the RC toolchain. Keccak sits on every gated circuit, so its row cost is load-bearing for the block budget.

***Blocked*** on Midnight Foundation primitive delivery (Keccak GA). Blocks mainnet deployment of `ShieldedMultiSigV2` and `ShieldedMultiSigToken`, and the corresponding audit scope.

PR #713, opened 2026-07-26
beta: Integrate ecdsa 

andrew-fleming:
WIP

component | repo has now | ECDSA-capable (available today, pre-release) | anticipated stable
-- | -- | -- | --
compactc (compiler) | 0.31.x line (repo compact toolchain) | 0.33.0-rc.2 | 0.33.0
compact-runtime | 0.16.0 | 0.18.0-rc.1 | 0.18.0
midnight-js-* | 4.1.1 | 5.0.0-beta.6 | 5.0.0
compact-js | — | 2.5.5-rc.7 | ~2.5.5
ledger-v9 (JS) @midnightntwrk/ledger-v9 | (uses ledger-v8@8.1.0) | 1.0.0-rc.3 | 1.0.0
onchain-runtime-v4 @midnightntwrk/onchain-runtime-v4 | — | 4.0.0-rc.3 | 4.0.0
node / ledger (on-chain) | mainnet = 1.0.x / ledger-8 | node 2.0.0-rc.4 / ledger 9.1.0.0-rc.3 | node 2.0.0 / ledger 9.1.0.0

PR #154, opened 2025-06-25
Change name to compact

andrew-fleming:
Fixes #153, Fixes #131

#### PR Checklist

- [ ] Tests
- [x] Documentation

PR #842, opened 2026-09-01
feat(multisig): verify sigs with secp256k1 ECDSA

### LINKED ISSUE #826: 10a. Integrate `ECDSA` primitive — replace `stubVerifySignature`
**🧐 Motivation**

The stateless presets (`ShieldedMultiSigV2`, `ShieldedMultiSigToken`) authorize operations by verifying ECDSA signatures on-chain, matching the format produced by EVM HSMs. Until the Compact primitive lands, the verification path is stubbed (`stubVerifySignature` always returns true) and the contracts cannot be deployed to production.

**📝 Details**

Split out of #475, which combined the ECDSA and Keccak work. The Keccak half is tracked in #827.

Scope:

* Replace `stubVerifySignature` with the real `ecdsaVerify` primitive across `EcdsaSignerManager` and the stateless presets.
* Confirm curve compatibility (secp256k1) with the target HSM signing infrastructure.
* Enforce the low-s form / malleability rule on accepted signatures.
* Make the stub hard-fail outside test builds so a stubbed contract cannot be deployed by accident.
* Measure real per-circuit cost (`@circuitInfo` k / rows) on the RC toolchain and benchmark proof time for threshold operations. This is the largest unmeasured cost term in the Monument preset feasibility analysis.

***Blocked*** on Midnight Foundation primitive delivery (ECDSA GA). Blocks mainnet deployment of `ShieldedMultiSigV2` and `ShieldedMultiSigToken`, and the corresponding audit scope.

0xisk:
Refs #826 (HSM-compatibility confirmation and proof-time benchmark remain open there). Supersedes the ECDSA half of #713; the keccak256 half stays in #827.

## Types of changes

Start reading at `crypto/Ecdsa.compact`, then `multisig/EcdsaSignerManager.compact`, then the presets.

Not visible in the diff:

- Both presets carried an identical verification fold, so it moved into `EcdsaSignerManager`; #827's keccak switch now lands in one file. The move also unified two divergent commitment domain tags (`MultiSig:signer:` vs `multisig:signer:`).
- `secp256k1EcdsaVerify` accepts both `s` halves by design and its docs push low-s to callers. `crypto/Ecdsa` is that gate, mirroring OZ Solidity's strict-only `ECDSA.recover`. It costs ~5k rows per signature, dominated by the `Scalar as Bytes<32>` foreign-field decomposition; per-circuit `k`/rows are in the `@circuitInfo` tags. Byte order of that cast is little-endian, pinned by the `n/2` / `n/2 + 1` / `n − 1` boundary tests.
- The bound local in `secp256k1EcdsaVerifyLowS` is deliberate: a short-circuit `&&` puts the stdlib verify behind a guard and ICEs compactc 0.34 (`missing-guard-workarounds`). Upstream report pending.
- `MockEcdsa`'s `Counter` exists only to keep its circuits impure so keys/ZKIR are produced and the gate is proven on live.
- `--feature-zkir-v3` is required (the Secp256k1 types live in the v3 library). Added to `compile:crypto` and `compile:multisig` only; the aggregate `compile` still runs flagless and silently writes nothing for those dirs. Repo-wide flag left as a maintainer call.

Breaking: commitments hash `pkX`/`pkY` under the unified domain (all must be recomputed); V2's `execute` digest is domain-separated and bound to `kernel.self()` and the full recipient; V3's `_instanceSalt` leaves `ledger()`; V2 no longer exports `VerificationState` / `SignerCommitmentInput`.

Deferred: EIP-191/712 digest helpers (#827 decides the signing format); root `compile` flag; `EmptyWitnesses` fixture consolidation.

## PR Checklist

## Summary by CodeRabbit

* **Security Enhancements**
  * Multisig approvals now use active secp256k1 ECDSA signature verification.
  * Added protection against invalid signatures, incorrect signing keys, duplicate signers, and replayed operations.

* **API Updates**
  * Multisig operations now accept typed public keys and ECDSA signatures.
  * Signer commitments now derive from separate public-key coordinates, improving signing compatibility.

* **Validation**
  * Expanded multisig coverage for nonce handling, contract-bound signatures, and cross-instance replay prevention.


0xisk:
@coderabbitai full review

0xisk:
**@coderabbitai** review

0xisk:
@coderabbitai review

0xisk:
**@coderabbitai** full review

contracts/src/multisig/test/EcdsaTestUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+  nonce: bigint;
+  toAddress: Uint8Array;
+  coinColor: Uint8Array;
+  amount: bigint;
+}): Uint8Array {
+  return persistentVec([
+    u256(params.nonce),
+    params.toAddress,
+    params.coinColor,
+    u256(params.amount),
```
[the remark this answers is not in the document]
0xisk: Done in 46d65162 — the digest now includes a `multisig:execute:` domain tag and `kernel.self()`, with a cross-instance replay test.

contracts/src/multisig/test/EcdsaTestUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+/** ShieldedMultiSigV2 `execute` digest. No domain prefix. */
+export function executeMsgHash(params: {
+  nonce: bigint;
+  toAddress: Uint8Array;
+  coinColor: Uint8Array;
+  amount: bigint;
+}): Uint8Array {
+  return persistentVec([
+    u256(params.nonce),
+    params.toAddress,
```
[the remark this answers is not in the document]
0xisk: Done in 46d65162 — the digest hashes the full `Recipient` (kind and address) via `persistentHash<Proposal_Recipient>`, with a kind-flip regression test.

contracts/package.json
```diff
@@ -28,7 +28,7 @@
     "compile:access": "compact-compiler --dir access",
     "compile:archive": "compact-compiler --dir archive",
     "compile:crypto": "compact-compiler --dir crypto",
-    "compile:multisig": "compact-compiler --dir multisig",
+    "compile:multisig": "compact-compiler --dir multisig --feature-zkir-v3",
```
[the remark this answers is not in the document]
0xisk: Done in b608d8ef — declared as a devDependency (`^2.2.0`).

contracts/src/multisig/test/EcdsaTestUtils.ts
```diff
@@ -0,0 +1,186 @@
...
+import {
+  type CompactType,
+  CompactTypeBoolean,
+  CompactTypeBytes,
+  CompactTypeVector,
+  convertBigintToBytes,
+  persistentHash,
+  type Secp256k1Point,
+} from '@midnight-ntwrk/compact-runtime';
+import { secp256k1 } from '@noble/curves/secp256k1.js';
```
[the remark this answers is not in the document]
0xisk: Done in b608d8ef (same fix as the package.json thread).

0xisk:
@coderabbitai full review

contracts/src/multisig/presets/ShieldedMultiSigV3.compact
```diff
@@ -33,47 +33,26 @@ pragma language_version >= 0.26.0;
  * hash prevent signatures for one operation type from being replayed as
  * the other.
  *
- * @notice ECDSA verification is stubbed. Replace `stubVerifySignature` with
- * `ecdsaVerify`, and `persistentHash` with `keccak256`, once the Compact
- * ECDSA and Keccak primitives are available.
+ * @notice Threshold authorization is enforced with secp256k1 ECDSA
+ * (`secp256k1EcdsaVerify`) over `persistentHash` digests, via
+ * `EcdsaSignerManager`. The approval vectors are fixed at `Vector<2, ...>`,
+ * so this contract is 2-of-3 only.
```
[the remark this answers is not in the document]
0xisk: Done in 34e046b4 — both presets now name `secp256k1EcdsaVerifyLowS` and state that high-s signatures are rejected.

0xisk:
@coderabbitai full review

contracts/src/multisig/presets/ShieldedMultiSigV2.compact
```diff
@@ -119,162 +98,128 @@ export circuit deposit(coin: ShieldedCoinInfo): [] {
...
 export circuit getNonce(): Uint<64> {
   return _nonce;
 }
 
+/**
+ * @description Returns the number of registered signers.
+ *
+ * @circuitInfo k=6, rows=34
+ *
+ * @returns {Uint<8>} The signer count.
```
pepebndc, line 198: Blocking (High): this claim is what breaks. The execute digest is `persistentHash([nonce, to.address, coin.color, amount])`: it commits to neither `kernel.self()` nor a domain tag, while V3's mint and burn digests commit both. Two V2 instances that register the same signer keys accept each other's execute signatures whenever their nonces align, and fresh deployments both start at 0, so "a message authorizes at most one operation" is false across instances. This was inert while the stub ignored signatures; now the digest is the sole authorization root. Fix: add `kernel.self().bytes` and a domain prefix (e.g. "multisig:execute:") to the digest, and port the V3 cross-instance replay test to V2 (a faithful port would fail today).

contracts/src/multisig/presets/ShieldedMultiSigV2.compact
```diff
@@ -119,162 +98,128 @@ export circuit deposit(coin: ShieldedCoinInfo): [] {
...
-  };
-
-  const finalState = fold(verifySignature, initialState, pubkeys, signatures);
-  Signer_assertThresholdMet(finalState.validCount);
+  EcdsaSignerManager_assertApprovals(msgHash, pubkeys, signatures);
 
   // Execute transfer
   const normalizedRecipient = Proposal_toShieldedRecipient(to);
   return Treasury__send(coin, normalizedRecipient, amount);
 }
```
pepebndc, line 154: Blocking (High): the digest commits `to.address` but not `to.kind`. The kind alone decides whether those 32 bytes are read as a ZswapCoinPublicKey or a ContractAddress (`Proposal_toShieldedRecipient`), so a relayer holding the collected signatures can flip the rail on an authorized spend before submitting, without invalidating anything. Commit the kind in the digest, or hash the canonicalized recipient the way V3's mint does. Whether live-node output rules would reduce the flip to griefing is an open question; the digest must commit the kind either way.

contracts/src/multisig/test/ShieldedMultiSigV3.test.ts
```diff
@@ -468,48 +567,37 @@ describe('ShieldedMultiSigV3', () => {
...
       it('should increment monotonically', async () => {
         for (let i = 0; i < 5; i++) {
-          await multisig.mint(
-            1n,
-            USER_RECIPIENT,
-            [PK1, PK2],
-            [DUMMY_SIG, DUMMY_SIG],
-          );
+          await mint(multisig, 1n, USER_RECIPIENT, [S1, S2]);
           expect(await multisig.getNonce()).toEqual(BigInt(i + 1));
```
pepebndc, line 571: Blocking (CI): this test fails on the live backend. Simulator 0.4.0 rejects an explicit `contractAddress` that differs from the deployed one at `create` (compact-tools #154), so `'11'.repeat(32)` throws before `mint` ever runs, and `unit-live` fails with the wrong error. Add `it.skipIf(isLiveBackend())` per this file's existing pattern, or skip the address override on live where two real deployments already differ.

pepebndc (changes requested):
The core fix is right and it held up under independent verification: I compiled 22/22 with `--feature-zkir-v3` on toolchain 0.34.0, ran both changed suites (60/60, dry backend), and re-measured the circuit costs (your table matches: execute k=17/84,235, mint 74,461, burn 82,125). The fold cannot count a failed or duplicate-signer verification, the pkX/pkY commitment change preserves domain separation, the identity point hard-fails before verification on every path, and the low-s argument is sound within a single instance. No verify-nothing path survives anywhere in the tree.

Three blockers, all V2-side or CI-side. The two digest findings exist because this PR promotes the digest from decoration to the sole authorization root: it now must commit everything that changes the operation's effect, and V2's does not. Inline comments carry the details.

Not blocking, but please handle before or at merge: change "Fixes #826" to "Refs #826" (merging would auto-close an issue whose HSM-compatibility bullet and proof-time benchmark are still open, and whose low-s deviation has no recorded acceptance). On your open questions: declaring `@noble/curves` as a pinned devDependency is the right call, and I would add `--feature-zkir-v3` to the aggregate `compile` script too, since turbo re-runs it over the same sources after `compile:multisig` and the exit-0-no-artifact failure mode is silent. The simulator `contractAddress` option is a supported API (it is documented in compact-tools 0.4.0), but see the inline comment on its live-backend behavior.

contracts/src/multisig/presets/ShieldedMultiSigV2.compact
0xisk: Agree — fixed in 46d65162.

contracts/src/multisig/presets/ShieldedMultiSigV2.compact
0xisk: Agree — fixed in 46d65162.

contracts/src/multisig/test/ShieldedMultiSigV3.test.ts
0xisk: Agree — fixed in cc3d6964.

0xisk:
@pepebndc all three blockers addressed (the two digest ones had landed in 46d65162 just before your review; cross-instance live fix above). `Fixes` → `Refs #826` done.

On the aggregate `compile` flag: not in this PR. Under `--feature-zkir-v3` `MockConfidentialFungibleToken` fails in zkir with `Unsupported test_eq: JubjubScalar == JubjubScalar` — the op is `assert(r != 0 as JubjubScalar, …)` at `crypto/ElGamal.compact:189`, reached via `_burn` → `ElGamal_encrypt`. The wrapper still reports success because `compact-builder` runs `script -qc` without `-e` and swallows the exit code. A follow-up PR will rewrite that check so the whole repo compiles under v3 and the flag can move to the aggregate script.

Low-s is now enforced in-circuit (`c4a106cb`, `crypto/Ecdsa`), so the deviation you noted no longer applies.

pepebndc (approved):

contracts/package.json
```diff
@@ -27,8 +27,8 @@
     "compile": "compact-compiler --exclude '*/archive/*'",
     "compile:access": "compact-compiler --dir access",
     "compile:archive": "compact-compiler --dir archive",
-    "compile:crypto": "compact-compiler --dir crypto",
-    "compile:multisig": "compact-compiler --dir multisig",
+    "compile:crypto": "compact-compiler --dir crypto --feature-zkir-v3",
+    "compile:multisig": "compact-compiler --dir multisig --feature-zkir-v3",
```
andrew-fleming: So this is incidentally safe. A potential risk is however `build` and `compile` are fixed. The risk I'm worried about is compiling and testing modules with v3 and shipping them targeting v2 or vice versa

This is a weird state to be in. Not all modules in `crypto/` and `multisig/` require v3. idk if there's a clean path to follow, just pointing this out

andrew-fleming: Also as a reminder, the preset contracts should NOT be shipped in the release

CHANGELOG.md
```diff
@@ -7,8 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
...hieldedMultiSigV2` and `ShieldedMultiSigV3` approvals with `secp256k1EcdsaVerifyLowS` from the new `crypto/Ecdsa` module, removing `stubVerifySignature` from both presets. `execute` / `mint` / `burn` now take `Vector<2, Secp256k1Point>` public keys and `Vector<2, Secp256k1EcdsaSignature>` signatures, and signer commitments hash the public-key coordinates (`pkX`, `pkY`) instead of a `Bytes<64>` key. `ShieldedMultiSigV2`'s `execute` digest is now domain-separated and bound to the contract instance (`kernel.self()`) and the full recipient (kind and address). Both presets share the verification logic through the new `EcdsaSignerManager` module, which unifies the signer-commitment domain separator on `multisig:signer:`; `ShieldedMultiSigV2` previously used `MultiSig:signer:`, so its commitments change again. The module owns the instance salt and the signer registry, so `ShieldedMultiSigV3`'s `ledger()` reader no longer exposes `_instanceSalt`, and `ShieldedMultiSigV2` no longer exports the `VerificationState` and `SignerCommitmentInput` structs, which were part of its generated artifact types. These primitives require ZKIR v3, so `compile:multisig` now passes `--feature-zkir-v3`. (#826)
```
andrew-fleming: I mentioned this in the other PR: I prefer we keep changelog entries focused and let the PR descriptions go into depth regarding the changes. Or if we want so many details, maybe we can organize the information better. As mentioned before, it's a stylistic preference 🤷‍♂️

contracts/src/multisig/EcdsaSignerManager.compact
```diff
@@ -0,0 +1,229 @@
...
+ * commitment in every instance and cannot be correlated across contracts.
+ *
+ * Wraps `Signer<Bytes<32>>` for the registry and threshold. Consuming
+ * contracts own the digest: they compute the message hash their signers
+ * signed, then hand it to `assertApprovals` with the parallel public-key and
+ * signature vectors.
+ *
+ * The module's ledger is not exported, so a contract importing it under a
+ * prefix keeps the salt and the signer registry out of its generated
+ * `ledger()` reader.
```
andrew-fleming, line 23: Why is this the behavior? Should it not be at the contract's discretion what is and is not readable? The ledger is still public whether it's exposed through the impl contract or not

contracts/src/multisig/EcdsaSignerManager.compact
```diff
@@ -0,0 +1,229 @@
...
+   */
+  export circuit initialize<#n>(
+    instanceSalt: Bytes<32>,
+    signerCommitments: Vector<n, Bytes<32>>,
+    threshold: Uint<8>
+  ): [] {
+    _instanceSalt = disclose(instanceSalt);
+    assert(threshold <= 2, "EcdsaSignerManager: threshold cannot exceed 2 (assertApprovals verifies 2 signatures)");
+    Signer_initialize<n>(signerCommitments, threshold);
+  }
```
andrew-fleming, line 85: Having `initialize` call another `initialize` violates the [extensibility rules](https://docs.openzeppelin.com/contracts-compact/extensibility#modules):

```
Modules MUST:
...
Avoid defining or calling constructors or initialize() directly.
```

The reason is that if some modules call other modules' `initialize`, it makes things very confusing for what needs to be initialized in the constructor and easy to accidentally corrupt the state

andrew-fleming: Further, if we create a similar `XSignerManager` that imports `Signer`, we just re-introduced the [dedup Initializable bug with transient deps](https://github.com/andrew-fleming/mn-module-shared-dep) for anyone importing both modules, no?

At the very least, there should be very loud documentation on this. I'd suggest refactoring instead though. Either:
- Move signer logic to a core module and explicitly document that only one signer may be used
- Remove the Signer dep in this module and rewrite the logic it needs

lmk if you disagree. Happy to consider a better approach if we have one

contracts/src/multisig/EcdsaSignerManager.compact
```diff
@@ -0,0 +1,229 @@
...
+ * `ledger()` reader.
+ */
+module EcdsaSignerManager {
+  import CompactStandardLibrary;
+  import "../crypto/Ecdsa" prefix Ecdsa_;
+  import "./Signer"<Bytes<32>> prefix Signer_;
+
+  // ─── State ──────────────────────────────────────────────────────────────────
+
+  ledger _instanceSalt: Bytes<32>;
```
andrew-fleming, line 32:
```suggestion
  sealed ledger _instanceSalt: Bytes<32>;
```
Should be `sealed`, no?

andrew-fleming (commented):
Forgive the post-merge review spam! I can open issues if preferred, but it's easier to discuss and highlight in case I'm missing something

contracts/src/multisig/EcdsaSignerManager.compact
0xisk: Agree — dropped the note in #887 (7b1497e) and exported `_instanceSalt` in #925.

contracts/src/multisig/EcdsaSignerManager.compact
0xisk: Agree — fixed in #925. `Signer` has no `initialize` any more; `EcdsaSignerManager_initialize` configures the registry through `_addSigner` / `_changeThreshold`, so no module calls another module's `initialize`.

contracts/src/multisig/EcdsaSignerManager.compact
0xisk: Agree — `export sealed` in #925.

